PR #21, opened 2015-10-19
Update ringpop-go to use updated bark import path

prashantv:
bark has moved from github.com/uber to github.com/uber-common

r: @akshayjshah


Godeps/Godeps.json
```diff
@@ -1,6 +1,9 @@
 {
 	"ImportPath": "github.com/uber/ringpop-go",
 	"GoVersion": "go1.5.1",
+	"Packages": [
+		"./..."
```
CorgiMan: What's this?

prashantv: When you use `godeps`, you should save dependencies for all subpackages as well, and so you should run:
```bash
godep save ./...
```

It looks like it was previously only doing godeps for the root package.

CorgiMan: Nice, good to know!

CorgiMan:
Let's get the Money!

PR #22, opened 2015-10-20
Fix headers sent with forwarded requests

akshayjshah:
Per the [TChannel spec](https://github.com/uber/tchannel/blob/master/docs/thrift.md), the minimum header is 2 bytes. Without this change, forwarded requests always time out.

### DIFF
--- a/forward/request_sender.go
+++ b/forward/request_sender.go
@@ -127,7 +127,12 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte) <-chan error 
 			return
 		}
 
-		_, arg3, _, err := raw.WriteArgs(call, []byte{0, 0}, s.request)
+		var arg3 []byte
+		if s.format == tchannel.Thrift {
+			_, arg3, _, err = raw.WriteArgs(call, []byte{0, 0}, s.request)
+		} else {
+			_, arg3, _, err = raw.WriteArgs(call, nil, s.request)
+		}
 		if err != nil {
 			errC <- err
 			return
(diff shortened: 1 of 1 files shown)


thanodnl:
Thanks for the find!

There is a problem though. These two bytes were there before, but it crashed a json example in the repository. PR: https://github.com/uber/ringpop-go/pull/14

This indicates to us we need to dig deeper into how support all TChannel protocols in HandleOrForward.

If this fix solves the problems you are having you can use this feature branch in F3 till we have a final fix for this. But merging in a branch which is known to fail somewhere else is not preferred.

akshayjshah:
Hmmm, I didn't realize that this would break the JSON forwarding.

Is it feasible to merge this into master and fix the JSON forwarding afterwards? As far as I know, I'm the only person using this library in production, and I'd really prefer to deploy the master branch. To me, it seems sensible to have master support production, with any non-prod features in development branches.

prashantv:
I think the simplest fix given the current API is to specify the headers as part of the HandleOrForward. That way the caller can choose to pass `nil` for json or `0,0` bytes for Thrift.

This should be temporary till we come up with a good API to forward the request directly from TChannel.

prashantv:
Take a look at #24

akshayjshah:
@prashantv and @thanodnl - the request format is in scope at the point we write the headers. I've amended this PR to switch on the request type, so this should now handle JSON and Thrift calls properly. The default is the current `nil` behavior, so it shouldn't change behavior for any non-Thrift callers.

Is this a workable solution?

thanodnl:
:+1: lgtm

Issue #23, opened 2015-10-20
Support all (3) protocols of TChannel in HandleOrForward


thanodnl:
Currently thrift over TChannel is not working as reported by https://github.com/uber/ringpop-go/pull/22. Before JSON over TChannel was not working as fixed by https://github.com/uber/ringpop-go/pull/14.

As one could see is that both PR's are the exact inverse of each other.

Since HandleOrForward is one of -maybe the biggest- driver to use Ringpop it should work on all supported TChannel protocols. We need to figure out how to do this the correct way.

PR #24, opened 2015-10-20
Add headers to forwarder and add test

prashantv:
Short-term fix for #23

Since thrift/JSON require different headers, pass headers from the caller.


akshayjshah:
Works for me.

akshayjshah:
Actually, this may not be necessary. I've amended #22 to switch on the request type - it's a much smaller change. What do you think?

prashantv:
I'm fine with either way, although I think sending headers might be important for a lot of applications, so this change might be needed soon.

akshayjshah:
Gotcha. This is definitely more flexible; if we're planning to make another pass at the forwarding API as a whole, though, it may be better to keep this round's changes small.

I'm honestly fine with whichever approach the ringpop team prefers.

prashantv:
Closing, this is now handled in #141

PR #26, opened 2015-11-10
Add forward method

iproctor:
Expose a forward method to make replication with LookupN possible.


jwolski:
yep lgtm. in general though, we should be careful to expose the best (programmatic) API that we can. I'm not sure there's a better place to expose `forward()` right now though.

PR #28, opened 2015-11-16
Disable TChannel retries and zipkin tracing for ringpop.

thanodnl:
Ringpop causes too many traces to be collected. This change is to disable tracing on internal ringpop calls. The biggest problems are tracing ping/pingreq, since these happen frequently from all machines to all machines.

cc @prashantv @uber/ringpop

### DIFF
--- a/shared/shared.go
+++ b/shared/shared.go
@@ -1,0 +1,20 @@
+package shared
+
+import (
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/uber/tchannel-go"
+)
+
+var retryOptions = &tchannel.RetryOptions{
+	RetryOn: tchannel.RetryNever,
+}
+
+func NewTChannelContext(timeout time.Duration) (tchannel.ContextWithHeaders, context.CancelFunc) {
+	return tchannel.NewContextBuilder(timeout).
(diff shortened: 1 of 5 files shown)


shared/shared.go
```diff
@@ -0,0 +1,20 @@
+package shared
+
+import (
+	"time"
+
+	"golang.org/x/net/context"
```
prashantv: nit: move this import below the github import

prashantv:
lgtm, thanks for making the change so quickly

jwolski:
`shared/shared.go` is an unfamiliar abstraction. Is this a common Go idiom?

Anyways, let's move more towards a `Client` abstraction that is akin to https://github.com/uber/ringpop-node/blob/master/client.js. All calls through TChannel and dealing with its nuances are meant to be centralized there. It seems like `shared.go` can turn into this. I'm not sure if that was its intention or whether it was just meant as a general utility library.

thanodnl:
@jwolski the `shared/shared.go` is there to not run into circular dependencies. I tried adding it to `ringpop.go` but that would not compile. If we need more of these functions we might want to move to your client abstraction.

PR #29, opened 2015-11-19
Expose a function on ringpop with all the current active members.

thanodnl:
Users that want to get a list of all their active members should be able to get a slice of all those members.

A use case for this is if the user wants to do a manual fanout to all running instances.

cc @uber/ringpop

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -362,8 +362,8 @@ func (rp *Ringpop) ringEvent(event interface{}) {
 	}
 }
 
-func (rp *Ringpop) GetMembers() []string {
-	return rp.node.GetMembers()
+func (rp *Ringpop) GetReachableMembers() []string {
+	return rp.node.GetReachableMembers()
 }
 
 //= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
--- a/ringpop.go
+++ b/ringpop.go
@@ -362,6 +362,17 @@ func (rp *Ringpop) ringEvent(event interface{}) {
 	}
 }
 
+func (rp *Ringpop) GetMembers() []string {
+	members := rp.node.MemberStats().Members
+	var addresses []string
+	for _, member := range members {
+		if member.Status == swim.Alive || member.Status == swim.Suspect {
+			addresses = append(addresses, member.Address)
+		}
+	}
+	return addresses
+}
+
 //= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
 //
 //	Stats
--- a/ringpop.go
+++ b/ringpop.go
@@ -363,14 +363,7 @@ func (rp *Ringpop) ringEvent(event interface{}) {
 }
 
 func (rp *Ringpop) GetMembers() []string {
-	members := rp.node.MemberStats().Members
-	var addresses []string
-	for _, member := range members {
-		if member.Status == swim.Alive || member.Status == swim.Suspect {
-			addresses = append(addresses, member.Address)
-		}
-	}
-	return addresses
+	return rp.node.GetMembers()
 }
 
 //= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -324,3 +324,17 @@ func (m *memberlist) String() string {
 func (m *memberlist) Iter() *memberlistIter {
 	return newMemberlistIter(m)
 }
+
+func (m *memberlist) GetActiveMemberAddresses() []string {
+	var active []string
+
+	m.members.RLock()
+	for _, member := range m.members.list {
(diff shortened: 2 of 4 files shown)


ringpop.go
```diff
@@ -362,6 +362,17 @@ func (rp *Ringpop) ringEvent(event interface{}) {
 	}
 }
 
+func (rp *Ringpop) GetMembers() []string {
+	members := rp.node.MemberStats().Members
```
jwolski: It seems slightly odd to depend on a Stats API to get the members? I'm pretty unfamiliar with the Go codebase. Is that the best API we've got from the Node type?

Also... I'd prefer that you did leak the 'Status' abstraction into the Ringpop type. Checking for 'Alive' and 'Suspect' shuld be limited to the Membership API.

bromanescu: For now we can do just with the Alive nodes.

jwolski: Yea, that's exactly what I was thinking. A mere pass-through.

thanodnl:
@jwolski @bromanescu I refactored as suggested and .... wait for it ..... added an actual test!

Please have a last look and lets merge it after that.

swim/node.go
```diff
@@ -510,3 +510,7 @@ func (n *Node) pingNextMember() {
 		"numErrors": len(errs),
 	}).Warn("ping request inconclusive due to errors")
 }
+
```
jwolski: Is there a reason why you didn't propagate the name 'GetActiveMemberAddresses' upwards?

I was thinking in place of using 'ActiveMember', which is introduces a new term into our Ringpop vocabulary, using 'GetPingableMembers'? What do you think?

swim/memberlist.go
```diff
@@ -324,3 +324,17 @@ func (m *memberlist) String() string {
 func (m *memberlist) Iter() *memberlistIter {
 	return newMemberlistIter(m)
 }
+
+func (m *memberlist) GetActiveMemberAddresses() []string {
+	var active []string
+
+	m.members.RLock()
```
jwolski: Is it necessary to have a read-lock while we append to the active list? I'm thinking that we can make the locking more fine-grained by simply locking while copying `m.members.list` and iterate over the copy while not locking.

thanodnl: I chose this over copying the list because a copy operation allocates new memory which we don't need. I don't know what the time difference between a copy and the iteration is. But now I run over the list once and only append the members I need to the new list. I'm also fine with copying the list and iterate over a copy.

First I used the accessors to the memberlist, but every operation tries to acquire the lock for its self. Here I just lock the list once, iterate over it and release the lock.

swim/node.go
thanodnl: I would be fine with PingableMembers as well, I'd prefer to keep the GetMembers on the Ringpop struct though since the application embeding Ringpop does not really care about the pingable concept in my opinion, they just want a list of members from ringpop, they only care about members that are currently in working state.

jwolski: Yea, this is a good point.

swim/memberlist.go
jwolski: I think the coarser-grained locking is acceptable. we can always revisit later.

swim/node.go
thanodnl: Your point about future compatibility makes a lot of sense, how about `GetReachableMembers`. Using `Pingable` feels wrong if you look at the definition of `[Memberlist::Pingable(Member)](https://github.com/uber/ringpop-go/blob/feature/get-members/swim/memberlist.go#L127-L131)` where a pingable member could never be its own host, while its own host is definitely part of the list of all 'reachable' nodes.

swim/memberlist.go
thanodnl: True. But looking at the code that is already in this package I see this same pattern over and over again. So I would vote for consistency at this point and in an other story figure out what the performance difference is between this method of locking vs only locking a copy operation.

1. [`NumPingableMembers`](https://github.com/uber/ringpop-go/blob/feature/get-members/swim/memberlist.go#L134-L144)
2. [`RandomPingableMembers`](NumPingableMembers)
3. [`GetMembers`](https://github.com/uber/ringpop-go/blob/feature/get-members/swim/memberlist.go#L168-L176) (not the method being added in this PR)

swim/node.go
jwolski: agreed re: the local member. That was the only nuance that felt out of place when using Pingable. I like Reachable a lot. Let's go with it.

thanodnl: I renamed the function to `GetReachableMembers` throughout all files now.

jwolski: love it

swim/node.go
```diff
@@ -510,3 +510,7 @@ func (n *Node) pingNextMember() {
...
+func (n *Node) GetMembers() []string {
```
bromanescu: Suggest renaming the function as GetActiveMembers. Same comment for the ringpop method. other than that, lgtm

PR #31, opened 2015-11-26
Feature/thrift forward adapter

thanodnl:
**TL;DR** This PR adds an abstraction for application owners to easily forward thrift calls that are invoked over tchannel to other nodes in a ringpop enabled application.
Blocked on: https://github.com/uber/tchannel-go/pull/94

[tchannel-go](https://github.com/uber/tchannel-go) provides an application called `thirft-gen` that generates code so thrift calls can be invoked over tchannel. This PR enhances ringpop with features to easily route those remote calls to a ringpop enabled application by wrapping around your service implementation. In the wrapping layer you *configure* how endpoints in services should be routed by providing the following 4 fields.

 - `N` This is the number of nodes on the ring that should be looked up based on the key
 - `Fanout` This is the actual number of nodes that the request will be forwarded to
 - `Key` This is a closure with the same arguments as your service implementation (including the thrift context) that returns a string. The string returned will be used as the key to do the ring lookup
 - `Reducer` When `Fanout` is bigger than 1 there are multiple response values. These get passed into the reducer to do a reduce phase comparable to MapReduce. The input is a slice for every output argument of your service endpoint. The output has the same signature as your service.

The adapter gets generated by `thrift-gen` based on the thrift specification you pass in. To accomplish this we made some changes to thrift-gen; https://github.com/uber/tchannel-go/pull/94. When those changes are finished and accepted `thrift-gen` accepts multiple `--template <path>` flags. With this flag you can pass a template file -suggested convention is to call these `<module>.thrift-gen`- that can generate your code based on the thrift specification. In this PR we add a `ringpop.thrift-gen` file to ringpop-go.

The usage of the generated code looks as follows;

`keyvalue.thrift`:
```thrift
exception KeyNotFound {
  1: required string key
}

exception InvalidKey {
  1: required string reason
}

service KeyValue {
  // If the key does not exist, KeyNotFound is returned.
  string Get(1: string key) throws (
    1: KeyNotFound notFound
  )

  // Set returns InvalidKey is an invalid key is sent.
  void Set(1: string key, 2: string value) throws (
    1: InvalidKey invalidKey
  )

}

// Returned when the user is not authorized for the Admin service.
exception NotAuthorized {}

service Admin {
  void clearAll() throws (1: NotAuthorized notAuthorized)
}
```
This thrift file describes a key value service. The implementation of this service will not be covered here, but we expect you to be able to create an implementation of the `keyvalue.TChanKeyValue` interface as an implementation of your service.

After you have this you can wrap ringpop around your implementation like the following code:

```go
var handler gen.TChanKeyValue
handler = handlers.NewKeyValueHandler(store)

// wrap the handler with the ringpop adapter
handler = gen.NewRingpopKeyValueAdapter(handler, ring, ch, gen.KeyValueConfiguration{
	Set: &gen.KeyValueSetConfiguration{
		N:      2,
		Fanout: 2,
		Reducer: func(errs []error) error {
			return errs[0]
		},
		Key: func(ctx thrift.Context, key string, value string) string {
			return key
		},
	},
	Get: &gen.KeyValueGetConfiguration{
		N:      2,
		Fanout: 1,
		Key: func(ctx thrift.Context, key string) string {
			return key
		},
	},
})

server.Register(gen.NewTChanKeyValueServer(handler))
```

In this example we *configure* the `Set` endpoint of our service with `N: 2` and `Fanout: 2`. This makes sure there will be 2 copies of the value for the key in the cluster during writing a key.

The `Get` endpoint is configured with `N: 2` and `Fanout: 1`. This is because we only care about 1 version of the key since the key should be replicated during the write phase. We need to specify `N: 2` because ringpop needs to know which nodes hold the data. The implementation of the fanout is that it prefers local execution over remote execution when possible.

Valid values for `N` and `Fanout` are:

1. Positive non 0 numbers
2. `ringpop.ALL`

The numbers indicate the number of nodes that should be selected as candidates, `ALL` indicates that the complete cluster will be a candidate for the operation. A good candidate for this would be `Admin::clearAll` from our thrift file. When `Admin::clearAll` is executed on 1 host it automatically gets forwarded to all online hosts.

Please take a look at the changes and be fierce in feedback!

cc @uber/ringpop and @prashantv in case you're interested in following along.


ringpop.thrift-gen
```diff
@@ -0,0 +1,221 @@
...
+        return {{ if $Method.ReturnType }}rs[0], {{ end }}errs[0]
+    default:
+        return t.{{ $cfg }}.Reducer({{ if $Method.ReturnType }}rs, {{end}}errs)
+    }
+}
+{{ end}}
+// ------ End of {{ $Service.Name }} ------
+{{ end}}
+
+func isFrowarded(ctx {{ contextType }}) bool {
```
jwolski: spelling mistake, `isFrowarded` -> `isForwarded`

ringpop.go
```diff
@@ -443,3 +463,15 @@ func DeserializeThrift(b []byte, s thrift.TStruct) error {
 	transport := thrift.NewStreamTransportR(reader)
 	return s.Read(thrift.NewTBinaryProtocolTransport(transport))
 }
+
+func (rp *Ringpop) SizeForReplicationFactor(rf ReplicationFactor) int {
+	switch {
+	case rf == ALL:
+		// TODO use memory effiecient algorithm over memory copying
+		return len(rp.GetReachableMembers())
```
jwolski: Expose an API from the memberlist like `GetReachableMemberCount()`

ringpop.go
```diff
@@ -443,3 +463,15 @@ func DeserializeThrift(b []byte, s thrift.TStruct) error {
...
+	// case rf == QUORUM:
```
jwolski: Get rid of these commented lines. We won't forget, I promise ;)

ringpop.thrift-gen
```diff
@@ -0,0 +1,221 @@
+{{ define "arguments" }}ctx {{ contextType }}{{ range .Arguments }}, {{ .Name }} {{ .Type | goType }}{{ end }}{{ end}}
```
jwolski: Is it necessary to have some of this intricate logic in generated code? Most of the implementation of the functions `getDestinations()`, `getClients()`, `getClientForDest()` do not require generation/templating whatsoever and are probably best served by baking them into core Ringpop APIs. It seems that only reason they are templated is because of their return type. Instead, can Ringpop preserve a cache of remote clients of `interface{}` and when it needs to generate a remote client, farm that off to Thrift-generated code?

For example, in core Ringpop, I'm thinking of a new API like, `NewRouter()`:

```
type RouterConfig struct {
    N ReplicationFactor
    Fanout ReplicationFactor
}

type RoutingClientGenerator interface {
    CreateClient() interface{}
}

func NewRouter(cfg RouterConfig, clientGenerator RoutingClientGenerator) (*RingpopRouter) {

}

func (*RingpopRouter) GetClients() ([]interface{}) {

}
```

then, in generated

```
clients := router.GetClients(...)
// cast to T
```

ringpop.go
```diff
@@ -347,6 +356,17 @@ func (rp *Ringpop) LookupN(key string, n int) []string {
 	return rp.ring.LookupN(key, n)
 }
 
+func (rp *Ringpop) LookupReplicationFactor(key string, n ReplicationFactor) []string {
```
jwolski: Instead what if we were to expose `LookupAll()` and later `LookupQuorum()` functions? Having `Lookup()`, `LookupN()`, and now `LookupReplicationFactor()` (that internally uses a set of consts to do the dispatching) is not as straight-forward an API.

jwolski:
This is cool.

It def. requires some tests. I'd even suggest pulling some of the fan-out logic out into other PRs, if possible since you've conflated both Thrift support with things that are relatively unrelated.

But more fundamentally, I think there's too much logic in the generated code. I think most of it can be pulled out into core Ringpop APIs, by potentially introducing something like a Ringpop Router object. Though, we'll have to think about it some more since we have to decide how much, internally, Ringpop should know about Thrift.

ringpop.go
```diff
@@ -96,6 +96,25 @@ func mergeDefault(opts *Options) *Options {
...
+type RingpopInterface interface {
+	Destroy()
+	Destroyed() bool
+	App() string
+	WhoAmI() string
+	Uptime() time.Duration
+	RegisterListener(l EventListener)
+	Bootstrap(opts *BootstrapOptions) ([]string, error)
+	HandleEvent(event interface{})
+	Checksum() uint32
```
jwolski: This should be qualified with Membership or Ring since both have checksums.

ringpop.go
```diff
@@ -366,6 +385,10 @@ func (rp *Ringpop) GetReachableMembers() []string {
 	return rp.node.GetReachableMembers()
 }
 
+func (rp *Ringpop) CountReachableMembers() int {
```
jwolski: Should be unsigned?

ringpop_stub_test.go
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
jwolski: Better to name this file 'ringpop_stub.go'? The _test suffix makes it sound like it contains tests.

router.go
```diff
@@ -0,0 +1,156 @@
...
+	"sync"
+
+	"github.com/uber/ringpop-go/swim"
+)
+
+// Replication constants
+type ReplicationFactor int
+
+const (
+	ALL ReplicationFactor = -1 * (iota + 1)
```
jwolski: what's the `-1` multiplier for?

ringpop.go
thanodnl: I copied all functions that were already exposed (aka starting with an upper case letter) here. But if you think it should not be in the public interface for ringpop I can remove it.

Ideally we would call this interface `Ringpop` but that might be something for a future PR because it gets kinda messy.

router.go
```diff
@@ -0,0 +1,156 @@
...
+func (r *router) HandleEvent(event interface{}) {
+	switch event := event.(type) {
+	case swim.MemberlistChangesReceivedEvent:
+		for _, change := range event.Changes {
+			r.hanldeChange(change)
+		}
+	}
+}
+
+func (r *router) hanldeChange(change swim.Change) {
```
jwolski: typo -> `handleChange`

router.go
```diff
@@ -0,0 +1,156 @@
...
+	r := &router{
+		ringpop:     rp,
+		factory:     f,
+		clientCache: make(map[string]interface{}),
+	}
+	rp.RegisterListener(r)
+	return r
+}
+
+func (r *router) HandleEvent(event interface{}) {
```
jwolski: Does `HandleEvent` have to be exported here?

router.go
```diff
@@ -0,0 +1,156 @@
...
+		for _, change := range event.Changes {
+			r.hanldeChange(change)
+		}
+	}
+}
+
+func (r *router) hanldeChange(change swim.Change) {
+	switch change.Status {
+	case swim.Faulty, swim.Leave:
+		// remove clients for nodes that we know are gone
```
jwolski: Comment is superfluous. The code is rather self-documenting.

ringpop.go
```diff
@@ -96,6 +96,25 @@ func mergeDefault(opts *Options) *Options {
 	return opts
 }
 
+type RingpopInterface interface {
```
jwolski: Let's think of a better name for RingpopInterface (I can't think of a better one right now). If there's something more idiomatic out there, let's use it.

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+
+    "github.com/uber/ringpop-go"
+    "github.com/uber/tchannel-go"
+    "github.com/uber/tchannel-go/thrift"
+)
+
+{{ range .AST.Services }}
+{{ $Service := . }}
+{{ $TChanInterface :=  print "TChan" $Service.Name }}
+{{ $Interface :=  print "Ringpop" $Service.Name "Adapter" }}
```
jwolski: Can you call this `$Adapter` instead of `$Interface`? I think it reads better.

router.go
jwolski: Is there a NONE option for replication? Many of the uses of Ringpop today do not require replication. And we shouldn't force our users to specify that there is or isn't. We should have a reasonable default, which is NONE.

thanodnl: One can achieve `NONE` by not configuring the thrift endpoint. This will take the zero value for the `<endpoint>Configuration` struct which is nil. This is tested in the [generated function](https://github.com/uber/ringpop-go/pull/31/files#diff-3702e77b9865700957f8e78e0de78a01R102).

Only endpoints that you want to forward need a configuration object in the construction of the adapter.

router.go
```diff
@@ -0,0 +1,156 @@
...
+		r.removeClient(change.Address)
+	}
+}
+
+// Find the destinations for a certain key with the given N and Fanout
+func (r *router) getDestinations(key string, n ReplicationFactor, fanout ReplicationFactor) []string {
+	var dests []string
```
jwolski: We should add ample debug logging (probably even stats too) into this function.

router.go
thanodnl: This implements the `ringpop.EventListener` interface, so yes this needs to be exported. It is only exported on the `router` type -which in its self is not exported- so it is not accessible on the `Router` interface returned from the constructor function.

ringpop.go
thanodnl: Best case would be to call it `Ringpop` and hide the struct behind a lower cased name so to not export the actual struct.

jwolski: How about `Interface` as @CorgiMan suggested?

ringpop.go
jwolski: Interesting. I didn't know it was already exposed. Let's avoid breaking API compatibility for for this function and ignore it for now.

router.go
jwolski: Ah, that's great! It's good not to have to expose this and only expose what's exported by `type Router interface`

router.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (r *router) handleChange(change swim.Change) {
+	switch change.Status {
+	case swim.Faulty, swim.Leave:
+		r.removeClient(change.Address)
+	}
+}
+
+// Find the destinations for a certain key with the given N and Fanout
+func (r *router) getDestinations(key string, n ReplicationFactor, fanout ReplicationFactor) []string {
```
jwolski: These replication controls were slightly confusing to me at first and that's because I thought you were going for Riak-type CAP controls. The way the N and R/W values work in Riak is as follows (from Riak docs: http://docs.basho.com/riak/1.2.1/tutorials/fast-track/Tunable-CAP-Controls-in-Riak/#A-Primer-on-N-R-and-W):

N-Value
```
All data stored in Riak will be replicated to a number of nodes in the cluster according to the N value (n_val) property set on the bucket. By default, Riak chooses an n_val of “3” for you. This means that data stored in the bucket will be replicated to three different nodes, thus storing three copies. For this to be effective, you need at least three physical nodes in your cluster. (We can, however, demonstrate its merits with local nodes.)
```

R-Value and Read Failure Tolerance
```
Riak allows the client to supply an “R value” on each direct fetch. The R value represents the number of Riak nodes which must return results for a read before the read is considered successful. This allows Riak to provide read availability even when nodes are down or laggy.
```

This is not the way you've defined these replication controls. And it also goes against the grain of how we've previously used N, R and W values in Ringpop Replicator. We _should_ get closer to Riak's definition. In the current implementation, `fanout` does not actually control the fanout factor.

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+            {{ else}}
+                errs[i] =
+            {{ end }}
+                client.{{ $PublicName }}({{ template "call" $Method }})
+        }(i, client)
+    }
+    wg.Wait()
+
+    // allow nil Reducer with only 1 client
+    if clientLength == 1 && t.{{ $cfg }}.Reducer == nil {
```
jwolski: There should be a rule that enforces the user to have a Reducer if they choose a higher fanout/replication factor than 1, maybe upon construction of a Config object?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+{{ end}}
+// ------ End of {{ $Service.Name }} ------
+{{ end}}
+
+// Forwarding helpers
+var (
+    staticForwardHeaders = map[string]string{ "forwarded": "true" }
+)
+
+func makeForwarded(ctx {{ contextType }}) {{ contextType }} {
```
jwolski: Can you come up with a better name for `makeForwarded`? It should be obvious (even reading pre-generated code) what this function is responsible for, but it's not. Would `createRequest()` or `makeRequest()` be more accurate?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+
+    var wg sync.WaitGroup
+    for i, untypedClient := range clients {
+        client := untypedClient.({{ $TChanInterface }})
+        wg.Add(1)
+        go func(i int, client {{ $TChanInterface }}) {
+            defer wg.Done()
+
+            {{ if $Method.ReturnType }}
+                rs[i], errs[i] =
```
jwolski: Are you sure this is thread-safe? Especially given the fact that these are slices? Doing some quick googling, tells me that we may not be as safe as we think we are here, but I haven't found the article yet that say mutating a slice like this is not thread-safe.

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+    {{ if $Method.ReturnType }}
+        rs := make([]{{ $Method.ReturnType | goType }}, clientLength)
+    {{ end }}
+    errs := make([]error, clientLength)
+
+    // make headers indicate a forwarded request
+    ctx = makeForwarded(ctx)
+
+    var wg sync.WaitGroup
+    for i, untypedClient := range clients {
```
jwolski: Is there any way not to make this fanout part of generated code? I ask because if we want tighter controls of N (and R/W values) like I stated above, it'd be better for us not to have that logic here, but somewhere in the Router.

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    if isForwarded(ctx) {
+        return t.impl.{{ $PublicName }}({{ template "call" $Method }})
+    }
+
+    // find the key to shard on
+    ringpopKey, err := t.{{ $cfg }}.Key({{ template "call" $Method }})
+    if err != nil {
+        return {{ if $Method.ReturnType }}r, {{ end }}fmt.Errorf("could not get key: %s", err)
```
jwolski: Since there are multiple types of errors in this function: "could not get key" + "no clients" does it make sense to give our users semantic / typed errors?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+    }
+    return t.{{ $cfg }}.Reducer({{ if $Method.ReturnType }}rs, {{end}}errs)
+}
+{{ end}}
+// ------ End of {{ $Service.Name }} ------
+{{ end}}
+
+// Forwarding helpers
+var (
+    staticForwardHeaders = map[string]string{ "forwarded": "true" }
```
jwolski: Shall we give this header a custom ringpop prefix to distinguish it from other headers? aka (Ringpop-Forwarded)

router.go
jwolski: What happens in the case where a request needs to be forwarded (aka the key does not hash to the current instance), and the user doesn't want replication? I don't think we should force users to specify N and Fanout. Those values should have sane default; in the Ringpop world today, they'd be set to `1`.

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
+{{ define "arguments" }}ctx {{ contextType }}{{ range .Arguments }}, {{ .Name }} {{ .Type | goType }}{{ end }}{{ end}}
```
jwolski: Can you drop in an example of the generated code somewhere? It'll be nice to review that side of the coin too.

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+    SetAdapter({{ $TChanInterface }})
+}
+
+func New{{ $Adapter }}(
+    impl   {{ $TChanInterface }},
+    rp     ringpop.RingpopInterface,
+    ch     *tchannel.Channel,
+    config {{ $ConfigName }},
+) *{{ $Adapter }} {
+    adapter := &{{ $Adapter }}{
```
jwolski: Are all of these params required? Would validation be appropriate here?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+        impl:    impl,
+        ringpop: rp,
+        ch:      ch,
+        config:  config,
+    }
+    // create ringpop router for routing based on ring membership
+    adapter.router = ringpop.NewRouter(rp, adapter)
+
+    // notify implementation it is adapted
+    if adaptable, ok := impl.(tchan{{ $Service.Name }}Adaptable); ok {
```
jwolski: What should happen if `!ok`? Is that valid?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+
+// ------ 'Generated' code for {{ $Method.Name | goPublicName }} ------ {{ $cfg }}
+type {{ $ConfigType }} struct {
+    N       ringpop.ReplicationFactor
+    Fanout  ringpop.ReplicationFactor
+    Reducer func({{ if $Method.ReturnType }}rs []{{ $Method.ReturnType | goType }}, {{ end }}errs []error) ({{ template "returnType" $Method}})
+    Key     func({{ template "arguments" $Method}}) (string, error)
+}
+
+func (t *{{ $Adapter }}) {{ $PublicName }} ({{ template "arguments" $Method }}) ({{ template "returnType" $Method }}) {
```
jwolski: can you rename `t` to `a`?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+// ------ Generated code for {{ $Service.Name }} ------
+type {{ $Adapter }} struct {
+    impl        {{ $TChanInterface }}
+    ringpop     ringpop.RingpopInterface
+    ch          *tchannel.Channel
+    config      {{ $ConfigName }}
+    router      ringpop.Router
+}
+
+type {{ $ConfigName }} struct {
```
jwolski: It'll be more obvious when I see the generated code, but looking at this template, it's confusing to glean meaning from the names `$ConfigName` and `$ConfigType` (below). Can you find more appropriate names?

Maybe `$ConfigName` should be `$EndpointConfigurations`?
Maybe `$ConfigType` should be `$RingpopRouterConfiguration`?

ringpop.thrift-gen
```diff
@@ -0,0 +1,170 @@
...
+    // structs for all service endpoints containing forwarding behavior
+    {{ range .Methods }}
+        {{ .Name | goPublicName }} *{{ $Service.Name }}{{ .Name | goPublicName }}Configuration
```
jwolski: Do you foresee a need where users might want to use the same configuration for multiple endpoints? Is there a way to do that now?

router.go
thanodnl: That is a good point that I have been thinking about as well. golang does not allow you to specify specific defaults for fields if they are not specified and initializes all fields to the zero value of the type. I was thinking about validating the configuration for all endpoints in the construction of the Adapter. So in this validation step we can check if `N` and `Fanout` are `0` and set them to the sane default of 1. In that case the configuration is very small and looks like this:

```go
&KeyValueGetConfiguration{
    Key: func(ctx thrift.Context, key string) (string,error) {
        return key, nil
    }
}
```

And it will work. Although since in the example from above `N` is set to 2 you still need to set that on this configuration.

router.go
thanodnl: Great thinking, I completely forgot to add stats throughout the code.

ringpop.thrift-gen
thanodnl: I will look into how slices work with concurrent access to different indices.

I was assuming it being thread safe since every `go func` is accessing its own entry in the slices. So there is no concurrent access to a given index in the slice, but it might be that te slice as a whole is accessed concurrently.

ringpop.thrift-gen
thanodnl: Unfortunately not. A couple of lines below we are relying heavily on the generation aspect of the code where we compile the actual names of the interfaces used into the code, and call the remote functions with the arguments provided in the thrift definition.

only way to do this outside of generated code would be with the use of reflection where our generated code takes the native arguments and prepare them for use in reflective calls. But I rather do not go that route because reflection is even harder to read what is happening than this code template.

ringpop.thrift-gen
thanodnl: Most of them are required. The only ones that can be omitted are `rp` and `config`, but since `config` is passed in as a value instead of a pointer you are forced to pass 1 in. The passed in config could be an empty one where all the functions will be configured to execute locally, but that defeats the purpose of using the RingpopAdapter.

The reason that you can omit (pass in a `nil`) for `rp` is because of a conversation with a user. They sometime configure the application to not use `Ringpop` and hence the internal variable to `Ringpop` will be `nil`. On could argue that the creation of the adapter can be conditional according to the value of ringpop so that rp should be a non `nil` value in here.

I will add validation around the hard required arguments. What is your take on `rp` being `nil` here?

ringpop.thrift-gen
thanodnl: No, this is a interface for convenience. If a service implementation wants to call its own public endpoints it can add a `SetAdapter` receiver to get access to the adapter for dispatching calls. This is used by a user that implements complicated application level replication with a custom fanout to other service endpoints. If no such receiver exists on the implementation the call on the line below will be ignored but that is completely valid behavior for most of the applications.

It is worth it to document this in examples though because it is a very powerful feature which is hard to find.

ringpop.thrift-gen
thanodnl: Interesting thought. But it will be an edge case in my opinion, since the only way it is technically possible to share configuration between endpoints is when the signatures of the methods are exactly the same.

To support this we need to name the structs after their signatures which makes it harder to configure in my opinion. Names like this `InStringStringOutStringErrorConfigure` is less telling to me than `KeyValueSetConfigure`.

If one can reuse a `Reducer` closure because they have the same signature it is still possible to assign it once to a variable and use that in two different Configuration structs. I expect a `Key` function to be trivial enough to write multiple times.

ringpop.thrift-gen
jwolski: I see. Your reasoning makes sense. I'm a bit iffy on the value this convenience interface has. The adapter itself is returned from this function, so the caller will have access to the adapter even without the SetAdapter "convenience". If the caller has a handle on the adapter and requires access to it, then it can manage the dependency itself. *shrug*. I suppose I'd have to see it in action.

ringpop.go
thanodnl: Thinking about this makes total sense, but it does not play nice with `len(slice)` and `LookupN`. They give compilation errors if you don't cast from `uint` back to `int` for assigning and comparison. I will keep it `int` for now to keep the code more readable.

ringpop.thrift-gen
jwolski: Yea, I know what you were assuming. And it's a reasonable assumption. I was just unsure whether there's something non-obvious about Go's slices.

The top answer in this SO question indicates that your solution is thread-safe: http://stackoverflow.com/questions/18499352/golang-cuncurrency-how-to-append-to-the-same-slice-from-different-goroutines. :+1:

jwolski: After doing some more reading and reviewing the internals of the implementation of a slice, I think what you've got here is OK. So long as you've specified the length of the slice ahead of time and aren't growing the slice, the assignment will be thread-safe.

router.go
thanodnl: Discussion outcome:

`N` => `Candidates`
`Fanout` => `Contact`

ringpop.thrift-gen
prashantv: That's a good point, since the user has a reference to the adapter, they could call `SetAdapter` themselves. Putting this in the generated code means that if the user had a Thrift method "SetAdapter", they couldn't use this.

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+
+// implement ringpop.ClientFactory
+func (a *{{ $Adapter }}) MakeClient(dest string) interface{} {
+    if dest == a.ringpop.WhoAmI() {
+        // use local client
+        return a.impl
+    }
+
+    // create remote client
+    client := thrift.NewClient(
```
prashantv: Could this logic live inside the router? Ideally, the only thing in the generated code would be the last line in this function, since nothing else is really specific to the adapter type

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (a *{{ $Adapter }}) {{ $PublicName }} ({{ template "arguments" $Method }}) ({{ template "returnType" $Method }}) {
+    // check if the function is configured to be ring aware
+    if a.{{ $cfg }} == nil {
+        return a.impl.{{ $PublicName }}({{ template "call" $Method }})
+    }
+
+    if hasForwardedHeader(ctx) {
+        return a.impl.{{ $PublicName }}({{ template "call" $Method }})
```
prashantv: optional nit: why not collapse the 2 cases?

```go
// if the function is not ring-aware, or has already been forwarded, call the local implementation.
if a.{{ $cfg }} == nil || hasForwardedHeader(ctx) {
  return a.impl.{{ $PublicName }} (...)
}
```

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    // find the key to shard on
+    ringpopKey, err := a.{{ $cfg }}.Key({{ template "call" $Method }})
+    if err != nil {
+        return {{ if $Method.ReturnType }}r, {{ end }}fmt.Errorf("could not get key: %s", err)
```
prashantv: Use `%q` for errors, not `%s` which is only intended for strings and byte slices

ringpop.thrift-gen
prashantv: This is totally safe since the length is preallocated and elements are accessed directly using the indexing operators. The same is not true of maps since you can't control whether a map setter will grow the map

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    clients := a.router.GetClients(ringpopKey, a.{{ $cfg }}.N, a.{{ $cfg }}.Fanout)
+    clientLength := len(clients)
+    if clientLength == 0 {
+        return {{ if $Method.ReturnType }}r, {{ end }}errors.New("no clients")
```
prashantv: optionally: for this error and the earlier error, it might be worth adding hte method name which returned that error. Otherwise, the remote side is going to get a very cryptic error "no clients". Try to add more information where possible to the errors, since they are the message the remote side will see.

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+    if clientLength == 1 && a.{{ $cfg }}.Reducer == nil {
+        return {{ if $Method.ReturnType }}rs[0], {{ end }}errs[0]
+    }
+    return a.{{ $cfg }}.Reducer({{ if $Method.ReturnType }}rs, {{end}}errs)
+}
+{{ end}}
+// ------ End of {{ $Service.Name }} ------
+{{ end}}
+
+// Forwarding helpers
```
prashantv: Could this code outside of the generated code?

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+var (
+    forwardedHeaderName  = "ringpop-forwarded"
+    staticForwardHeaders = map[string]string{ forwardedHeaderName: "true" }
+)
+
+func setForwardedHeader(ctx {{ contextType }}) {{ contextType }} {
+    return thrift.WithHeaders(ctx, staticForwardHeaders)
```
prashantv: If the user has custom headers on their context, this throws them away right now.

Definitely need a way to unit test this. I'd suggest create a simple example implementation, add generated code for that (which will also make reviewing the output code easier), and test using that simple implementation. You might want to take a look at how I test Thrift in tchannel:
https://github.com/uber/tchannel-go/tree/master/thrift

ringpop.thrift-gen
jwolski: Thanks for following up @prashantv :+1:

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+
+    // create remote client
+    client := thrift.NewClient(
+        a.ch,
+        a.ch.ServiceName(),
+        &thrift.ClientOptions{
+            HostPort: dest,
+        },
+    )
+    return New{{ $TChanInterface }}Client(client)
```
jwolski: I'm looking at the generated code now. Where does this type come from? Is it meant to be defined, by the user, within the same package as the generated file?

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+        wg.Add(1)
+        go func(i int, client {{ $TChanInterface }}) {
+            defer wg.Done()
+
+            {{ if $Method.ReturnType }}
+                rs[i], errs[i] =
+            {{ else }}
+                errs[i] =
+            {{ end }}
+                client.{{ $PublicName }}({{ template "call" $Method }})
```
jwolski: We should track (log/stat) the results based on expected candidate/contact values.

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+        }(i, client)
+    }
+    wg.Wait()
+
+    // allow nil Reducer with only 1 client
+    if clientLength == 1 && a.{{ $cfg }}.Reducer == nil {
+        return {{ if $Method.ReturnType }}rs[0], {{ end }}errs[0]
+    }
+    return a.{{ $cfg }}.Reducer({{ if $Method.ReturnType }}rs, {{end}}errs)
```
jwolski: We've left room for a possible nil pointer reference panic when accessing Reducer.

It might not be possible now given existing logic, but I think it's better to be safe without making any assumptions on today's behavior.

Above we're checking a condition where clientLength == 1 AND Reducer == nil. There's a possibility that clientLength == 2 AND Reducer == nil.

ringpop.thrift-gen
```diff
@@ -0,0 +1,171 @@
...
+    if a.{{ $cfg }} == nil {
+        return a.impl.{{ $PublicName }}({{ template "call" $Method }})
+    }
+
+    if hasForwardedHeader(ctx) {
+        return a.impl.{{ $PublicName }}({{ template "call" $Method }})
+    }
+
+    // find the key to shard on
+    ringpopKey, err := a.{{ $cfg }}.Key({{ template "call" $Method }})
```
jwolski: Is it enforced anywhere that Key needs to be specified? Otherwise, nil pointer.

router.go
jwolski: I still think it's a bit premature, at this phase, to add such logic. The way that Candidates/Contact is implemented below doesn't add much value. It seems that the only benefit that exists now is that if the local node is part of the candidates selection list, it'll be placed ahead in the dests list. Should we really be optimizing for this now? Do we actually know whether this'll improve performance if a greater portion of requests are handled by the local node?

router.go
```diff
@@ -0,0 +1,155 @@
...
+		fanoutInt = r.ringpop.CountReachableMembers()
+	}
+
+	// prefer WhoAmI in the lists of destinations
+	if fanoutInt < len(dests) {
+		// see if WhoAmI is after the fanout position
+		me := r.ringpop.WhoAmI()
+		for i := fanoutInt; i < len(dests); i++ {
+			if dests[i] == me {
+				dests[0] = me // use WhoAmI as the first one to fanout
```
jwolski: We need stats/logs here to detect when "local" is preferred.

ringpop.thrift-gen
prashantv: This comes from the thrift-gen generated code for TChannel itself. e.g.
https://github.com/uber/tchannel-go/blob/master/thrift/gen-go/test/tchan-test.go#L41

ringpop.go
```diff
@@ -96,6 +96,25 @@ func mergeDefault(opts *Options) *Options {
 	return opts
 }
 
+type Interface interface {
```
dansimau: Calling this `Interface` doesn't sit right with me. Mocks generated with `mockery` will give it an ambiguous name:

```
$ mockery -name Interface -output testutils/mocks/
Generating mock for: Interface
```

Also referring to the Ringpop interface within the package will mean signatures like:
```
func foo(r Interface)
```

I haven't come across any convention that sits perfectly with me, but I'd probably prefer `ringpop.RingpopInterface` over `ringpop.Interface`. Especially since third-party packages may import the interface.

dansimau: Actually, I'm coming back to `ringpop.RingpopObject`. Because that best describes what one is looking for when you see that interface in a type signature.

CorgiMan: I think I gave this suggestion. This It's used throughout the standard library, which is considered as excellent coding practice.

https://golang.org/pkg/container/heap/#Interface
https://golang.org/pkg/sort/#Interface

As a third party you see the type as ringpop.Interface which makes sense.

Admittedly I found this confusing when I came by it, but after a few moments it made sense to me. It also gives the third party a good hint that this is the most important thing in the package.

ringpop.thrift-gen
thanodnl: I was thinking about that, but since there is an early escape on top if the destination is the local node that returns `a.impl` I kept it here so far. What would be the downside of creating a client that is not being used? It will be called only once this way during the lifetime of the application though, so it might be not too bad. However you still need to pass through the `dest` next to the client.

Other option would be to add an extra method on `ringpop.ClientFactory` that gets the local implementation. Something like `GetLocalClient() interface{}`. If that makes things better.

ringpop.thrift-gen
thanodnl: I found it better readable what happens by having them separate. But since it is generated code it doesn't really matter of course.

ringpop.go
dansimau: OK. If it's in the standard library then I think that's a pretty strong argument. It may be possible to work around the issue of conflicting mock names. That or we can file a bug or PR for mockery.

dansimau: Looks like mockery has the `-inpkg` flag which will deal with this. E.g.:
```
mockery -name Interface -inpkg
```

OK, I'm good with `ringpop.Interface` :smile:

ringpop.thrift-gen
thanodnl: Not if `contextType` could be changed in `thrift-gen` in the future.

@prashantv What is the reason to use `{{ context.Type }}` over hardcoded `thrift.Context` in the original template in `thrift-gen`?

ringpop.thrift-gen
thanodnl: I added validation of the Configuration struct. The validation makes sure that if a configuration for an endpoint is provided that `Key` is set to a non-nil value.

ringpop.thrift-gen
jwolski: Yep, that's fine. Let's not build signature-specific types.

ringpop.thrift-gen
jwolski: I think `GetLocalClient()` is a really excellent idea.

ringpop.thrift-gen
jwolski: Excellent.

ringpop.thrift-gen
jwolski: @thanodnl Thoughts about this suggestion?

ringpop.thrift-gen
jwolski: Looks like this comment needs some attn. @prashantv is suggesting we're losing headers.

ringpop.thrift-gen
thanodnl: I didn't got to this yet, was figuring out other things. But now that I look into this I think it is better to name them as you suggested. Although on the first 3 reads of your earlier comment the `$EndpointConfigurations` didn't ment anything to me in the plural form. But now that I read it again and looked at the complete file it makes sense since their is an entry for every endpoint in that struct. Will rename.

ringpop.thrift-gen
prashantv: Ahh I see. It's probably fine to assume it will be `thrift.Context`

I was considering letting the user choose between `tchannel.ContextWithHeaders` and `thrift.Context`. However, it ended up just being a constant, so you can pretty safely assume it'll be `thrift.Context` I think.

ringpop_stub.go
```diff
@@ -0,0 +1,172 @@
...
+
+import (
+	"time"
+
+	"github.com/uber/ringpop-go/forward"
+	"github.com/uber/tchannel-go"
+)
+
+// Implement a fake Ringpop for testing purposes
+type ringpopStub struct {
```
dansimau: Can I suggest that we look at using [testify](https://github.com/stretchr/testify) mocks and autogenerate the mocks using [mockery](https://github.com/vektra/mockery)? It eliminates the need to create custom stubs because it generates the mocks from your interfaces.

This is what is being done in PR #33, and we use testify in TChannel and other Golang projects.

forward/forwarder_test.go
```diff
@@ -186,3 +187,47 @@ func (s *ForwarderTestSuite) TestRequestNoReroutes() {
...
+	}
+
+	ctx, _ = thrift.NewContext(0 * time.Second)
+	ctx = thrift.WithHeaders(ctx, map[string]string{
+		"keep": "this key",
+	})
+	ctx = SetForwardedHeader(ctx)
+
+	if ctx.Headers()["ringpop-forwarded"] != "true" {
+		t.Errorf("rinpop forwarding header is not set if there were headers set already")
```
jwolski: typo `rinpop` -> `ringpop`

ringpop_stub.go
thanodnl: I use `mockery` now for tests against the generated code. But when I tried to replace the stubs in the `router_test.go` I got into a circular dependency hell.

`router_test.go` would imoprt `mocks` which in turn would import the `ringpop` package where the `Router` resided. Do you have any idea in how to solve this?

prashantv: To avoid circular dependencies, rename the test package to `ringpop_test`.

That way, `ringpop_test -> mocks -> ringpop` is fine as a dependency chain.

ringpop.go
```diff
@@ -45,7 +45,7 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/apache/thrift/lib/go/thrift"
+	athrift "github.com/apache/thrift/lib/go/thrift"
```
jwolski: why athrift?

Makefile
```diff
@@ -1,3 +1,5 @@
+export PATH := $(realpath ./scripts/travis/thrift-release/linux-x86_64):$(PATH)
```
dansimau: `realpath` (part of GNU coreutils) is not installed on OS X by default:
```
$ realpath
-bash: realpath: command not found
```

I would use `$(echo $PWD)/scripts/travis/thrift-release/linux-x86_64:$(PATH)` or whatever.

router/router.go
```diff
@@ -0,0 +1,138 @@
...
+// provided
+type ClientFactory interface {
+	GetLocalClient() interface{}
+	MakeRemoteClient(client thrift.TChanClient) interface{}
+}
+
+// NewRouter creates an instance that validates the Router interface. A Router
+// will be used to get implementations of service interfaces that implement a
+// distributed microservice.
+func NewRouter(rp ringpop.Interface, f ClientFactory, ch *tchannel.Channel) Router {
```
dansimau: We need to agree on constructor names here. Shouldn't this be `router.New()`?

test/gen-mocks
```diff
@@ -9,6 +9,7 @@ mock_directory="${0%/*}/mocks"
 # Generate mocks for bark
```
dansimau: Comment seems out of date.

thanodnl: Yes! But has nothing todo with these changes ;)

.gitignore
```diff
@@ -1 +1,4 @@
 Godeps/_workspace
+.DS_Store
+testpop
```
dansimau: FYI, `testpop` exclude covered by [`scripts/testpop/.gitignore`](https://github.com/uber/ringpop-go/blob/master/scripts/testpop/.gitignore).

test/gen-mocks
dansimau: The command is rendered out-of-date by this patch...

ringpop.go
dansimau: Stands for Apache Thrift. To make a distinction between that and the Uber/TChannel thrift package.

.gitignore
thanodnl: `testpop` shows up for me in the root of the project when build with `make testpop`. Hence I added it to this `.gitignore`. Does it show up in `scripts/testpop` for you?

dansimau: Ah. Probably I was changing directory to build it. Can we remove the `scripts/testpop/.gitignore` then? It's better off in the `.gitignore` if that's the case.

PR #32, opened 2015-11-30
Add Ringpop.Ready() method

dansimau:
Returns whether or not Ringpop is bootstrapped and thus ready to receive
requests. This is analogous to the IsReady flag in ringpop-node.

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -230,6 +230,12 @@ func (rp *Ringpop) Bootstrap(opts *BootstrapOptions) ([]string, error) {
 	return joined, nil
 }
 
+// Ready returns whether or not ringpop is bootstrapped and should receive
+// requests
+func (rp *Ringpop) Ready() bool {
+	return rp.node.Ready()
+}
+
 //= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
 //
 //	SWIM Events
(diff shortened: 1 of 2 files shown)


ringpop.go
```diff
@@ -230,6 +230,12 @@ func (rp *Ringpop) Bootstrap(opts *BootstrapOptions) ([]string, error) {
 	return joined, nil
 }
 
+// Ready returns whether or not ringpop has is bootstrapped and should receive
```
jwolski: typo in comment: `has is`

ringpop_test.go
```diff
@@ -139,6 +139,22 @@ func (s *RingpopTestSuite) TestHandleEvents() {
 	s.Equal(11, listener.EventCount(), "expected 11 total events to be recorded")
 }
 
+func (s *RingpopTestSuite) TestRingpopReady() {
+	s.False(s.ringpop.Ready())
+	// Create single node cluster.
+	s.ringpop.Bootstrap(&BootstrapOptions{
+		swim.BootstrapOptions{
+			Hosts: []string{"127.0.0.1:3001"},
```
jwolski: I'm not entirely sure why this call to Bootstrap is successful. do you?

dansimau: Yes, because calling Bootstrap with a host list that contains only yourself creates a single node cluster.

jwolski: Got it. Makes sense.

jwolski:
lgtm.

PR #33, opened 2015-12-02
Change options pattern for Ringpop constructor

dansimau:
Change options pattern for Ringpop constructor

This is a breaking change which makes future options changes easier.

Instead of taking constructor options as arguments (or in a struct) we
use a modifier pattern, where each option is a function which modifies
the object being created.

This allows certain options to be ... optional.

This change is to support a future change which will make the "address"
argument optional in the constructor.

Since this is a breaking change, we also take the opportunity to rename
the constructor to ringpop.New, which is more idiomatic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,18 @@
-testpop:	clean
-	godep go build scripts/testpop/testpop.go
+.PHONY: clean testpop mocks out
+
+out:	testpop
 
 clean:
 	rm -f testpop
 
-.PHONY: clean testpop
+clean-mocks:
+	rm -f test/mocks/*.go
+
+mocks:
+	test/gen-mocks
+
+test: mocks
+	go test ./...
+
+testpop:	clean
+	godep go build scripts/testpop/testpop.go
--- a/examples/ping-ring/.gitignore
+++ b/examples/ping-ring/.gitignore
@@ -1,0 +1,1 @@
+ping-ring
--- a/examples/ping-ring/main.go
+++ b/examples/ping-ring/main.go
@@ -95,12 +95,19 @@ func main() {
 
 	logger := log.StandardLogger()
 
+	rp, err := ringpop.New("ping-app",
+		ringpop.Channel(ch),
+		ringpop.Identity(*hostport),
+		ringpop.Logger(bark.NewLoggerFromLogrus(logger)),
+	)
+	if err != nil {
+		log.Fatalf("Unable to create Ringpop: %v", err)
+	}
+
 	worker := &worker{
 		channel: ch,
-		ringpop: ringpop.NewRingpop("ping-app", *hostport, ch, &ringpop.Options{
-			Logger: bark.NewLoggerFromLogrus(logger),
-		}),
-		logger: logger,
+		ringpop: rp,
+		logger:  logger,
 	}
 
 	if err := worker.RegisterPong(); err != nil {
--- a/options.go
+++ b/options.go
@@ -35,8 +35,6 @@ type Configuration struct {
 	// that App is taken as an argument of the Ringpop constructor and not a
 	// configuration option. This is to prevent accidental misconfiguration.
 	App string
-
-	ReplicaPoints int
 }
 
 // "Options" are modifier functions that configure/modify a real Ringpop
--- a/options.go
+++ b/options.go
@@ -1,0 +1,164 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ringpop
+
+import (
+	"errors"
+	"io/ioutil"
+
+	"github.com/Sirupsen/logrus"
+	log "github.com/uber-common/bark"
+	"github.com/uber/ringpop-go/shared"
+)
+
+type Configuration struct {
+	// App is the name used to uniquely identify members of the same ring.
+	// Members will only talk to other members with the same app name. Note
+	// that App is taken as an argument of the Ringpop constructor and not a
+	// configuration option. This is to prevent accidental misconfiguration.
+	App string
+
+	ReplicaPoints int
+}
+
+// "Options" are modifier functions that configure/modify a real Ringpop
+// object.
+//
+// There are typically two types of runtime options you can provide: flags
+// (functions that modify the object) and value options (functions the accept
+// user-specific arguments and then return a function that modifies the
+// object).
+//
+// For more information, see:
+// http://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html
+//
+type Option func(*Ringpop) error
+
+// applyOptions applies runtime configuration options to the specified Ringpop
+// instance.
+func applyOptions(r *Ringpop, opts []Option) error {
+	for _, option := range opts {
+		err := option(r)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// checkOptions checks that the Ringpop instance has been properly configured
+// with all the required options.
+func checkOptions(rp *Ringpop) []error {
+	errs := []error{}
+	if rp.channel == nil {
+		errs = append(errs, errors.New("Channel is required"))
+	}
+	if rp.identityResolver == nil {
+		errs = append(errs, errors.New("Identity resolver is nil"))
+	}
+	return errs
+}
+
+// Runtime options
+
+func Channel(ch shared.TChannel) Option {
+	return func(r *Ringpop) error {
+		r.channel = ch
+		return nil
+	}
+}
+
+func HashRingConfig(c *HashRingConfiguration) Option {
+	return func(r *Ringpop) error {
+		r.configHashRing = c
+		return nil
+	}
+}
+
+func Logger(l log.Logger) Option {
+	return func(r *Ringpop) error {
+		r.logger = l
+		r.log = l
+		return nil
+	}
+}
+
+func Statter(s log.StatsReporter) Option {
+	return func(r *Ringpop) error {
+		r.statter = s
+		return nil
+	}
+}
+
+// IdentityResolver is a function that returns the listen interface/port
+// that Ringpop should identify as.
+type IdentityResolver func() (string, error)
+
+func IdentityResolverFunc(resolver IdentityResolver) Option {
+	return func(r *Ringpop) error {
+		r.identityResolver = resolver
+		return nil
+	}
+}
+
+// Identity specifies a static hostport string as Ringpop's identity.
+func Identity(hostport string) Option {
+	return IdentityResolverFunc(func() (string, error) {
+		return hostport, nil
+	})
+}
+
+// Default options
+
+// defaultIdentityResolver sets the default identityResolver
+func defaultIdentityResolver(r *Ringpop) error {
+	r.identityResolver = r.channelIdentityResolver
+	return nil
+}
+
+// defaultLogger is the default logger that is used for Ringpop if one is not
+// provided by the user.
+func defaultLogger(r *Ringpop) error {
+	return Logger(log.NewLoggerFromLogrus(&logrus.Logger{
+		Out: ioutil.Discard,
+	}))(r)
+}
+
+func defaultStatter(r *Ringpop) error {
+	return Statter(noopStatsReporter{})(r)
+}
+
+func defaultHashRingOptions(r *Ringpop) error {
+	return HashRingConfig(defaultHashRingConfiguration)(r)
+}
+
+// defaultOptions are the default options/values when Ringpop is created. They
+// can be overridden at runtime.
+var defaultOptions = []Option{
+	defaultIdentityResolver,
+	defaultLogger,
+	defaultStatter,
+	defaultHashRingOptions,
+}
+
+var defaultHashRingConfiguration = &HashRingConfiguration{
+	ReplicaPoints: 100,
+}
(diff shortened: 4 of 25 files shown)


ringpop.go
```diff
@@ -129,37 +88,77 @@ type Ringpop struct {
 	startTime time.Time
 }
 
-// NewRingpop returns a new Ringpop instance
-func NewRingpop(app, address string, channel *tchannel.Channel, opts *Options) *Ringpop {
-	opts = mergeDefault(opts)
+// NewRingpop returns a new Ringpop instance.
```
jwolski: I'd have preferred an exclamation point.

options.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/uber/ringpop-go/shared"
+)
+
+type Configuration struct {
+	// App is the name used to uniquely identify members of the same ring.
+	// Members will only talk to other members with the same app name. Note
+	// that App is taken as an argument of the Ringpop constructor and not a
+	// configuration option. This is to prevent accidental misconfiguration.
+	App string
+	// ???
```
jwolski: ??? = The number of times a Ringpop node is hashed and added to the hash ring to given better spread/distribution around the ring.

options.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Runtime options
+
+func Channel(ch shared.TChannel) Option {
```
jwolski: Why do these need to be exported?

dansimau: Outside packages will use them to configure the options. E.g.:

```
ch, _ := tchannel.NewChannel("test", nil)
rp, _ := ringpop.NewRingpop("test",
    ringpop.Channel(ch),
    ringpop.Identity("127.0.0.1:3000"),
)
ch.ListenAndServe("127.0.0.1:3000")

// etc
```

ringpop.go
```diff
@@ -129,37 +88,77 @@ type Ringpop struct {
...
+			App: app,
+		},
 	}
 
-	if channel != nil {
-		ringpop.channel = channel.GetSubChannel("ringpop", tchannel.Isolated)
-		ringpop.registerHandlers()
+	err = applyOptions(ringpop, defaultOptions)
+	if err != nil {
+		panic(fmt.Errorf("Error applying default Ringpop options: %v", err))
```
jwolski: This seems like a good use of panic(). Nice.

options.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+// IdentityResolver is a function that returns the listen interface/port
+// that Ringpop should identify as.
+type IdentityResolver func() (string, error)
+
+func IdentityResolverFunc(resolver IdentityResolver) Option {
+	return func(r *Ringpop) error {
+		r.identityResolver = resolver
```
jwolski: We should be careful about nils. In theory, a user can pass in a nil `resolver`, right?

ringpop.go
```diff
@@ -39,78 +39,37 @@ package ringpop
...
-}
-
 // Ringpop is a consistent hash-ring that uses a gossip prtocol to disseminate changes around the
 // ring
 type Ringpop struct {
-	app     string
-	address string
+	config *Configuration
+
+	identityResolver IdentityResolver
```
jwolski: Is it necessary to hang onto the resolver like this? Why choose this way over resolving identity and hanging onto that instead?

shared/interfaces.go
```diff
@@ -0,0 +1,13 @@
+package shared
```
jwolski: There is much good here. Nice job.

options.go
jwolski: yep, thx.

jwolski:
Looking good. I imagine that we're going to want to switch over to `Option`s-all-the-way down, meaning passing Options into the SWIM and Ring packages. Let's not do that now though.

Ping us when this PR is out of the WIP stage and ready for full review.

ringpop.go
dansimau: Because later on, we're going to delay resolution until we actually need it.

options.go
dansimau: Excellent question. I am writing a `checkOptions` method as we speak.

ringpop.go
jwolski: Yep, that's a good reason.

shared/interfaces.go
jwolski: Re: the `shared` package and naming... 

'Shared'/'util'-type things tend to become dumping grounds. How about making this a root-level file called `deps.go` instead? `shared.interfaces` is a bit ambiguous. We're likely to have internal interfaces defined with Ringpop that probably shouldn't go in this file. I think what the interfaces in this file represents best are interfaces for Ringpop's 3rd party dependencies.

testutils/mocks/Logger.go
```diff
@@ -0,0 +1,81 @@
+package mocks
```
jwolski: Is this auto gen'd code? Is it possible to put a note here about that? Or somewhere?

jwolski: Also, how about just a root-level mocks folder instead of `testutils/mocks`? I think mocks imply that there only testing utilities.

examples/ping-ring/main.go
```diff
@@ -95,12 +95,19 @@ func main() {
 
 	logger := log.StandardLogger()
 
+	rp, err := ringpop.New("ping-app",
+		ringpop.Channel(ch),
```
jwolski: `ringpop.Channel(...)`, `ringpop.Identity(...)` doesn't read as well as it could. I feel like these options should be qualified more clearly w/ config-related terms, for example:

```
ringpop.options.Identity(*hostport)
```

or 

```
ringpop.config.Identity(*hostport)
```

or

```
ringpop.IdentityOption(*hostport)
```

What do you think?

examples/ping-ring/.gitignore
```diff
@@ -0,0 +1 @@
+ping-ring
```
jwolski: eh?

options.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+type Configuration struct {
+	// App is the name used to uniquely identify members of the same ring.
+	// Members will only talk to other members with the same app name. Note
+	// that App is taken as an argument of the Ringpop constructor and not a
+	// configuration option. This is to prevent accidental misconfiguration.
+	App string
+
+	ReplicaPoints int
```
jwolski: Should this be removed now that it's in HashRingConfiguration?

options.go
```diff
@@ -0,0 +1,155 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// checkOptions checks that the Ringpop instance has been properly configured
+// with all the required options.
+func checkOptions(rp *Ringpop) []error {
+	errs := []error{}
```
jwolski: I feel like this function makes the assumption that our default options will always "do the right thing," meaning, if an identity resolver is required, we assume that there is a default that fills it in. Is this true? If so, I feel like this assumption is unsafe and that checkOptions should verify the existence of all required fields. Thoughts?

options.go
```diff
@@ -0,0 +1,155 @@
...
+// Runtime options
+
+func Channel(ch shared.TChannel) Option {
+	return func(r *Ringpop) error {
+		r.channel = ch
+		return nil
+	}
+}
+
+func HashRingConfig(c *HashRingConfiguration) Option {
```
jwolski: Clever, I like it.

options.go
```diff
@@ -0,0 +1,155 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ringpop
```
jwolski: The Option pattern appears really nice, but I feel like devs will have to be reminded  about what needs to be defined when a new Option comes about. Can you put instructions somewhere (at the top of this file?) saying something like:

```
An Option is a function value.

To make a new Option available, define a public Option "applier" which is a function that returns a func(*Ringpop) (error) value. This value is meant to set a field of your choice on the *Ringpop. You should/must also define a default Option applier in case a user-specified value is not provided.
```

or similar.

shared/interfaces.go
dansimau: I agree that "shared" could become a dumping ground. However, I disagree that this is for deps.

Interfaces (and currently a TChannel util function) in the shared package are for use by the ringpop, forward, replica and swim packages. These are all separate packages that are inside the ringpop repository.

There's another reason they are in a separate package and not just in ringpop: circular dependencies. Ringpop imports the swim package, which refers to the interfaces. We can't import them back from the ringpop package again. So they have to be in a separate package.

Open to suggestions re: naming.

testutils/mocks/Logger.go
dansimau: I was paving the way to move `test_utils.go` into a separate package inside `testutils/`, with the view that these test utilities should be used by all sub-packages.

examples/ping-ring/main.go
dansimau: I did initially consider putting options into their own package, which would result in an API like:

```
ringpop.New(ringpop.options.Identity("..."))
```

However in the end I opted for brevity and simplicity of layout. If you think `ringpop.Identity` and `ringpop.Channel` is confusing or unclear though, I think that's a strong case to move the options modifiers into their own package.

examples/ping-ring/.gitignore
dansimau: yeah

options.go
dansimau: Yes, well spotted, thanks!

options.go
dansimau: Agreed. Will fix.

options.go
dansimau: Good call. I haven't seen the pattern documented well anywhere. So you're right. We should document it more extensively.

options_test.go
```diff
@@ -0,0 +1,158 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ringpop
+
+import (
```
jwolski: I don't see a test that necessarily exercises cases where:

- no Logger/Statter option is provided by the user and the default is used. 
- a field is not required, no option is provided by the user, and the nil value is OK.

Did I miss that? If not, add one or two in.

dansimau: > no Logger/Statter option is provided by the user and the default is used.

Tests for default options are here:
https://github.com/uber/ringpop-go/pull/33/files#diff-4d94a0f6a8f4f5183ccea19235551c1bR46

Does that satisfy this case?

> a field is not required, no option is provided by the user, and the nil value is OK.

I don't think such a field exists. For the options we have right now, there are defaults for all cases where the user doesn't specify something. We don't have an option that has a valid nil value.

examples/ping-ring/main.go
jwolski: Agreed that we should shoot for brevity, but not where it comprises on readability/understandability. I think we're tip-toeing that line here. This pattern reminds me slightly of Scala's Options (for example, http://danielwestheide.com/blog/2012/12/19/the-neophytes-guide-to-scala-part-5-the-option-type.html) and there it's clear that you are dealing with Options.

It's not so clear what the thing is that is living in the ringpop namespace if it's just called `ringpop.Channel()`. I think the name`ringpop.options` may be too literal and that `ringpop.config.Channel()` seems the _right_ level of abstraction.

options_test.go
jwolski: > Does that satisfy this case?

Ah, yes. I missed that test case!

> I don't think such a field exists. For the options we have right now, there are defaults for all cases where the user doesn't specify something. We don't have an option that has a valid nil value.

Good call. Forget I said it. Will cross that bridge when we get there.

options_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+// TestDefaults tests that the default options are applied to a Ringpop
+// instance when none are specified by the user.
+func (s *RingpopOptionsTestSuite) TestDefaults() {
+
+	rp, err := New("test", Channel(s.channel))
+	s.Require().NotNil(rp)
+	s.Require().NoError(err)
+
+	s.NotNil(rp.logger)
```
jwolski: shall we also assert that `rp.logger` and `rp.statter` are equal to the default values?

shared/interfaces.go
jwolski: > I agree that "shared" could become a dumping ground. However, I disagree that this is for deps.
> 
> Interfaces (and currently a TChannel util function) in the shared package are for use by the ringpop, forward, replica > and swim packages. These are all separate packages that are inside the ringpop repository.

That's my point. The TChannel interface you've defined is an interface standing in as an abstraction built around a 3rd party dependency of Ringpop (including all of Ringpop's packages). `interfaces.go` is too general. I think there is a difference in these interfaces (for TChannel et. al) and interfaces that Ringpop may define and share internally. That's why I'd make the distinction between them.

> There's another reason they are in a separate package and not just in ringpop: circular dependencies...

:+1:  Good to know

Makefile
```diff
@@ -1,7 +1,15 @@
...
+GOPATH = $(shell echo $$GOPATH)
+
+out:	testpop
 
 clean:
 	rm -f testpop
 
-.PHONY: clean testpop
+mocks:
+	mockery -dir $(GOPATH)/src/github.com/uber-common/bark -name Logger -output testutils/mocks/ -recursive
```
jwolski: This generates a file like `Logger.go` and `StatsReporter.go` (with captial letters). Is this required? Can we make the generated files more lowercase-y.

testutils/mocks/Logger.go
jwolski: I see. I wasn't aware of `test_utils.go`. In general, we should strive for less utility-type packages/modules and be as specific as we can be. If we can group functions / types defined within our `test_utils.go` more specifically, I'd be for that. But let's not do that now. Carry on.

examples/ping-ring/main.go
dansimau: Just leaving a note here to say: after some discussion and prototyping, this is proving too difficult to do right now. We looked at moving `Option`s to a `config` package but this resulted in circular dependencies that would take significant time effort to resolve.

options_test.go
dansimau: Yes. Done.

Makefile
dansimau: There is no mechanism to change this in mockery. Though you can output the mocks to stdout and pipe them wherever you like.

I think for simplicity we should just stick with mockery's defaults.

examples/ping-ring/main.go
jwolski: :+1:

options.go
prashantv: +1 for documenting. Btw, the options pattern is documented here:
http://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html

Might be worth putting a link to it, if people want to read more about it.

dansimau: :+1: thanks Prashant

Makefile
dansimau: I lied. It's possible. Will fix.

PR #34, opened 2015-12-04
Add lazy initialization

dansimau:
This change delays initialization of swim node set up, stats set up and
default handler registration until bootstrapping.

Crucially, this also delays resolution of the listen address until
bootstrap is called. This means a TChannel object that is not yet
listening can be passed to the ringpop constructor.


jwolski:
Looking good!

errors.go
```diff
@@ -0,0 +1,5 @@
+package ringpop
+
+import "errors"
+
+var ErrNotBootstrapped = errors.New("Ringpop has not been bootstrapped")
```
thanodnl: Errors in golang are usually lowercased. eg: http://blog.golang.org/error-handling-and-go

forward/request_sender.go
```diff
@@ -91,8 +91,10 @@ func (s *requestSender) Send() (res []byte, err error) {
 			return s.ScheduleRetry()
 		}
 
+		identity, _ := s.sender.WhoAmI()
```
thanodnl: how likely is it that there will be an error here? I can imagine that Send already errored earlier if ringpop is not initialized. Otherwise we should take the error here and handle it.

handlers.go
```diff
@@ -59,7 +59,12 @@ type lookupResponse struct {
 }
 
 func (rp *Ringpop) adminLookupHandler(ctx json.Context, req *lookupRequest) (*lookupResponse, error) {
-	return &lookupResponse{Dest: rp.Lookup(req.Key)}, nil
+	dest, err := rp.Lookup(req.Key)
+	if err != nil {
+		return nil, err
+	} else {
+		return &lookupResponse{Dest: dest}, nil
```
thanodnl: Don't need to indent here, idiomatic go is

```go
if err != nil {
    return nil, err
}
// do useful stuff
```

forward/request_sender.go
dansimau: In this case I chose not to handle the error, because this is a log message. The result will be that the identity won't be printed to the log (because it cannot be resolved).

An alternative might be to print "Unknown identity" in the log, but as you point out, that code path is unlikely.

errors.go
jwolski: I see this in that article:

```
type SyntaxError struct {
    msg    string // description of error
    Offset int64  // error occurred after reading Offset bytes
}

func (e *SyntaxError) Error() string { return e.msg }
```

I think this error is meant to be exported.

handlers.go
dansimau: > idiomatic go is

Have you got a source/reference for this?

If the method has more than two return points, then I use the pattern you described. However, if there is only two, I prefer the if/else because I think it's clearer.

ringpop.go
```diff
@@ -164,25 +172,30 @@ func (rp *Ringpop) identity() (string, error) {
 // TChannel object on the Ringpop instance.
 func (rp *Ringpop) channelIdentityResolver() (string, error) {
 	hostport := rp.channel.PeerInfo().HostPort
+	// Check that TChannel is listening. By default, TChannel listens on an
+	// ephemeral host/port. The real port is then assigned by the OS when
+	// ListenAndService is called. If the hostport is 0.0.0.0:0, it means
+	// TChannel is not yet listening and the hostport cannot be resolved.
+	if hostport == "0.0.0.0:0" {
+		return "", fmt.Errorf("Unable to resolve valid listen address (TChannel hostport is %s)", hostport)
```
thanodnl: lower casing error. Normally I settle for lower casing everything except names (eg. `TChannel`), but we should figure out if that is the right way or that it should all be lower cased.

ringpop.go
```diff
@@ -213,6 +233,20 @@ func (rp *Ringpop) RegisterListener(l EventListener) {
 	rp.listeners = append(rp.listeners, l)
 }
 
+// getState gets the state of the current Ringpop instance.
+func (rp *Ringpop) getState() state {
+	rp.stateMutex.Lock()
```
thanodnl: Can you make this a read lock so multiple threads can read concurrently.

ringpop.go
```diff
@@ -213,6 +233,20 @@ func (rp *Ringpop) RegisterListener(l EventListener) {
...
+	defer rp.stateMutex.Unlock()
```
thanodnl: and not use defer. calls to `getState` are littered throughout the code and defer has serious performance impact.

```go
rp.stateMutex.RLock()
r := rp.state
rp.stateMutex.RUnlock()
return r
```

ringpop.go
```diff
@@ -239,6 +283,9 @@ func (rp *Ringpop) Bootstrap(opts *BootstrapOptions) ([]string, error) {
 // Ready returns whether or not ringpop is bootstrapped and should receive
 // requests
 func (rp *Ringpop) Ready() bool {
+	if rp.state != Ready {
```
thanodnl: use `getState()` so the access to `rp.state` is always guarded by the lock.

handlers.go
thanodnl: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

ringpop.go
dansimau: Yep, good idea. Thanks.

ringpop.go
dansimau: Nice, good tip to know. I'll change it. Is there a good source/reference showing the performance impact of defer?

ringpop.go
dansimau: Yes I also spotted that. Will fix. Thanks.

ringpop.go
thanodnl: Here is a blog post with some benchmarking results: http://lk4d4.darth.io/posts/defer/

thanodnl: By the way, this does not mean that you should never use defer, it gives some nice guarantees. If I remember correctly defer will also be executed if a panic happens. But accessing a single field should not fail catastrophically.

errors.go
dansimau: I think he means the error text. Thanks, noted! Will fix.

jwolski: oh word. thx

handlers.go
dansimau: Alright you win. I changed it :smile:

ringpop.go
dansimau: Fixed by `Unable` -> `unable`

ringpop.go
dansimau: Cool. Changed. Thanks.

ringpop_test.go
```diff
@@ -147,6 +153,96 @@ func (s *RingpopTestSuite) TestRingpopNotReady() {
...
+}
+
+// TestStateInitialized tests that Ringpop is in an initialized state after
+// a failed bootstrap attempt.
+func (s *RingpopTestSuite) TestStateInitialized() {
+
+	// Create channel and start listening so we can actually attempt to
+	// bootstrap
+	ch, _ := tchannel.NewChannel("test2", nil)
+	ch.ListenAndServe("127.0.0.1:0")
```
jwolski: Is there any sort of `Close()` on TChannel? This seems like it would need it.

ringpop_test.go
```diff
@@ -147,6 +153,96 @@ func (s *RingpopTestSuite) TestRingpopNotReady() {
...
+		},
+	})
+	s.NoError(err)
+
+	// Destroy
+	s.ringpop.Destroy()
+	s.Equal(destroyed, s.ringpop.state)
+}
+
+// TestDestroyIsIndempotent tests that Destroy() can be called multiple times.
```
jwolski: Idempotent not Indempotent

ringpop_test.go
dansimau: How so?

ringpop_test.go
dansimau: Good spot, thanks.

ringpop.go
```diff
@@ -347,16 +403,21 @@ func (rp *Ringpop) Lookup(key string) string {
 	rp.emit(LookupEvent{key, time.Now().Sub(startTime)})
 
 	if !ok {
-		rp.log.WithField("key", key).Warn("could not find destination for key")
-		return rp.WhoAmI()
+		identity, _ := rp.WhoAmI()
+		err := errors.New("could not find destination for key")
+		rp.log.WithField("key", key).Warn(err)
+		return identity, err
```
jwolski: Seems a little peculiar to return both identity and err. Why not just `err`?

dansimau: Brain not functioning. Will fix!

ringpop.go
```diff
@@ -164,25 +171,29 @@ func (rp *Ringpop) identity() (string, error) {
...
 
 // Destroy Ringpop
 func (rp *Ringpop) Destroy() {
-	rp.node.Destroy()
 
-	rp.state.Lock()
-	rp.state.destroyed = true
-	rp.state.Unlock()
-}
+	if rp.node != nil {
```
jwolski: Should we add a condition above?

```
if rp.getState() == destroyed {
  return;
}
```

I feel like this'd support the claim that this operation were idempotent better. Otherwise, you'd then be dependent upon whether `rp.node.Destroy()` were idempotent too.

ringpop.go
```diff
@@ -226,12 +258,23 @@ type BootstrapOptions struct {
 
 // Bootstrap starts the Ringpop
 func (rp *Ringpop) Bootstrap(opts *BootstrapOptions) ([]string, error) {
+
```
jwolski: remove blank space. It's inconsistent with other methods.

ringpop.go
```diff
@@ -334,12 +380,22 @@ func (rp *Ringpop) handleChanges(changes []swim.Change) {
...
+	return rp.ring.Checksum(), nil
 }
 
-// Lookup hashes a key to a server in the ring
-func (rp *Ringpop) Lookup(key string) string {
+// Lookup returns the address of the server in the ring that is responsible
+// for the specified key. It returns an error if the Ringpop instance is not
+// yet initialised/bootstrapped.
+func (rp *Ringpop) Lookup(key string) (string, error) {
+
```
jwolski: Remove blank line.

ringpop_test.go
jwolski: If we `ListenAndServe` on TChannel, I feel there should be a corresponding `Close()`.

dansimau: Done; added a `s.channel.Close()` in the `TearDownTest`.

ringpop.go
dansimau: This is what I had initially, but I changed it to the current code. That is, if `rp.node` is not `nil`, call `Destroy()`.

> Otherwise, you'd then be dependent upon whether rp.node.Destroy() were idempotent too.

`rp.node.Destroy()` *is* idempotent, as it should be (IMHO). The test passes. This seems like the correct behaviour to me.

ringpop_test.go
thanodnl: I would suggest to use a `defer ch.Close()` here in the case that something in the test does blow up.

ringpop.go
dansimau: Disagree. I think having a newline at the top is perfectly valid, if there is more than one block of code separated by newlines.

E.g. I would personally do:
```
func (f *Foo) Age() string {
    age := calcAge(f.dob) // No whitespace
    return age
}
```
```
func (f *Foo) SomethingComplicated() error {

    err := f.somethingComplicated
    if err != nil {
        return err
    }

    // But we should put whitespace at the top of this one because there are
    // two code blocks and it's more readable.
    return errors.New("Do you agree, Jeff?")
}
```

ringpop_test.go
dansimau: `defer` runs at the end of the method, which is not what we want.

We want it to run at the end of the test, which is what TearDownTest does. I haven't confirmed, but I'd expect this is is always run, even if the test panics.

ringpop.go
jwolski: I very much disagree.

ringpop_test.go
thanodnl: Than you should revisit line 215, because it is closing the channel at the end of the function.

dansimau: That's because it creates a custom channel at the start of that particular test.

ringpop.go
prashantv: +1 to no blank line at the beginning or end of a function. The only time I think I'd add a blank line is if you have a lot of arguments split over multiple lines, e.g.
```go

func(arg1 *somepackage.ReallyBigType, arg2 *somepackage.AnotherBigType, arg2 *somepackage.ReallyBigType,
    arg4 *somepackage.AnotherType, arg5 *somepackage.TooManyArgs) {

  // code here
}

jwolski: Yes, I contemplated this over the weekend. @dansimau I think you're in the large minority here and should remove the whitespaces.

dansimau: Yep, already did it:
https://github.com/uber/ringpop-go/commit/587528f5811bb555c549abb0c6ea23513a7d472b

jwolski: I consider this a great victory for mankind :|

jwolski:
lgtm

thanodnl:
This PR changes the signature of `WhoAmI` but does not update the `Interface`. This is partly because this branch was created before the `Interface` was on master, but this will break if ringpop is passed to its interface.

Besides, I see Travis now failing on `rake` instead of the failure on `make`. Don't know how it relates to this PR though.

thanodnl:
Travis: `Could not find .travis.yml, using standard configuration.` So I do think a rebase on master or a merge of master into this branch will help with both the `Interface` and the `rake` vs `make` issues.

dansimau:
Closing; will re-open PR against master.

PR #36, opened 2015-12-04
The ringpop.Interface was not in sync with the implementation.

thanodnl:
@dansimau I updated the interface to have the same signature as Ringpop, added a test to make sure you can assign a Ringpop instance to its interface. And as a bonus, updated the examples with the `swim.Bootstrap` change.


dansimau:
Sweet, good catch.

PR #37, opened 2015-12-04
Fix broken replica package test by providing tchannel.Format argument…

jwolski:
… to ForwardRequest method

We've got a bunch of broken tests in subpackages. Here's to fixing them.

@uber/ringpop


thanodnl:
:+1:

PR #38, opened 2015-12-05
Improve HashRing API

jwolski:
This work stemmed from me trying to wrie a benchmark for Ringpop's hash ring lookup. There I noticed the HashRing implementation exporting too much of itself and tests digging into its internals. What I wound up with was this:

* Reduction of HashRing public API
* Defining its interface, located in internal/native, which uses Go's internal package feature (I _think_ it's useful to do it this way, but not entirely sure).
* Improving tests so that they treat HashRing more like a blackbox.

Check it out.

@uber/ringpop

### DIFF
--- a/hashring_test.go
+++ b/hashring_test.go
@@ -21,9 +21,9 @@
 package ringpop
 
 import (
-	"fmt"
 	"testing"
 
+	"github.com/dgryski/go-farm"
 	"github.com/stretchr/testify/suite"
 	"github.com/uber/tchannel-go"
 )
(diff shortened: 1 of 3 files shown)


hashring_test.go
```diff
@@ -140,7 +135,7 @@ func (s *RingTestSuite) TestLookup() {
 
 	s.ring.RemoveServer("server1")
 	s.ring.RemoveServer("server2")
-	s.ring.RemoveServer(s.ring.ringpop.node.Address())
+	s.ring.RemoveServer(s.ringpop.WhoAmI())
```
thanodnl: the signature of `WhoAmI` has changed to `func () (string, error)` so this needs to be updated.

jwolski: werd. will update.

PR #39, opened 2015-12-06
Ring package

jwolski:
This PR does a couple of things all related to moving the hash ring implementation into its own package:

* The rbtree package has been folded into it since the red-black tree, at a higher-level, is more just an implementation detail of the ring itself
* Removes the dependency the ring had on Ringpop. Now it is standalone and able to fire its events as an `EventEmitter` rather than a thing that depends on Ringpop to do so for it.
* Removes the Ring* events out of the global events package and into an `events.go` source file that is part of the ring package itself. 

Test will come...

@uber/ringpop

### DIFF
--- a/hash_ring_configuration.go
+++ b/hash_ring_configuration.go
@@ -18,18 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package events
+package ringpop
 
-import "time"
-
-// An EventListener handles events given to it by the Ringpop, as well as forwarded events from
-// the SWIM node contained by the ringpop. HandleEvent should be thread safe.
-type EventListener interface {
-	HandleEvent(event interface{})
-}
-
-// A LookupEvent is sent when a lookup is performed on the Ringpop's ring
-type LookupEvent struct {
-	Key      string
-	Duration time.Duration
+type HashRingConfiguration struct {
+	ReplicaPoints int
 }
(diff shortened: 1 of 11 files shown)


hash_ring_configuration.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
thanodnl: shouldn't this file be in the `ring` package as well, since the `ReplicaPoints` are part of the ring implementation as well.

internal/native/event_listener.go
```diff
@@ -0,0 +1,22 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package native
+
+type EventListener func(interface{})
```
thanodnl: The package used to use a `HandleEvent` on an interface instead of making it a function directly. What is de benefit of defining this as a `func` instead of an `interface`.

PR #40, opened 2015-12-07
Add lazy initialization

dansimau:
This change delays initialization of swim node set up, stats set up and
default handler registration until bootstrapping.

Crucially, this also delays resolution of the listen address until
bootstrap is called. This means a TChannel object that is not yet
listening can be passed to the ringpop constructor.

See PR #34 for history.

@uber/ringpop

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,0 +1,5 @@
+package ringpop
+
+import "errors"
+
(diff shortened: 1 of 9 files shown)


thanodnl:
:+1:

PR #41, opened 2015-12-07
Automatically add self to bootstrap list

dansimau:
When bootstrapping, the swim node expects to see the local (current)
host identity in the bootstrap list. It prints a warning if not.

This change makes ringpop automatically add the current host identity to
the bootstrap list if it is missing.

More importantly, if ringpop.Bootstrap is called with no hosts (i.e. no
bootstrap file and no hosts in the bootstrap list) this will
automatically create a single-node cluster.

@uber/ringpop

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -278,7 +278,7 @@ func (rp *Ringpop) setState(s state) {
 //= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
 
 // Bootstrap starts the Ringpop
-func (rp *Ringpop) Bootstrap(opts *swim.BootstrapOptions) ([]string, error) {
+func (rp *Ringpop) Bootstrap(userBootstrapOpts *swim.BootstrapOptions) ([]string, error) {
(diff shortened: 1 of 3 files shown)


ringpop.go
```diff
@@ -286,6 +286,18 @@ func (rp *Ringpop) Bootstrap(opts *swim.BootstrapOptions) ([]string, error) {
...
+	identity, err := rp.identity()
+	if err != nil {
+		return nil, err
+	}
+
+	// Check we're in the bootstrap host list and add ourselves if we're not
+	// there. If the host list is empty, this will create a single-node
+	// cluster.
+	if !stringInSlice(opts.Hosts, identity) {
+		opts.Hosts = append(opts.Hosts, identity)
```
jwolski: this'll modify the opts.Hosts passed in by the caller.. which is not desirable. I'd recommend a copy of the hosts.

dansimau: Fixed.

jwolski:
lgtm

PR #43, opened 2015-12-08
Move hashring implementation and its tree into ring package

jwolski:
This is a reopened PR#39.

=====

This PR does a couple of things all related to moving the hash ring implementation into its own package:

The rbtree package has been folded into it since the red-black tree, at a higher-level, is more just an implementation detail of the ring itself
Removes the dependency the ring had on Ringpop. Now it is standalone and able to fire its events as an EventEmitter rather than a thing that depends on Ringpop to do so for it.
Removes the Ring* events out of the global events package and into an events.go source file that is part of the ring package itself.
Test will come...

@uber/ringpop


jwolski:
Closing this out since @CorgiMan and others are making improvements that supplant this PR.

PR #44, opened 2015-12-09
WIP: Event-based statter

jwolski:
This is a demonstration of the design of a more event-based stats emitter. A few points:

* Removes the HandleEvent function from the public API. This was a nasty thorn hanging off of our exported API. Things like this shouldn't exist.
* Stats emission is now centralized. Individual packages are meant to emit events when something "important" happens. These events could then be translated into logging, stats, etc.
* There is no centralized events package. Each package is responsible for defining its own events: `ringpope/events.go`, `ring/events.go`, `swim/events.go`, etc.
* In general, global things like `util.go`, `shared` pkg, etc should go away. There'll always be exceptions, but they should be rare.

This code doesn't work yet and this PR is not intended to be merged, but act as a demonstration of better design.

@uber/ringpop


jwolski:
Closing this out since our discussion and understanding of events within the codebase will evolve.

PR #45, opened 2015-12-10
Fix bug preventing bootstrapping from file

dansimau:
PR #41 introduced an enhancement where Ringpop will automatically add
the current/local node's identity to the Host list of
swim.BootstrapOptions.

This broke file-based bootstrapping, as the host identity was always
added to the Hosts field, even when file-based bootstrapping was used.

swim.Node reads the Host list as priority over reading a file, so this
caused file-based bootstrapping to always create a single-node cluster.

This fix checks that a File has not been specified before first adding
the node's identity to the Host list.

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -87,7 +87,7 @@ type Ringpop struct {
 
 	channel    shared.TChannel
 	subChannel shared.SubChannel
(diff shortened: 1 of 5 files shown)


jwolski:
lgtm

PR #46, opened 2015-12-11
Implement nodejs stats

thanodnl:
With this change we get `ringpop-go` in sync with the stats that are emitted in the nodejs implementation. Documentation of the stats can be found in `docs/running_ringpop.md`. This table is copied from the nodejs repo and augmented with a column indicating which stats are implemented in `ringpop-go`.


docs/running_ringpop.md
```diff
@@ -0,0 +1,84 @@
+## Monitoring
```
jwolski: Let's maintain only one copy of this doc. Source of truth for the docs is in the Node.js version.

swim/events.go
```diff
@@ -71,6 +71,23 @@ type JoinCompleteEvent struct {
 	Joined    []string      `json:"joined"`
 }
 
+// JoinFailedReason indicates the reason a join failed
+type JoinFailedReason string
+
+const (
+	// ERROR as a JoinFailedReason indicates that the join failed because of an error
```
jwolski: Is there a reason to ALL CAPS these constants? It's inconsistent with other constants that you've defined in this PR. Convention seems to be `PascalCase`

swim/events.go
```diff
@@ -140,3 +157,31 @@ type ChecksumComputeEvent struct {
 	Duration time.Duration `json:"duration"`
 	Checksum uint32        `json:"checksum"`
 }
+
+// A ChangesCalculatedEvent is sent when the disseminator generated the list of changes to send in a ping or its response
+type ChangesCalculatedEvent struct {
+	Changes []Change `json:"changes"`
```
jwolski: Is the JSON annotation necessary?

forward/events.go
```diff
@@ -0,0 +1,64 @@
+package forward
+
+// An EventListener handles events given to it by the SWIM node. HandleEvent should be thread safe.
+type EventListener interface {
+	HandleEvent(interface{})
```
jwolski: Can we refrain from polluting the exported API of a type by making an event listener a `func (interface{})` instead?

test_utils.go
```diff
@@ -91,14 +91,16 @@ func genAddresses(host, fromPort, toPort int) []string {
 	return addresses
 }
 
-func genChanges(addresses []string, status string) []swim.Change {
+func genChanges(addresses []string, statusses ...string) []swim.Change {
```
jwolski: statuses

forward/events.go
thanodnl: In that case I'd rather use the EventListener that is already in the `events` package. But since `swim` defined its own listener as well I opted for this.

I'm not really keen on passing it as a `func` in this case, especially since the `EventListener` is passed around throughout the project already. Rather keep 1 way of passing them around.

swim/events.go
thanodnl: Don't think so, but all the events in this package had them defined. Didn't really look into why, but kept it consistent here.

Happy to not add them to these events, but rather not remove them from the existing ones since I don't really know why they are on there.

forward/events.go
```diff
@@ -0,0 +1,64 @@
+package forward
+
+// An EventListener handles events given to it by the SWIM node. HandleEvent should be thread safe.
+type EventListener interface {
```
CorgiMan: I am a little bit confused about the terminology, but shouldn't this be an EventHandler?

I see an EventListener as something that you can register handlers on.

forward/events.go
CorgiMan: I agree that we should share the same interface in the entire project. I am not a fan of the state of how we handle events in ringpop-go right now so I think I will do some work on them. For now, I think it's best to use a shared interface.

jwolski: I think we're accruing debt by adding to the public/exported interfaces of our types. It's an API design smell when parts of an API are meant for only certain callers. I think we should not perpetuate bad patterns.

swim/join_sender.go
```diff
@@ -229,6 +231,10 @@ func (j *joinSender) JoinCluster() ([]string, error) {
 
 	for {
 		if j.node.Destroyed() {
+			j.node.emit(JoinFailedEvent{
```
jwolski: What pattern for event emission are you deciding on? Is there one event emitter per package? Should there be multiple emitters per package? Instead of assuming that the `node` will always be an event emitter, how about passing in an EventEmitter type?

forward/events.go
```diff
@@ -0,0 +1,64 @@
...
+	HandleEvent(interface{})
+}
+
+type eventEmitter interface {
+	emit(interface{})
```
jwolski: @dansimau had the very good idea of making this sort of thing read:

```
type event interface{}

type eventEmitter interface {
    emit(event)
}
```

dansimau: :+1:

swim/events.go
jwolski: Yea, I agree. Only add them if you know why. Leave the existing ones as is

forward/forwarder.go
```diff
@@ -109,15 +115,64 @@ func NewForwarder(s Sender, ch shared.SubChannel, logger log.Logger) *Forwarder
 	}
 }
 
+func (f *Forwarder) emit(event interface{}) {
+	for _, listener := range f.listeners {
+		go listener.HandleEvent(event)
```
CorgiMan: I'm not sure what the implications for concurrency complexity are for a line like this. Is this safe? What are the disadvantages to do it procedural?

jwolski: yea, think we should be careful about sneaky concurrency in our application. not to say that this is wrong, but depending upon the number of listeners (should be very small) and the rate of updates, we may need to be more careful about where we gorountine and where we don't. I don't see an any immediate need for this being async.

thanodnl: https://github.com/uber/ringpop-go/blob/master/ringpop.go#L251
https://github.com/uber/ringpop-go/blob/master/swim/node.go#L219

swim/join_sender.go
thanodnl: to emit events from the join sender I had a look what was already available and found: https://github.com/uber/ringpop-go/blob/master/swim/join_sender.go#L278 .

I'd rather pass in an event emitter here to be more verbose about emitting events, but for that I would like more clarity around the event emitter pattern that is used in ringpop go to see if we want to keep it and if we need to improve on this.

jwolski:
lgtm

CorgiMan:
LgtM

PR #48, opened 2015-12-18
Fix some race conditions that were found by go test —race.

thanodnl:
cc @CorgiMan


dansimau:
lgtm :+1:

CorgiMan:
Let's get the MONEY

thanodnl:
Travis crashes, but I ran the tests locally and the suite is not in a worse state than it was so I'm merging.

PR #49, opened 2015-12-28
Embed dependencies that broke travis in the past

thanodnl:
To make the testsuite more reliable we are embedding dependencies that broke travis in the past. This includes:

 - `thrift-gen`
 - The source of external dependencies, since godep broke the build at some point


dansimau:
Looks good. :+1:

thanodnl:
@uber/ringpop any objections on merging this? It makes travis again -and hopefully in the future- not crash during its test run.

jwolski:
lgtm

PR #50, opened 2015-12-28
Improved router test coverage from 66% to 97.1% with 1 untestable lin…

thanodnl:
…e because of race conditions.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -100,7 +100,9 @@ func (r *router) GetClient(key string) (interface{}, error) {
 
 	// no match so far, get a complete lock for creation
 	r.rw.Lock()
-	defer r.rw.Unlock()
+	defer func() {
+		r.rw.Unlock()
+	}()
 
 	// double check it is not created between read and complete lock
 	client, ok = r.clientCache[dest]
--- a/router/router.go
+++ b/router/router.go
@@ -100,9 +100,7 @@ func (r *router) GetClient(key string) (interface{}, error) {
 
 	// no match so far, get a complete lock for creation
 	r.rw.Lock()
-	defer func() {
-		r.rw.Unlock()
-	}()
+	defer r.rw.Unlock()
 
 	// double check it is not created between read and complete lock
 	client, ok = r.clientCache[dest]
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -19,16 +19,17 @@
 package router
 
 import (
+	"errors"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/uber/ringpop-go/swim"
 	"github.com/uber/ringpop-go/test/mocks"
 	"github.com/uber/tchannel-go"
 )
 
-// ringpop.Interface compatible struct with stubbed methods
 func TestRingpopRouterGetLocalClient(t *testing.T) {
-	// setup test
 	cf := &mocks.ClientFactory{}
 	cf.On("GetLocalClient").Return(nil)
 
@@ -44,7 +45,6 @@ func TestRingpopRouterGetLocalClient(t *testing.T) {
 }
 
 func TestRingpopRouterGetLocalClientCached(t *testing.T) {
-	// setup test
 	cf := &mocks.ClientFactory{}
 	cf.On("GetLocalClient").Return(nil)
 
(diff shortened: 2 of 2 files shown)


router/router.go
```diff
@@ -100,7 +100,9 @@ func (r *router) GetClient(key string) (interface{}, error) {
 
 	// no match so far, get a complete lock for creation
 	r.rw.Lock()
-	defer r.rw.Unlock()
+	defer func() {
```
dansimau: What does this change do?

thanodnl: Good catch, this was part of my testing yesterday to see if I could test the 1 line that is in a race condition.

```go
defer func(){ ... }()
```

is the same as 

```go
defer r.rw.Unlock()
```

But instead of calling unlock on closing it calls the closure.

router/router_test.go
```diff
@@ -44,7 +45,6 @@ func TestRingpopRouterGetLocalClient(t *testing.T) {
 }
 
 func TestRingpopRouterGetLocalClientCached(t *testing.T) {
-	// setup test
```
danielheller: agreed :)

router/router_test.go
```diff
@@ -108,3 +104,129 @@ func TestRingpopRouterMakeRemoteClientCached(t *testing.T) {
...
+}
+
+func TestRingpopRouterNotRemoveClientOnSwimSuspectEvent(t *testing.T) {
+	rp := &mocks.Ringpop{}
+	rp.On("RegisterListener", mock.Anything).Return()
+
+	router := New(rp, nil, nil).(*router)
+
+	hostport := "127.0.0.1:3000"
+	router.clientCache[hostport] = "Some client"
```
danielheller: This approach of directly manipulating the map in the tests gives me a bit of pause.  I'm not saying that we *definitely* shouldn't do this (sometimes it's a favorable tradeoff), but since the tests are in the same package, I wonder whether a method like getClientNoCreate or clientExists or something might not give us a little encapsulation that would make the tests more resilient to refactoring.  What do you think?

thanodnl: We don't need any extra methods to set a specific key on a specific value in the cache. For this we can callout to `GetClient`. But that makes this and the other tests rely on the correctness of `GetClient` which is tested elsewhere. I used to do that in the past in similar situations, but it is hard to read from the test result that the only function failing is `GetClient` opposed to the invalidation of the cache. Therefor I now fall back to manually setting the initial state via access to the private members.

If `clientCache` gets refactored it will be clear from the compile error on this line.

router/router_test.go
```diff
@@ -108,3 +104,129 @@ func TestRingpopRouterMakeRemoteClientCached(t *testing.T) {
...
+
+func TestRingpopRouterRemoveExistingClient(t *testing.T) {
+	rp := &mocks.Ringpop{}
+	rp.On("RegisterListener", mock.Anything).Return()
+
+	router := New(rp, nil, nil).(*router)
+
+	hostport := "127.0.0.1:3000"
+	router.clientCache[hostport] = "Some client"
+	assert.Equal(t, "Some client", router.clientCache[hostport])
```
jwolski: Is this strictly necessary? You're essentially asserting the behavior of a native Go map. Would you rather expose an API like `clientExists()`?

router/router_test.go
jwolski: I disagree. The tests should verify the behavior and API of the Router regardless of how the clients are to be stored internally (map or otherwise).

router/router_test.go
```diff
@@ -108,3 +104,129 @@ func TestRingpopRouterMakeRemoteClientCached(t *testing.T) {
...
+
+	rp := &mocks.Ringpop{}
+	rp.On("RegisterListener", mock.Anything).Return()
+	rp.On("Lookup", "hello").Return("", stubError)
+	rp.On("WhoAmI").Return("127.0.0.1:3000", nil)
+
+	router := New(rp, cf, nil)
+
+	_, err := router.GetClient("hello")
+	assert.Equal(t, errors.New("Ringpop not started"), err)
```
jwolski: "not started" should is inconsistent with the other ways we talk about Ringpop. "not ready" is more consistent.

router/router_test.go
```diff
@@ -108,3 +104,129 @@ func TestRingpopRouterMakeRemoteClientCached(t *testing.T) {
...
+}
+
+func TestRingpopRouterGetClientForwardWhoAmIError(t *testing.T) {
+	cf := &mocks.ClientFactory{}
+	cf.On("GetLocalClient").Return(nil)
+
+	stubError := errors.New("Ringpop not started")
```
jwolski: See point above. "not started" is not really part of the Ringpop lexicon.

router/router_test.go
thanodnl: You are completely right, this assert is useless. But as discussed on an other line I will rewrite how the cache is populated to not tamper with the internals of the router during tests. So this assert will become a necessity.

jwolski:
I hereby approve of this Github Pull Request

PR #52, opened 2015-12-29
Fix swim tests

dansimau:
This PR fixes a couple of out-of-date tests. However, the most significant change is a fix to flappy tests.

Previously, tests were binding to static ports ("127.0.0.1:3001", etc). Some tests were flapping due to previous tests already listening on these ports. Tests that require actual listening channels now listen on port 0.

Tests that don't actually require anything to be listening are now using the 192.0.2.0 network, which is defined by RFC 5737 as TEST-NET-1 and should be non-routable or at least non-existent on the internet.

### DIFF
--- a/swim/disseminator_test.go
+++ b/swim/disseminator_test.go
@@ -37,7 +37,10 @@ type DisseminatorTestSuite struct {
 
 func (s *DisseminatorTestSuite) SetupTest() {
 	s.incarnation = util.TimeNowMS()
-	s.node = NewNode("test", "127.0.0.1:3001", nil, nil)
+
+	fakeAddress := fakeHostPorts(1, 1, 1, 1)[0]
+	s.node = NewNode("test", fakeAddress, nil, nil)
+
 	s.node.memberlist.MakeAlive(s.node.Address(), s.incarnation)
 	s.d = s.node.disseminator
 	s.m = s.node.memberlist
@@ -48,7 +51,7 @@ func (s *DisseminatorTestSuite) TearDownTest() {
 }
 
 func (s *DisseminatorTestSuite) TestChangesAreRecorded() {
-	addresses := genAddresses(1, 2, 4)
+	addresses := fakeHostPorts(1, 1, 2, 4)
 
 	for _, address := range addresses {
 		s.m.MakeAlive(address, s.incarnation)
@@ -58,7 +61,7 @@ func (s *DisseminatorTestSuite) TestChangesAreRecorded() {
 }
 
 func (s *DisseminatorTestSuite) TestFullSync() {
-	addresses := genAddresses(1, 2, 4)
+	addresses := fakeHostPorts(1, 1, 2, 4)
 
 	for _, address := range addresses {
 		s.m.MakeAlive(address, s.incarnation)
--- a/swim/join_test.go
+++ b/swim/join_test.go
@@ -39,7 +39,11 @@ type JoinSenderTestSuite struct {
 }
 
 func (s *JoinSenderTestSuite) SetupTest() {
-	s.tnode = newChannelNode(s.T(), "127.0.0.1:3001")
+	// Explicitly set up a node with a fake address in the TEST-NET-1 range for
+	// join tests where we are expecting certain hostport combinations given
+	// a set of test data in the 192.0.2.1-255 range. TEST-NET-1 is defined
+	// in http://tools.ietf.org/html/rfc5737.
+	s.tnode = newChannelNodeWithHostPort(s.T(), "192.0.2.1:1")
 	s.node = s.tnode.node
 }
 
@@ -61,7 +65,10 @@ func (s *JoinSenderTestSuite) TestJoinNoBootstrapHosts() {
 }
 
 func (s *JoinSenderTestSuite) TestSelectGroup() {
-	seedBootstrapHosts(s.node, genAddresses(1, 1, 3))
+	fakeHosts := fakeHostPorts(1, 1, 2, 3)
+	bootstrapHosts := append(fakeHosts, s.node.Address())
+
+	seedBootstrapHosts(s.node, bootstrapHosts)
 
 	joiner, err := newJoinSender(s.node, nil)
 	s.Require().NoError(err, "cannot have an error")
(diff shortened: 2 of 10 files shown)


swim/join_test.go
```diff
@@ -92,20 +95,20 @@ func (s *JoinSenderTestSuite) TestSelectMultipleGroups() {
...
 	joiner, err := newJoinSender(s.node, nil)
 	s.Require().NoError(err, "cannot have an error")
 	s.Require().NotNil(joiner, "joiner cannot be nil")
 
-	group := sort.StringSlice(joiner.SelectGroup([]string{"127.0.0.1:3003", "127.0.0.1:3005"}))
+	group := sort.StringSlice(joiner.SelectGroup([]string{"192.0.2.1:3", "192.0.2.1:5"}))
 	group.Sort()
 
-	s.EqualValues([]string{"127.0.0.1:3002", "127.0.0.1:3004"}, group)
+	s.EqualValues([]string{"192.0.2.1:2", "192.0.2.1:4"}, group)
```
CorgiMan: Are these special IPs? Why has this changed?

swim/test_utils.go
```diff
@@ -137,21 +155,32 @@ func destroyNodes(tnodes ...*testNode) {
...
+//
+// The IP addresses returned are from the TEST-NET-1 block which are defined in
+// RFC 5737 as to be used for documentation and recommended to be unrouteable.
+// In practice, these addresses may also just time out.
+//
+// See:
+// http://tools.ietf.org/html/rfc5737
+// http://stackoverflow.com/questions/10456044/what-is-a-good-invalid-ip-address-to-use-for-unit-tests
+//
+func fakeAddresses(host, fromPort, toPort int) []string {
```
CorgiMan: Should this be renamed to fakeHostPorts?

swim/join_test.go
dansimau: I am using fake IP addresses here to make it clear that these tests are not doing real network operations:
https://github.com/uber/ringpop-go/blob/bugfix/swim-tests/swim/test_utils.go#L158

CorgiMan: I just saw the comment below. Maybe refer to that from here?

swim/test_utils.go
```diff
@@ -88,21 +88,39 @@ func (n *testNode) Destroy() {
...
+func newChannelNodeWithHostPort(t *testing.T, hostport string) *testNode {
+	ch, err := tchannel.NewChannel("test", nil)
+	require.NoError(t, err, "channel must create successfully")
+
+	node := NewNode("test", hostport, ch.GetSubChannel("test"), nil)
+
+	return &testNode{node, ch}
+}
+
+func genChannelNodes(t *testing.T, n int) (nodes []*testNode, hostports []string) {
```
CorgiMan: Is the second return value needed? I don't see it used anywhere. Is the hostport information available from the node itself?

swim/test_utils.go
```diff
@@ -137,21 +155,32 @@ func destroyNodes(tnodes ...*testNode) {
...
+	return hostports
 }
 
-func genAddressesDiffHosts(fromPort, toPort int) []string {
-	var hosts []string
-	for i := fromPort; i <= toPort; i++ {
-		hosts = append(hosts, fmt.Sprintf("127.0.0.%v:3001", i))
+// fakeAddressesDiffHosts returns a slice of fake IP address/port combinations
+// similar to fakeAddresses but over n unique hosts (IP addresses).
+func fakeAddressesDiffHosts(fromHost, toHost int) []string {
```
CorgiMan: Rename addresses to hostports?

CorgiMan: We can slam fakeAdresses and fakeAddressesDiffHosts together into one function
fakeHostPorts(fromHost, toHost, fromPort, toPort int)
Or
fakeHostPort(toHost, toPort int) //... since we always start from the first port/host anyway

swim/test_utils.go
dansimau: Yep, will do that.

swim/test_utils.go
dansimau: You're right, `hostports` is not actually used by any tests. I'll remove it.

swim/test_utils.go
dansimau: Good idea

swim/join_test.go
dansimau: I don't think we need to refer to the comment in every place an IP address is used. It is an IP address, the same as 127.0.0.1 or any other IP address.

The question here is surrounding the patch/change and the reasoning for this change is described in the commit message.

swim/test_utils.go
dansimau: So I have gone with `fakeHostPorts(fromHost, toHost, fromPort, toPort int)`, even though I'm not totally happy with the signature. But I think this can be improved in a later PR.

swim/join_test.go
CorgiMan: Isn't it a special IP that you use for a special reason? Maybe we can make it a constant
const fakeIPPrefix = "192.0.2" then it is sort of self documented. I dunno

dansimau: Maybe. I think we're overthinking it for now. This is just to fix the tests. We can revisit it later when we have more familiarity with the code. There is a lot to be refactored here anyway, I think.

CorgiMan: fair enough

CorgiMan:
I like two arrays better than four ints, but the rest lgtm

swim/join_test.go
danielheller: I actually agree with Wieger here; I think most people reading this code will burn mental cycles figuring out why we're using specific but unfamiliar IP addresses, and a quick explanation will save them pain.

dansimau: Can you guys suggest an appropriate comment? Should we add this comment to *all* lines that explicitly use this address as a test expectation? Maybe we just add this comment to line 42 at the top?

CorgiMan: I think moving it to the top would be good. We could refer to that comment on every mention of the IP is. You could also refer to the link http://tools.ietf.org/html/rfc5737 but I prefer the prior.

swim/ping_test.go
```diff
@@ -53,14 +54,22 @@ func (s *PingTestSuite) TestPing() {
...
-	res, err := sendPing(s.node, "127.0.0.1:3003", time.Second)
-	s.Error(err, "expected ping to pail")
+	// Create a channel with no handlers registered. Any requests to this
+	// channel should result in an error being returned immediately.
+	ch, err := tchannel.NewChannel("test", nil)
+	ch.ListenAndServe("127.0.0.1:0")
+	s.Require().NoError(err, "channel must create successfully")
+
+	res, err := sendPing(s.node, ch.PeerInfo().HostPort, time.Second)
+	s.Error(err, "expected ping to fail")
```
danielheller: While we're here, I'd suggest closing the channel when done; avoiding leaking resources may save us some pain in the future.

dansimau: Yep, good catch. Fixed in a subsequent local commit. Will push.

swim/disseminator_test.go
```diff
@@ -37,7 +37,10 @@ type DisseminatorTestSuite struct {
 
 func (s *DisseminatorTestSuite) SetupTest() {
 	s.incarnation = util.TimeNowMS()
-	s.node = NewNode("test", "127.0.0.1:3001", nil, nil)
+
+	fakeAddress := fakeHostPorts(1, 1, 1, 1)[0]
```
danielheller: (thumbsup)

danielheller:
lgtm overall!

swim/join_test.go
dansimau: Done :+1:

PR #53, opened 2015-12-30
create hashring package and move rbtree and hashring.go there

CorgiMan:
@thanodnl @danielheller

PR #54, opened 2015-12-30
Ws rbtree cleanup

CorgiMan:
@danielheller @thanodnl 

Some renaming
Rewritten iteration over red-black tree
Cleaned up and added tests for hashring

### DIFF
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -79,7 +79,7 @@ func New(hashfunc func([]byte) uint32, replicaPoints int) *HashRing {
 	}
 
 	ring.servers.byAddress = make(map[string]bool)
-	ring.servers.tree = new(RedBlackTree)
+	ring.servers.tree = &RedBlackTree{}
 	return ring
 }
 
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -45,15 +45,14 @@ type Configuration struct {
 // HashRing stores strings on a consistent hash ring. HashRing internally uses
 // a Red-Black Tree to achieve O(log N) lookup and insertion time.
 type HashRing struct {
+	sync.RWMutex
+
 	hashfunc      func(string) int
 	replicaPoints int
 
-	servers struct {
-		byAddress map[string]bool
-		tree      *RedBlackTree
-		checksum  uint32
-		sync.RWMutex
-	}
+	hasServer map[string]bool
+	tree      *RedBlackTree
+	checksum  uint32
 
 	listeners []events.EventListener
 }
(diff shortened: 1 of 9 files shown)


hashring/hashring.go
```diff
@@ -41,13 +42,15 @@ type HashRingConfiguration struct {
 	ReplicaPoints int
 }
 
+// HashRing stores strings on a consistent hash ring. HashRing internally uses
+// a Red-Black Tree to achieve O(log N) lookup and insertion time
```
dansimau: nit: missing full stop at end of sentence.

hashring/hashring.go
```diff
@@ -66,53 +69,46 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
 	r.listeners = append(r.listeners, l)
 }
 
-func NewHashRing(hashfunc func([]byte) uint32, replicaPoints int) *HashRing {
+// New instantiates and returns a new HashRing,
```
dansimau: s/,/.

hashring/hashring.go
```diff
@@ -66,53 +69,46 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
+
 	r.emit(events.RingChecksumEvent{
 		OldChecksum: old,
 		NewChecksum: r.servers.checksum,
 	})
-
-	r.servers.Unlock()
 }
 
+// AddServer adds all replicas of a server to the HashRing
```
dansimau: Full stop

hashring/hashring.go
```diff
@@ -66,53 +69,46 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
 		replicaPoints: replicaPoints,
+		hashfunc: func(str string) int {
+			return int(hashfunc([]byte(str)))
+		},
 	}
 
 	ring.servers.byAddress = make(map[string]bool)
-	ring.servers.tree = &RBTree{}
-
+	ring.servers.tree = new(RedBlackTree)
```
dansimau: Any reason for switching to `new`? I've been using the &Foo{} syntax because that is what is used elsewhere in this codebase.

hashring/hashring.go
```diff
@@ -123,43 +119,42 @@ func (r *HashRing) AddServer(address string) {
...
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
-		r.servers.tree.Insert(int(r.hashfunc([]byte(address))), server)
+		r.servers.tree.Insert(r.hashfunc(address), server)
 	}
-
-	r.servers.Unlock()
 }
 
+// RemoveServer removes all replicas of a server from the HashRing
```
dansimau: Full stop

hashring/hashring.go
```diff
@@ -123,43 +119,42 @@ func (r *HashRing) AddServer(address string) {
...
+		r.servers.tree.Delete(r.hashfunc(address))
 	}
-
-	r.servers.Unlock()
 }
 
-// adds and removes servers in a batch with a single checksum computation at the end
+// AddRemoveServers adds and removes all replicas of the given servers to the
+// HashRing. This function only computes the new checksum once, and thus is an
+// optimalization of AddServer and RemoveServer.
```
dansimau: s/optimalization/optimization

hashring/hashring.go
```diff
@@ -187,81 +181,69 @@ func (r *HashRing) AddRemoveServers(add []string, remove []string) bool {
 	return changed
 }
 
-// hasServer returns true if the server exists in the ring, false otherwise
+// HasServer returns true if the server exists in the ring, false otherwise
```
dansimau: Suggest: "HasServer returns whether or not the specified server exists in the ring." Or at least, add a full stop.

hashring/hashring.go
```diff
@@ -187,81 +181,69 @@ func (r *HashRing) AddRemoveServers(add []string, remove []string) bool {
...
-	server := r.servers.byAddress[address]
-	r.servers.RUnlock()
-
-	return server
+	defer r.servers.RUnlock()
+	return r.servers.byAddress[address]
 }
 
-func (r *HashRing) GetServers() (servers []string) {
+// GetServers returns all servers stored in the HashRing
```
dansimau: Full stop

hashring/hashring.go
```diff
@@ -187,81 +181,69 @@ func (r *HashRing) AddRemoveServers(add []string, remove []string) bool {
...
 	r.servers.RLock()
-	count := len(r.servers.byAddress)
-	r.servers.RUnlock()
-
-	return count
+	defer r.servers.RUnlock()
+	return len(r.servers.byAddress)
 }
 
+// Lookup return the owner of the given key
```
dansimau: Full stop :grin:

hashring/hashring.go
```diff
@@ -187,81 +181,69 @@ func (r *HashRing) AddRemoveServers(add []string, remove []string) bool {
...
 func (r *HashRing) Lookup(key string) (string, bool) {
-	r.servers.RLock()
-
-	iter := r.servers.tree.IterAt(int(r.hashfunc([]byte(key))))
-	if iter.Nil() {
-		r.servers.RUnlock()
+	strs := r.LookupN(key, 1)
+	if len(strs) == 0 {
 		return "", false
```
danielheller: should we return nil instead of "" while we're at it?  Would that be valid here?

hashring/hashring.go
```diff
@@ -123,43 +119,42 @@ func (r *HashRing) AddServer(address string) {
...
 }
 
+// RemoveServer removes all replicas of a server from the HashRing
 func (r *HashRing) RemoveServer(address string) {
 	if !r.HasServer(address) {
 		return
 	}
 
-	r.RemoveReplicas(address)
+	r.removeReplicas(address)
```
danielheller: What do you think about atomically removing the replicas and updating the checksums?  It would be a nice invariant to have that as long as you hold a lock, the list of servers and checksums match.

hashring/hashring.go
CorgiMan: nil is not a valid string. "" is the zero value of a string

danielheller: Indeed, thanks.

hashring/hashring.go
CorgiMan: What do you mean by "automatically"?

CorgiMan: I think that makes sense.

hashring/hashring.go
```diff
@@ -66,53 +69,46 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
 
+// Checksum returns the checksum of all stored servers in the HashRing
+// Use this value to find out if the HashRing is mutated.
 func (r *HashRing) Checksum() uint32 {
 	r.servers.RLock()
-	checksum := r.servers.checksum
-	r.servers.RUnlock()
-
-	return checksum
+	defer r.servers.RUnlock()
```
thanodnl: Is there a reason you rewrote the function to use `defer`? `defer` has a [non-negligible overhead](http://lk4d4.darth.io/posts/defer/)

hashring/hashring.go
```diff
@@ -187,81 +181,72 @@ func (r *HashRing) AddRemoveServers(add []string, remove []string) bool {
...
-// hasServer returns true if the server exists in the ring, false otherwise
+// HasServer returns true if the server exists in the ring, false otherwise
 func (r *HashRing) HasServer(address string) bool {
 	r.servers.RLock()
-	server := r.servers.byAddress[address]
-	r.servers.RUnlock()
-
-	return server
+	defer r.servers.RUnlock()
+	return r.servers.byAddress[address]
```
thanodnl: The right way to test membership of a map is:

```go
value, ok := r.servers.byAddress[address]
```

`ok` is false if the entry was not in the map. If there is a zero value in the map you will not see the difference in `value`, but you would in `ok`

hashring/hashring.go
thanodnl: a string cannot be `nil`. `""` is the right thing to return here.

hashring/rbtree.go
```diff
@@ -20,30 +20,28 @@
...
+// RedBlackTree is an implemantation of a Red Black Tree
+type RedBlackTree struct {
+	root *Node
 	size int
 }
 
-// An RBNode is a node in the RBTree
-type RBNode struct {
+// Node is a node of the RedBlackTree
+type Node struct {
```
thanodnl: Do we need to export `Node` from the `hashring` package?

hashring/hashring.go
```diff
@@ -66,53 +69,46 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
-		buffer.WriteString(";")
-	}
+	defer r.servers.Unlock()
 
+	addresses := r.getServersNoLock()
+	sort.Strings(addresses)
+	bytes := []byte(strings.Join(addresses, ";"))
 	old := r.servers.checksum
-	r.servers.checksum = farm.Fingerprint32(buffer.Bytes())
+	r.servers.checksum = farm.Fingerprint32(bytes)
```
danielheller: Do we have a test to confirm that this method is producing checksums identical to what the old one did?  Otherwise we could be in for a sad time with upgrades.

danielheller: We might want to add an explicit test of this nature if not (even hard-coding an expected checksum).

hashring/iter.go
```diff
@@ -20,146 +20,81 @@
...
 }
 
-// An RBIter iterates over nodes in an RBTree
-type RBIter struct {
-	tree      *RBTree
-	current   *RBNode
-	ancestors []*RBNode
+// NewIteratorAt instantiates and returns a new Iterator that iterates over
+// a red black from the node that owns the given value v to the last.
+func NewIteratorAt(tree *RedBlackTree, v int) *Iterator {
```
danielheller: Why is this now a function rather than a method, given that it takes a tree arg?

hashring/rbtree.go
CorgiMan: We don't, as with RedBlackTree

hashring/iter.go
danielheller: Oh, I see, nevermind.  Agreed.

hashring/hashring.go
CorgiMan: @danielheller The integration tests take care of that. Is that satisfying enough?

CorgiMan:
Closing, this pr is split in multiple prs,
https://github.com/uber/ringpop-go/pull/53
https://github.com/uber/ringpop-go/pull/56
https://github.com/uber/ringpop-go/pull/57

PR #55, opened 2015-12-30
Tests for forwarder package (100%)

thanodnl:
There are a bunch of files added to the `test` directory, containing generated code from tchannel to implement a basic `tchannel/thrift` service to test the forwarding of thrift frames.

Because `gen-mocks` now also generates the 'thrift' and `tchannel` files I renamed it to `gen-testfiles`.

Some mocks are added to the `forward` package to prevent circular dependencies. The `mocks` package imports `forward` :disappointed:.


forward/request_sender.go
```diff
@@ -145,18 +152,38 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte) <-chan error
...
 			_, arg3, _, err = raw.WriteArgs(call, []byte{0, 0}, s.request)
 		} else {
-			_, arg3, _, err = raw.WriteArgs(call, nil, s.request)
+			var resp *tchannel.OutboundCallResponse
+			_, arg3, resp, err = raw.WriteArgs(call, nil, s.request)
+
+			// check if the response is an application level error
+			if err == nil && resp.ApplicationError() {
+				// parse the json from the application level error
+				errResp := struct {
```
danielheller: There's no named definition of this type anywhere?

thanodnl: Not to my knowledge, they create the struct inline as well where it gets send. I can make a type in this file if it makes it more readable.

https://github.com/uber/tchannel-go/blob/master/json/handler.go#L156-L162

forward/forwarder_test.go
```diff
@@ -65,17 +61,43 @@ type Pong struct {
...
+	}).Return(nil, &pingpong.PingError{})
+
+	server := thrift.NewServer(channel)
+	server.Register(pingpong.NewTChanPingPongServer(thriftHandler))
 }
 
 func (s *ForwarderTestSuite) SetupSuite() {
-	s.sender = &DummySender{"127.0.0.1:3001", "127.0.0.1:3001"}
+	sender := &MockSender{}
+	sender.On("Lookup", "me").Return("127.0.0.1:3001", nil)
```
danielheller: Could we use port 0 here to avoid collisions?

forward/forwarder_test.go
```diff
@@ -145,8 +255,7 @@ func (s *ForwarderTestSuite) TestKeysDiverged() {
 func (s *ForwarderTestSuite) TestRequestTimesOut() {
 	var ping Ping
 
-	s.sender.lookup = "127.0.0.2:3001"
-	dest, err := s.sender.Lookup("some key")
+	dest, err := s.sender.Lookup("unreachable")
```
danielheller: Could this get funky if someone *is* listening on that port?  Should we listen, connect, and fail to reply?

forward/request_sender.go
danielheller: This seems fine if there's no help from the framework.

test/gen-testfiles
```diff
@@ -18,11 +23,17 @@ mockery -name Interface -print \
...
     |sed 's/_m \*NodeInterface/_m \*SwimNode/g' \
     |sed 's/type NodeInterface/type SwimNode/g' \
     > "$mock_directory"/swim_node.go
 
 mockery -case=underscore -dir router -recursive -output "$mock_directory" -name ClientFactory
+
+# make a EventListenerMock for forward (can't use on from $mock_directory because of circular dependency)
+mockery -name EventListener -dir events -recursive -print \
+    |sed 's/package mocks/package forward/g' \
+    > "${0%/*}/../forward/mock_event_listener_test.go"
```
danielheller: I must confess that I don't grok this substitution—what's it going for?

forward/forwarder_test.go
```diff
@@ -65,17 +61,43 @@ type Pong struct {
...
+		Key: "error",
+	}).Return(nil, &pingpong.PingError{})
+
+	server := thrift.NewServer(channel)
+	server.Register(pingpong.NewTChanPingPongServer(thriftHandler))
 }
 
 func (s *ForwarderTestSuite) SetupSuite() {
-	s.sender = &DummySender{"127.0.0.1:3001", "127.0.0.1:3001"}
+	sender := &MockSender{}
```
danielheller: Maybe we can rename this "ForwardingRouter" or something more... accurate when we get the chance.

danielheller:
lgtm overall, a few minor questions inline.

forward/forwarder_test.go
thanodnl: This name and the implementation is code generated by mockery based on the name of the interface that is used. I don't know if Sender is the best name for an interface here. I will have a look to see if I can come up with a better name. The `Mock` part will at least stay in the name here because of mockery.

forward/forwarder_test.go
thanodnl: Don't really know what you mean with port 0 and collisions. These lines are hardcoded responses to lookup for specific keys. So no actual listening is happening here. And most of the ports are not being listened to. Only port 3002 has a listener iirc. The listener for this port is created on line 109.

I don't know what the implications of port 0 would be. If a ring is used it is deterministic now since the ip:port's of all nodes are hardcoded.

forward/forwarder_test.go
thanodnl: Yes, because this test is actually testing the behavior when the tcp connection is not being accepted. But if you look closely it is using an ip that is not normally in use, 127.0.0.**2**. I will test and see if we can use an ip from the test block that @dansimau recently send around.

test/gen-testfiles
thanodnl: it is to work around strange mockery behavior and cyclic recursion.

Normally mockery makes the mocks in the `mocks` package. But the interface for Ringpop imports the `forward` package, preventing the `forward` package its self to import our shared mocks package.

Therefor I instruct mockery to print the generated code to `stdout` and replace the package name in the generated code with the package name I am going to write to file to.

Because the file will be written to a `_test.go` file it is only available to the `forward` package, and only when run with `go test`. During `go build` this generated file will be ignored.

forward/forwarder_test.go
dansimau: Just FYI, on Ubuntu — and probably other distros — the entire 127.0.0.0/8 block is an alias for the loopback interface.

E.g.:

```
# nc -l 0.0.0.0 3002 &
[1] 22932
# telnet 127.184.192.2 3002
Trying 127.184.192.2...
Connected to 127.184.192.2.
Escape character is '^]'.
^]
telnet> quit
Connection closed.

[1]+  Stopped                 nc -l 0.0.0.0 3002
# 
```

So if something is listening on `0.0.0.0:3002` then it will respond when you connect to `127.0.0.2:3002`.

forward/forwarder_test.go
dansimau: We should listen on port 0 if possible. It's a best-practice. Though I accept it's sometimes additional work if it means assertions must become programmatic.

forward/forwarder_test.go
thanodnl: I change this lookup to 192.0.2.128:1 which should not have anything bound to it.

forward/forwarder_test.go
danielheller: I may have placed the comment on the wrong line, but I'm referring to the use of fixed port 3002 for listening.

test/gen-testfiles
danielheller: Sorry, I understood the package adjustment—I didn't follow "${0%/*}" as a bash operation.  What exactly does that one do?

forward/forwarder_test.go
danielheller: To be clear I am not critiquing your code here.  It looked to me like the "sender" was actually more of a "router," based on a quick read.

test/gen-testfiles
thanodnl: For that I need to ask @dansimau. I copied his bash-fu from line 7.

dansimau: It gets the path part of command line string, i.e. so that the resulting path will be relative to the script location and not the current working directory.

`$0` means the command line string; `${variable_name%/*}` is a [string manipulation](http://tldp.org/LDP/abs/html/string-manipulation.html) operation that strips the shortest match of `"/*"` from the back of `$VAR`.

forward/forwarder.go
```diff
@@ -150,7 +151,10 @@ func (f *Forwarder) decrementInflight() {
 	inflight := f.inflight
 	f.inflightLock.Unlock()
 
-	f.emit(InflightRequestsChangedEvent{inflight})
+	if pre != inflight {
+		// emit a change
```
jwolski: comment unnecessary

forward/forwarder_test.go
```diff
@@ -112,14 +133,88 @@ func (s *ForwarderTestSuite) TestForward() {
...
+
+	res, err := s.forwarder.ForwardRequest(bytes, dest, "test", "PingPong::Ping", []string{"other 1"},
+		tchannel.Thrift, nil)
+	s.NoError(err, "expected request to be forwarded")
+
+	var response pingpong.PingPongPingResult
+
+	err = DeserializeThrift(res, &response)
+
+	s.Equal("127.0.0.1:3002", response.Success.Source)
```
jwolski: Be consistent with your assertion messages strings. Why provide a message string to `NoError()` above and not this?

forward/forwarder_test.go
```diff
@@ -188,6 +293,85 @@ func (s *ForwarderTestSuite) TestRequestNoReroutes() {
...
 
+func (s *ForwarderTestSuite) TestRegisterListener() {
+	listener := &EventListener{}
+	listener.On("HandleEvent").Return()
+
+	s.forwarder.RegisterListener(listener)
+	s.Assertions.Equal(1, len(s.forwarder.listeners), "Expected 1 listener to be registered")
+
+	// remove all listeners
+	s.forwarder.listeners = nil
```
jwolski: Why must this be done? Is there only one test subject (forwarder) created for all tests? That's typically not a great pattern. We should only need to clean up a test subject if there are resources like connections, etc that can leak, but not because to remove listeners.

forward/forwarder_test.go
```diff
@@ -188,6 +293,85 @@ func (s *ForwarderTestSuite) TestRequestNoReroutes() {
...
+
+	listener := &EventListener{}
+	listener.On("HandleEvent", mock.Anything).Run(func(args mock.Arguments) {
+		wg.Done()
+	}).Return()
+
+	s.forwarder.RegisterListener(listener)
+
+	// emit an empty struct
+	s.forwarder.emit(struct{}{})
```
jwolski: Shouldn't this code path be considered private and exercised as a result of public methods calling it? This test should be unnecessary.

forward/forwarder_test.go
dansimau: One reason you might do this is because the output of a failed test to `NoError` is less descriptive than to `Equal`.

`NoError` doesn't tell you the name of the thing you tried to check the error against. But `Equal` will output `"Foo" does not equal "Bar"` so you can immediately see in the output what it is failing on.

forward/forwarder_test.go
thanodnl: I'm not really sure, but I was getting unexpected events to the stub after the test was ran. If there are calls that don't have a mocked response mock/mockery panics and the test crashes.

I have validated that a new suite will be created for every test run so why the unexpected events are emitted is unclear to me.

I also tried removing all listeners in the `TearDown` routine, but this still resulted in panics.

forward/forwarder_test.go
thanodnl: This test validates that the event emitting works as expected. I added it to cover the lines in emit. 

Preferably we should move the `emit` and `RegisterListener` to it's own package and implementation that one can embed in their own types. This would allow use to reuse the paradigm easier without duplicating it all over the place. For now I want to make sure that the emitting actually emits to a registered listener.

jwolski: ok

forward/forwarder_test.go
```diff
@@ -235,3 +425,32 @@ func TestHasForwardedHeader(t *testing.T) {
 		t.Errorf("ringpop was not able to identify that the forwarded header was set in the case of alread present headers")
 	}
 }
+
+// SerializeThrift takes a thrift struct and returns the serialized bytes
+// of that struct using the thrift binary protocol. This is a temporary
+// measure before frames can forwarded directly past the endpoint to the proper
+// destinaiton.
```
jwolski: typo -> `destination`

forward/forwarder_test.go
```diff
@@ -235,3 +425,32 @@ func TestHasForwardedHeader(t *testing.T) {
 		t.Errorf("ringpop was not able to identify that the forwarded header was set in the case of alread present headers")
 	}
 }
+
+// SerializeThrift takes a thrift struct and returns the serialized bytes
+// of that struct using the thrift binary protocol. This is a temporary
+// measure before frames can forwarded directly past the endpoint to the proper
```
jwolski: type -> `frames can be`

forward/forwarder_test.go
```diff
@@ -235,3 +425,32 @@ func TestHasForwardedHeader(t *testing.T) {
...
+
+	if err := transport.Flush(); err != nil {
+		return nil, err
+	}
+	return buffer.Bytes(), nil
+}
+
+// DeserializeThrift takes a byte slice and attempts to write it into the
+// given thrift struct using the thrift binary protocol. This is a temporary
+// measure before frames can forwarded directly past the endpoint to the proper
```
jwolski: same typos in this comment

jwolski:
other than the typos, lgtm.

PR #57, opened 2015-12-31
Ws hashring locks cleanup

CorgiMan:
### First commmit
computeChecksum logic now uses strings.Join in stead of joining strings manually.

Defer unlocks
put locks on Exported methods

flatten servers struct into HashRing struct

### Second commmit
moved some functions around which has been another sad lesson in refactoring in push requests.

### Third commit (100% coverage)
test events


hashring/hashring.go
```diff
@@ -41,16 +42,17 @@ type HashRingConfiguration struct {
...
+	hashfunc      func(string) int
 	replicaPoints int
 
-	servers struct {
-		byAddress map[string]bool
-		tree      *RedBlackTree
-		checksum  uint32
-		sync.RWMutex
-	}
+	servers  map[string]struct{}
```
danielheller: I liked "byAddress" more—it encodes the type and purpose of the variable.

danielheller: Perhaps serversByAddress, not to get too java-y...

hashring/hashring.go
```diff
@@ -66,119 +68,77 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
 }
 
-// inserts server replicas into ring
-func (r *HashRing) addReplicas(server string) {
-	r.servers.Lock()
-	r.servers.byAddress[server] = true
+// LookupN returns the N servers that own the given key. Duplicates in the form
+// of virtual nodes are skipped to maintain a list of unique servers. If there
+// are less servers then N, we simply return all existing servers.
+func (r *HashRing) LookupN(key string, n int) []string {
```
danielheller: Why reorder all the functions?  There may actually be a good reason (not asserting otherwise), but it makes the diff harder to follow so should be well-motivated.

hashring/hashring.go
```diff
@@ -187,71 +147,90 @@ func (r *HashRing) AddRemoveServers(add []string, remove []string) bool {
 	return changed
 }
 
-// hasServer returns true if the server exists in the ring, false otherwise
-func (r *HashRing) HasServer(address string) bool {
-	r.servers.RLock()
-	server := r.servers.byAddress[address]
-	r.servers.RUnlock()
+// addServer adds all replicas of a server to the HashRing.
```
danielheller: I think it would be desirable to have a comment indicating that the lock must be held in front of every method which has that precondition.

hashring/hashring.go
danielheller: Oh, I see, this is a set, sorry.

danielheller:
I think we can do better than "xxx changed" in our commit messages.  Why did it change?  How did it change?

danielheller:
This is going in the right direction, I think.  I'll take a look right away if you drop me a line after addressing these comments.

hashring/hashring.go
CorgiMan: If you look at the three commits one by one it should be easy to follow.

danielheller: That does help.  Does the second commit contain exclusively moves of unmodified code?

danielheller: I think we should explain this explicitly the commit message as well, so git blamers have a clear understanding of what's happening and know that this commit could not be the cause of a bug, even though all the lines in the file show this commit as the last commit.  Something like, "Reorganize methods in file to more intuitive order (no functional changes)."

CorgiMan:
@jwolski, @thanodnl @danielheller 
We had a discussion about this and I chose this route because it's safer and more readable. I believe we came to the conclusion that if we see performance problems we change it then and only there were it is necessary. I believe Nils said that it's a 3x performance loss. I am not sure how impactful that is on locking. Do you agree/disagree? I'd like to hear your thoughts on this.

jwolski:
@CorgiMan I am in favor of using traditional Lock/Unlock especially in cases where the defer is only used in place of wrapping a single statement that we know is safe.

dansimau:
> We had a discussion about the defer locks being less performant and I chose this route because it's safer and more readable. I believe we came to the conclusion that if we see performance problems we change it then and only there were it is necessary.

I actually think the output of that discussion was something like: "use defer for readability, unless it is a single statement, or where performance is a concern".

If it's a single statement it's clear to see the unlock, so a defer is not required. Also, performance does matter here I think.

hashring/hashring.go
thanodnl: I had some questions about using a `slice` or a `map` to implement a set, but gladly we have people on the internet that had those questions first! http://blog.dubbelboer.com/2015/08/15/slice-vs-map.html

a `map` seems to be the right choice here.

PR #58, opened 2015-12-31
Ws iter and rbtree cleanup

CorgiMan:


hashring/rbtree.go
```diff
@@ -259,11 +247,42 @@ func (n *RBNode) search(val int) (string, bool) {
...
+// LookupNAt iterates through the tree from the node with value val, and
+// returns the next n unique strings. This function is not guaranteed to
+// return n strings.
+func (t *RedBlackTree) LookupNAt(n int, val int, unique map[string]bool) {
+	dfs(t.root, n, val, unique)
+}
+
+// dfs is a depth-first-search that finds n unique strings
+// with a value bigger or equal than val
+func dfs(node *Node, n int, val int, unique map[string]bool) {
```
danielheller: Could this be named something more descriptive than "dfs"?  Perhaps a name that captures the face that it's returning multiple values?

hashring/rbtree.go
```diff
@@ -20,47 +20,35 @@
...
-func (t *RBTree) Size() int {
+// Size returns the number of nodes in the RedBlackTree
+func (t *RedBlackTree) Size() int {
 	return t.size
 }
 
-// Child returns the left or right node of the RBTree
-func (n *RBNode) Child(right bool) *RBNode {
+// Child returns the left or right node of the RedBlackTree
+func (n *Node) Child(right bool) *Node {
```
danielheller: This package now exports a type called hashring.Node, which feels a little off to me (since it's not really a "HashRing node" so much as it is an implementation detail).  Should it be private to the package?

hashring/rbtree.go
```diff
@@ -259,11 +247,42 @@ func (n *RBNode) search(val int) (string, bool) {
...
+// returns the next n unique strings. This function is not guaranteed to
+// return n strings.
+func (t *RedBlackTree) LookupNAt(n int, val int, unique map[string]bool) {
+	dfs(t.root, n, val, unique)
+}
+
+// dfs is a depth-first-search that finds n unique strings
+// with a value bigger or equal than val
+func dfs(node *Node, n int, val int, unique map[string]bool) {
+	if len(unique) >= n || node == nil {
```
danielheller: I think it looks a bit funny to the reader to have two recursive calls both operating on node.right.  Is it cleaner if we do something more like the following?

```
 func dfs(node *Node, n int, val int, unique map[string]bool) {
 	if len(unique) >= n || node == nil {
 		return
 	}
 
 	// don't ivestigate left branch when all values there are smaller than the
 	// target value
 	if node.val > val {
	 	dfs(node.left, n, val, unique)
 	}

	if len(unique) >= n {
		return
	}

 	unique[node.str] = true
        dfs(node.right, n, val, unique) 
 }
```

hashring/hashring.go
```diff
@@ -215,53 +215,43 @@ func (r *HashRing) ServerCount() int {
...
 }
 
+// Lookup returns the owner of the given key.
 func (r *HashRing) Lookup(key string) (string, bool) {
-	r.servers.RLock()
-
-	iter := r.servers.tree.IterAt(int(r.hashfunc([]byte(key))))
-	if iter.Nil() {
-		r.servers.RUnlock()
+	strs := r.LookupN(key, 1)
```
danielheller: I like the move to use just one call for all the lookups.

danielheller: Actually, wait, I'm not sure I do.  If the key isn't present in the map, won't we still get a result this way?

danielheller: OK yes, I do like it more, sorry. :)

hashring/hashring.go
```diff
@@ -215,53 +215,43 @@ func (r *HashRing) ServerCount() int {
...
-
 	r.servers.RLock()
-	var servers []string
-	var unique = make(map[string]bool)
+	defer r.servers.RUnlock()
 
-	iter := r.servers.tree.IterAt(int(r.hashfunc([]byte(key))))
-	if iter.Nil() {
-		r.servers.RUnlock()
+	if n >= r.ServerCount() {
```
danielheller: Nice optimization!

hashring/rbtree.go
```diff
@@ -85,25 +73,25 @@ func singleRotate(oldroot *RBNode, dir bool) *RBNode {
...
-func doubleRotate(root *RBNode, dir bool) *RBNode {
+func doubleRotate(root *Node, dir bool) *Node {
 	root.setChild(!dir, singleRotate(root.Child(!dir), !dir))
 	return singleRotate(root, dir)
 }
 
 // Insert inserts a value and string into the tree
 // Returns true on succesful insertion, false if duplicate exists
-func (t *RBTree) Insert(val int, str string) (ret bool) {
+func (t *RedBlackTree) Insert(val int, str string) (ret bool) {
```
danielheller: Likewise, should the RedBlackTree be public or an implementation detail of the "hash ring"?

danielheller:
lgtm modulo the questions asked inline.

hashring/hashring.go
CorgiMan: I don't understand. You prefer this to only have LookupN and not Lookup?

hashring/rbtree.go
CorgiMan: Good point, I'll make it private, as with the RedBlackTree.

hashring/rbtree.go
CorgiMan: agreed

hashring/rbtree.go
CorgiMan: What do you mean with returning multiple values, the function doesn't return any value?

We can name it searchNUniqueBiggerThan. I think I'll swap the val and the n parameters as well.

hashring/hashring.go
danielheller: Sorry, I was just saying that I like what you've done, I think it is an improvement.

hashring/hashring.go
```diff
@@ -215,53 +215,43 @@ func (r *HashRing) ServerCount() int {
...
-
-	r.servers.RUnlock()
-
-	return server, true
+	return strs[0], true
 }
 
+// LookupN returns the N servers that own the given key. Duplicates in the form
+// of virtual nodes are skipped to maintain a list of unique servers. If there
+// are less servers then N, we simply return all existing servers.
```
jwolski: `then` -> `than`

hashring/rbtree.go
```diff
@@ -259,11 +247,42 @@ func (n *RBNode) search(val int) (string, bool) {
...
+}
+
+// dfs is a depth-first-search that finds n unique strings
+// with a value bigger or equal than val
+func dfs(node *Node, n int, val int, unique map[string]bool) {
+	if len(unique) >= n || node == nil {
+		return
+	}
+
+	// don't ivestigate left branch when all values there are smaller than the
```
thanodnl: `investigate`

thanodnl:
@danielheller did a good job with his comments already. So after those are fixed it will lgtm.

hashring/hashring.go
```diff
@@ -215,53 +215,43 @@ func (r *HashRing) ServerCount() int {
...
 func (r *HashRing) LookupN(key string, n int) []string {
-	serverCount := r.ServerCount()
-	if n > serverCount {
-		n = serverCount
-	}
-
 	r.servers.RLock()
-	var servers []string
-	var unique = make(map[string]bool)
+	defer r.servers.RUnlock()
```
jwolski: Undo this use of defer since it's a particular code hotspot.

jwolski:
I'm OK with the rest of these changes, but I'd like to run benchmark on the lookup changes to make sure they're not drastically worse.

CorgiMan:
I added a benchmark for lookup and it seems to be more than ten times faster on a tree with 1000 servers with 100 vnodes each where we lookup k = rand()%100 servers for 100 different keys.
```
(Non recursive) BenchmarkHashRingLookupN 100 22475834 ns/op
(Recursive) BenchmarkHashRingLookupN   1000 1966686 ns/op
```

CorgiMan:
Don't merge please! Changes to rbtree_test.go are gone...

CorgiMan:
k=100 or 10 I don't see a 10x difference. It's a bit faster than the old way but not. The new way is specifically faster for smaller k.

CorgiMan:
I found out that with k=0 the old version traversed the entire tree I think. Therefor it skewed the results of the test. The new version does not have this bug

jwolski:
Interesting find. Go ahead and merge.

PR #59, opened 2016-01-05
Cleanup and increased test coverage of swim package

dansimau:
The aim of this PR is to improve tests for the swim package.

However, this PR also includes some refactoring to the way bootstrapping works. `File` and `Hosts` fields in the `BootstrapOptions` become deprecated in favour of a `HostProvider` interface (though the old options remain for backwards-compatibility).

@uber/ringpop


swim/join_sender.go
```diff
@@ -65,6 +65,10 @@ type joinOpts struct {
 	size              int
 	maxJoinDuration   time.Duration
 	parallelismFactor int
+
+	// hostProvider is the HostProvider that this joinSender will use to
+	// enumerate bootstrap hosts.
+	hostProvider HostProvider
```
dansimau: I am thinking this should be renamed to `discoverProvider` / `DiscoverProvider`.

swim/node_bootstrap.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// JSONFileHostProvider is a HostProvider that reads a list of hosts from a
+// JSON file.
+type JSONFileHostProvider struct {
+	filePath string
+}
+
+// NewBootstrapFileHostProvider returns a new JSONFileHostProvider.
+func NewJSONFileHostProvider(filePath string) (*JSONFileHostProvider, error) {
```
dansimau: Should this constructor be removed? It's not strictly necessary. Also there is no possible error condition. In retrospect, this constructor seems pointless.

swim/join_sender.go
jwolski: "Host" is definitely the wrong word. We should move away from it for bootstrap-related code. I'd prefer the word Node since it somewhat abstracts away the details of a Ringpop Node mapping directly to a core or a host or a container or otherwise.

swim/join_sender.go
```diff
@@ -112,6 +130,9 @@ func newJoinSender(node *Node, opts *joinOpts) (*joinSender, error) {
 		node: node,
 	}
 
+	// Parse bootstrap hosts into a map
```
jwolski: This step seems like it would be very easy for a programmer to forget to perform. Parsing bootstrap hosts seems like it should be a detail of the HostProvider, don't you think?

swim/join_sender.go
dansimau: Good point. In the magical utopia, we might not even be connecting via TCP, so it might not even be a host:port.

swim/handlers_test.go
```diff
@@ -0,0 +1,222 @@
...
+	errTest := errors.New("test error")
+
+	logger.On("WithField", "error", errTest).Return(logger)
+	logger.On("Info", []interface{}{"error occurred"})
+
+	originalLogger := s.testNode.node.log
+	s.testNode.node.log = logger
+
+	s.testNode.node.errorHandler(s.ctx, errTest)
+	logger.AssertExpectations(s.T())
```
jwolski: This feels over-specified to me. I'd prefer that we assert that an error handler of our choice was invoked rather than depending upon a call to a logger. How about adding an API to swim.Node that would allow us to install an error handler of some kind?

swim/join_sender.go
dansimau: Possibly, although the fact that it needs to be parsed into a map is an implementation detail of the `joinSender`. This implementation of joinSender does special peer selection based on host and by having it in a map life becomes easier for the joinSender.

Conceptually, it doesn't really matter to the discoverProvider how these nodes are grouped so I don't think it should be a concern of the discoverProvider.

swim/handlers_test.go
```diff
@@ -0,0 +1,222 @@
...
+// protocol tick to occur, which means the number of ticks should have
+// incremented by one.
+func (s *HandlerTestSuite) TestTickHandler() {
+	s.testNode.node.Stop()
+	protocolPeriods := s.testNode.node.gossip.protocol.numPeriods
+
+	ping, err := s.testNode.node.tickHandler(s.ctx, &arg{})
+	s.NoError(err, "calling handler should not result in error")
+
+	s.Equal(protocolPeriods+1, s.testNode.node.gossip.protocol.numPeriods, "protocol period count should have increased by one")
```
jwolski: I'd strongly prefer that we eliminate violations of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) as much as possible. `s.testNode.node.gossip.protocol.numPeriods` feels way too invasive.

swim/join_sender.go
```diff
@@ -73,6 +77,10 @@ type joinSender struct {
 	node    *Node
 	timeout time.Duration
 
+	// bootstrapHostsMap is a map of unique hosts each containing a slice of
+	// the instances (hostsports) on that particular host.
+	bootstrapHostsMap map[string][]string
```
jwolski: Grouping by host should be an implementation detail of a Bootstrap Selector-type-of-abstraction.

swim/handlers_test.go
dansimau: Can you suggest an alternative method for testing this specific handler that doesn't violate this ideal?

What is the negative impact of violating this in a test?

swim/join_sender.go
dansimau: What do you mean by "Bootstrap Selector-type-of-abstraction"? Are you referring to the hostProvider/discoverProvider?

See my comment here:
https://github.com/uber/ringpop-go/pull/59#discussion_r48941935

Right now, the grouping of hosts is a concern of the JoinSender. If we were to push this concern to the discoverProvider, we would need to push the entire selection algorithm (i.e. which nodes to contact first) there as well, no? Which doesn't seem right, because that algorithm is dependent on local information about the current node (e.g. do I want to contact peers that are on the same host as me, or not).

swim/handlers_test.go
jwolski: First thing that comes to mind as an alternative would be for the tick handler to return or respond with the data that is required by the assertion. The negative impact of violating is that code becomes more brittle. You have a handlers test that will break if the contract of `node`, `gossip`, or `protocol` were to ever change. In other words, you taken on 3 more dependencies (at one time) in this test when you could potentially take on only one, the contract of `node` or node's tickHandler.

dansimau: Good point. Will fix.

swim/join_sender.go
jwolski: OK. Your perspective is clearer. I was coming from the perspective that the join sender and its fan-out is a separate concern from how the join sender selects nodes from the bootstrap list. The process of grouping by host would be encapsulated within the "Bootstrap Selector-type-of-abstraction."

I think it would be fine for the DiscoverProvider to require that some context (like the identity of the node calling it) would be passed to it. If you want to leave the selection concern out of the DiscoverProvider for now, I'd be OK with that.

swim/join_sender.go
jwolski: I haven't thought too much about how the concerns of a join sender would be separated. It /was/ clear to me that the fan-out sending/aggegration were one thing and who you're fanning-out to was another. But now, I'm not so sure that's true. Maybe they're more inter-related than I originally thought. I was thinking that a Selector abstraction would provide a sort of Iterator API like `NextGroup()` and the join sender could use it until "joinSize" were satisfied. You don't have to address these refactoring concerns now.

dansimau: Agreed, I think that's a good pattern. I'm gonna note that down as a TODO item for Refactor Week™.

swim/node_bootstrap.go
jwolski: Seems unnecessary to me 2. Though I'm unaware whether declaring a constructor like this one is idiomatic and whether it's preferred to define a placeholder like this constructor for eventually encapsulating more complex logic.

swim/handlers_test.go
dansimau: Based on our conversation yesterday, I focused on adding test coverage for significant branches. I stopped trying to test all the branches that were merely logging.

However, this test has already been written so I figure it might as well stay. Do you agree?

swim/handlers_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/uber/ringpop-go/shared"
+	"github.com/uber/ringpop-go/swim/test/mocks"
+	"github.com/uber/tchannel-go"
+)
+
+type HandlerTestSuite struct {
+	suite.Suite
+
+	testNode *testNode
+	peers    []*testNode
```
jwolski: The word peers stems more from the TChannel lexicon than the Ringpop. It's rarely used, if ever, in ringpop-node. And that's intentional so as not to make the concept of peers in TChannel the same as Ringpop. To be consistent with history and ringpop-node, can you call rename peers to nodes or make `cluster` itself destroyable?

That said, `peers` does not appear to be used for anything other than to contain `testNode`. Can we get rid of `peers` altogether?

swim/handlers_test.go
```diff
@@ -0,0 +1,205 @@
...
+			node.Destroy()
+		}
+	}
+	if s.cluster != nil {
+		s.cluster.Destroy()
+	}
+}
+
+func (s *HandlerTestSuite) TestGossipStartHandler() {
+	s.testNode.node.gossip.Stop()
```
jwolski: This is another example of handler tests that I do not necessarily agree with. I don't want you to change it if it will add more time to completion, but I want to make note of it. I feel the test being performed here should be defined where the `gossip` object is the subject under test (SUT), aka `swim/gossip_test.go`. 

Nevertheless, it's good to have coverage here. And I appreciate that you've done the work. :+1:

swim/handlers_test.go
```diff
@@ -0,0 +1,205 @@
...
+	// Toggle gossip off
+	_, err = s.testNode.node.gossipHandler(s.ctx, &arg{})
+	s.NoError(err, "calling handler should not result in error")
+	s.True(s.testNode.node.gossip.Stopped())
+}
+
+func (s *HandlerTestSuite) TestAdminLeaveJoinHandlers() {
+	// Join test node to cluster
+	s.cluster.Add(s.testNode.node)
+	s.Require().Equal(5, s.testNode.node.CountReachableMembers(), "expect a cluster of 5")
```
jwolski: I know that you didn't define this method, but I want to call out some inconsistencies in our codebase for future reference. In `node.go` we define a function named `CountReachableMembers()`. In `hashring.go`, we define `ServerCount()`. We should go with `Count` being the prefix or the suffix, but not both.

swim/handlers_test.go
jwolski: It's fine. Let's move on from this. We'll address next time around. Thanks for the "ack."

swim/handlers_test.go
```diff
@@ -0,0 +1,205 @@
...
+	s.NoError(err, "calling handler should not result in error")
+
+	s.Equal(s.testNode.node.memberlist.Checksum(), ping.Checksum, "checksum should be returned by tickHandler")
+}
+
+// TestJoinHandlerError tests that the join handler returns an error if the
+// join request is invalid.
+func (s *HandlerTestSuite) TestJoinHandlerError() {
+	// Construct an invalid joinRequest
+	req := &joinRequest{}
```
jwolski: Generally speaking, I think we need to do a better job of building self-documenting fixtures. This is a place where I feel a function or fixture can exist that explicitly defines an "invalid request" rather than the comment stating so and the line beneath it. I'm not sure what idiomatic fixtures in Go look like though. So let's postpone and discuss at a later date.

swim/handlers_test.go
dansimau: Ah, yes, I'll remove `peers` here. Note that I copied and pasted the test suite in `bootstrap_test.go`, so the user of the term `peers` was there already. It's also used in other places in the `swim` and `replica` packages.

swim/handlers_test.go
dansimau: Understand where you're coming from. If I remove the lines that are delving into `s.testNode.node.gossip`, then I'm just testing that the handler always returns an empty arg struct. Is that would you would do?

swim/handlers_test.go
dansimau: Agreed

swim/handlers_test.go
dansimau: :+1:

swim/node.go
```diff
@@ -296,27 +290,36 @@ func (n *Node) Ready() bool {
...
+	// Minimum number of nodes to join to satisfy a bootstrap.
 	JoinSize int
 
-	// Maximum time to attempt joins before joining cluster times out
+	// Maximum time to attempt joins before the entire boostrap process times
+	// out.
 	MaxJoinDuration time.Duration
 
-	// number of nodes to attempt to join
+	// Number of nodes to attempt to join at once.
```
jwolski: This comment is slightly in correct insofar as it not matching what the Node.js version does. ParallelismFactor is actually a _multiple_ applied to `JoinSize` to determine the number of nodes that will receive a join request.

swim/node.go
```diff
@@ -331,22 +334,11 @@ func (n *Node) Bootstrap(opts *BootstrapOptions) ([]string, error) {
...
-
-	err := util.CheckLocalMissing(n.address, n.bootstrapHosts[util.CaptureHost(n.address)])
-	if err != nil {
-		n.log.Warn(err.Error())
-	}
-
-	mismatched, err := util.CheckHostnameIPMismatch(n.address, n.bootstrapHosts)
+	// This exists to resolve the bootstrap hosts provider implementation from
+	// the deprecated "File" and "Hosts" options in BootstrapOptions.
+	discoverProvider, err := n.getDiscoverProvider(opts)
```
jwolski: I think Go generally discourages `getXXX` prefixes: https://golang.org/doc/effective_go.html#Getters

swim/node_bootstrap.go
```diff
@@ -18,28 +18,49 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package util
+package swim
```
jwolski: Good naming! I like how you named it specific to bootstrap rather than applying a catchall `util` name to it.

swim/node_bootstrap_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	go func() {
+		_, err := s.node.Bootstrap(&BootstrapOptions{
+			Hosts:       fakeHostPorts(1, 1, 1, 10),
+			JoinTimeout: time.Millisecond,
+		})
+		lerr.Set(err)
+	}()
+
+	time.Sleep(2 * time.Millisecond)
```
jwolski: Sleeping why?

dansimau: Not sure. This is a renamed file so I haven't seen this code before. It might be to induce a context switch, but that's just a guess; I don't know if time.Sleep() does this or not.

jwolski: I see. Let's ignore it if it's a holdover from earlier times.

swim/handlers.go
```diff
@@ -21,15 +21,17 @@
...
 	"github.com/uber/tchannel-go/json"
 	"golang.org/x/net/context"
 )
 
-// An Arg is a blank argument used as filler for making TChannel calls that require
-// nothing to be passed to Arg3
-type Arg struct{}
+// An arg is a blank argument used as filler for making TChannel calls that require
+// nothing to be passed to arg3
+type arg struct{}
```
CorgiMan: I came across this name today and had to read some code to understand what it is for.
Do you think that emptyArg3 is a better name?

swim/handlers.go
```diff
@@ -42,17 +44,24 @@ const (
...
+
 func (n *Node) registerHandlers() error {
 	handlers := map[string]interface{}{
 		"/protocol/join":      n.joinHandler,
 		"/protocol/ping":      n.pingHandler,
 		"/protocol/ping-req":  n.pingRequestHandler,
-		"/admin/debugSet":     n.debugSetHandler,
-		"/admin/debugClear":   n.debugClearHandler,
+		"/admin/debugSet":     notImplementedHandler,
+		"/admin/debugClear":   notImplementedHandler,
```
CorgiMan: I like this

swim/handlers.go
dansimau: Totally. Will change it.

util/util.go
```diff
@@ -106,23 +87,58 @@ func CheckHostnameIPMismatch(local string, hostsMap map[string][]string) ([]stri
...
+	return false
+}
 
-	var n int
-	for _, hostports := range hostsMap {
-		n += len(hostports)
+// HostListToHostMap parses a list of host/port conbinations and creates a map
+// of unique hosts each containing a slice of the hostsport instances for each
+// host.
+func HostListToHostMap(hostports []string) (hostMap map[string][]string) {
```
CorgiMan: I prefer the name, HostPortsToHostMap or HostPortsByHost.

util/util_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// indexOf returns the index of element in slice, or -1 if the element is not in slice
+func indexOf(slice []string, element string) int {
```
CorgiMan: I prefer to return (int, bool) where the bool marks wether the slice contains the element.

util/util.go
```diff
@@ -37,28 +35,11 @@ var HostportPattern = regexp.MustCompile(`^(\d+.\d+.\d+.\d+):\d+$`)
...
-	data, err := ioutil.ReadFile(file)
-	if err != nil {
-		return nil, err
-	}
-
-	err = json.Unmarshal(data, &hosts)
+	host, _, err := net.SplitHostPort(hostport)
 	if err != nil {
-		return nil, err
+		return ""
```
CorgiMan: I am not sure I like the way this swallows the error. I see its convenience but think it's a bit too subtle. Seeing as you just moved and not written it yourself it's fine. But I'd like to take a note of it :).

util/util_test.go
```diff
@@ -0,0 +1,233 @@
...
+		},
+		"192.0.2.1": []string{
+			"192.0.2.1:1",
+		},
+	}))
+}
+
+func TestMin(t *testing.T) {
+	assert.Equal(t, 1, Min(1, 2))
+	assert.Equal(t, 2, Min(10, 2))
```
CorgiMan: assert.Equal(t, 3, Min(3, 3))
:)

util/util_test.go
dansimau: Agreed. Instead of improving this though, I just unexported it and left it in as a test utility. To return a bool would make the test assertions significantly more verbose. I'm opposed to spending more time on it in this diff.

util/util.go
dansimau: This one I think we could change, because it's only used in a couple of places and one of those places is already doing error checking of an empty string.

Also it's exported so it's better to change it sooner rather than later.

util/util_test.go
dansimau: Good one, will add

swim/events.go
```diff
@@ -31,7 +31,7 @@ type EventListener interface {
 	HandleEvent(events.Event)
 }
 
-// A MaxPAdjustedEvent occurs when the disseminator adjusts the max propogation
+// A MaxPAdjustedEvent occurs when the disseminator adjusts the max propagation
 // count for changes
```
CorgiMan: period

dansimau: Haha, this is pay back, right? :smile:

swim/handlers.go
```diff
@@ -42,17 +44,24 @@ const (
...
 
+// Status contains a status string of the response from a handler.
 type Status struct {
 	Status string `json:"status"`
 }
 
+// notImplementedHandler is a dummy handler that returns an error explaining
+// this method is not implemented.
+func notImplementedHandler(ctx json.Context, req *arg) (*arg, error) {
+	return &arg{}, errors.New("handler not implemented")
```
CorgiMan: Shouldn't the first return value be nil?

dansimau: Yep, will fix

swim/node.go
dansimau: Yeah. Maybe `resolveDiscoverProvider`? That's kinda what it is doing.

util/util.go
dansimau: I like `HostPortsByHost`. Thanks for the suggestions, will update.

swim/handlers_test.go
jwolski: I would make the gossip handler return something meaningful and stub out any dependencies that the handler might have to make sure that all possible return values / responses are covered. Does that make sense? Be careful about my suggestions exploding scope though.

util/util.go
dansimau: Actually, I rescind this comment. I started pulling the thread here but `CaptureHost` is used in some pretty important methods and it opens up questions about error handling which then affects exported methods.

I think we should leave this one as-is for now.

swim/events.go
dansimau: This was simply a find/replace of a misspelling. I'm not too keen on going through the entire swim codebase and adding full-stops to the end of all the sentences.

CorgiMan: We should run your regex against the entire code base and be done with it for once and for all

util/util.go
CorgiMan: ack

dansimau: Also worth noting that the short diff here is misleading. It looks like I replaced a sensible return with a blank one. But it's just Github diff getting confused.

util/util_test.go
CorgiMan: Sure. No biggy

swim/handlers_test.go
dansimau: Yeah, I think doing this would require more work to understand where these handlers are being used and thus what should be returned over the wire. I think leave as-is for now.

swim/node.go
```diff
@@ -384,51 +375,22 @@ func (n *Node) Bootstrap(opts *BootstrapOptions) ([]string, error) {
...
-			}
-			n.log.WithField("file", "./hosts.json").Warnf("could not read host file: %v", err)
-			return errors.New("unable to read hosts file")
-		}
+// resolveDiscoverProvider reads BootstrapOptions and determines which
+// DiscoverProvider implementation to use for enumerating a list of bootstrap
+// hosts.
+// TODO: Remove this and legacy File/Hosts fields from BootstrapOptions and
+// accept only a DiscoverProvider.
+func (n *Node) resolveDiscoverProvider(opts *BootstrapOptions) (DiscoverProvider, error) {
```
CorgiMan: n isn't used, right? Does it make sense to separate the DiscoverProvider from the swin.Node? This function can then be moved to node_bootstrap.go. Let me know what you think.

util/util.go
CorgiMan: Yeah that's nice indeed. I was aware of it though

util/util_test.go
CorgiMan: (thumbsup)

CorgiMan:
LGTM

PR #61, opened 2016-01-07
Add coveralls

dansimau:
This PR adds coverage reporting to ringpop-go via coverage.io, including a badge in the README.

@uber/ringpop

### DIFF
--- a/test/update-coveralls
+++ b/test/update-coveralls
@@ -1,0 +1,39 @@
+#!/bin/bash
+#
+# Send coverage data to coveralls.io
+#
+# `go test -coverprofile=` only supports generating a coverage profile for one
+# package at a time. This script combines coverage data from multiple packages
+# and then sends the coverage data to coveralls.io
+#
+echo "mode: set" > acc.out
+FAIL=0
+
+packages="$(find . -type d -maxdepth 3 \
+	! -path './.git*' \
+	! -path '*/_*' \
+	! -path './Godeps*' \
+	! -path './test*' \
+)"
+
+# Standard go tooling behavior is to ignore dirs with leading underscores
+for dir in $packages;
+do
+  if ls $dir/*.go &> /dev/null; then
+    go test -coverprofile=profile.out $dir || FAIL=$?
+    if [ -f profile.out ]
+    then
+      cat profile.out | grep -v "mode: set" | grep -v "mocks.go" >> acc.out
+      rm profile.out
+    fi
+  fi
+done
+
+# Failures have incomplete results, so don't send
+if [ "$FAIL" -eq 0 ]; then
+  goveralls -service=travis-ci -v -coverprofile=acc.out
+fi
+
+rm -f acc.out
+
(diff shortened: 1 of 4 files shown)


.test-cover.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+echo "mode: set" > acc.out
+FAIL=0
+
+# Standard go tooling behavior is to ignore dirs with leading underscors
```
jwolski: typo "underscores"

dansimau: :scream: Will fix.

(Stole this script from https://github.com/uber/go-torch/blob/master/.test-cover.sh)

.test-cover.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
jwolski: Can you add a brief description describing what this script is intending to do? I know these type of scripts are typically less readable than normal given that they're just working around Travis/Coveralls quirks, but it'd be nice to know why we need this.

thanodnl:
is there a way to keep [`test/remoteservice`](https://travis-ci.org/uber/ringpop-go/builds/101017947#L183) out of the coverage? There are 4 files in that package that are generated by `thrift` and `tchannel` to test the 1 file that we generate. Coverage of this package will _never_ be high.

dansimau:
@thanodnl ah yes, we can exclude that quite easily. Will fix.

.test-cover.sh
dansimau: Good point. I'll add something.

test/update-coveralls
```diff
@@ -0,0 +1,39 @@
...
+# and then sends the coverage data to coveralls.io
+#
+echo "mode: set" > acc.out
+FAIL=0
+
+packages="$(find . -type d -maxdepth 3 \
+	! -path './.git*' \
+	! -path '*/_*' \
+	! -path './Godeps*' \
+	! -path './test*' \
```
thanodnl: As I read the PR it seems to me that the tests that are in the `./test` directory are not run anymore since this script replaced `go test ./...` from the `Makefile`. I think we still want to run the tests in the `./test` directory, but not collect the coverage report.

dansimau: Ah, nice spot. I am thinking maybe we should separate the running of tests from the coverage report. What do you think? It will add ~60 seconds to the build time for Travis, but it keeps the concerns separate.

thanodnl: I think that will be a great thing todo actually. It can be an other entry in the matrix so it can be ran in parallel.

dansimau: Jeez, you don't ask much, do you!

dansimau: OK, done. Here is an example build:
https://travis-ci.org/uber/ringpop-go/builds/104288832

jwolski:
kewl. u r a Travis god. lgtm

thanodnl:
lgtm

PR #62, opened 2016-01-08
Rework hashring locks plus 100% test coverage

CorgiMan:
Some refactoring due to locks. Now only locking exported functions and make a very clear distinction in comments and function names on which functions close the lock.

Flattened out the servers struct in the HashRing which provided no clear benefit.

More idiomatic go naming: e.g. NewHashRing -> New

Use strings.Join in computeChecksum instead of a manual approach.

### DIFF
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -18,12 +18,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+// Package hashring provides a hashring implementation that uses a red-black
+// Tree.
 package hashring
 
 import (
-	"bytes"
 	"fmt"
 	"sort"
+	"strings"
 	"sync"
 
 	"github.com/uber/ringpop-go/events"
@@ -31,9 +33,9 @@ import (
 	"github.com/dgryski/go-farm"
 )
 
-// HashRingConfiguration is a configuration struct that can be passed to the
+// Configuration is a configuration struct that can be passed to the
 // Ringpop constructor to customize hash ring options.
-type HashRingConfiguration struct {
+type Configuration struct {
 	// ReplicaPoints is the number of positions a node will be assigned on the
 	// ring. A bigger number will provide better key distribution, but require
 	// more computation when building or traversing the ring (typically on
@@ -41,16 +43,17 @@ type HashRingConfiguration struct {
 	ReplicaPoints int
 }
 
+// HashRing stores strings on a consistent hash ring. HashRing internally uses
+// a Red-Black Tree to achieve O(log N) lookup and insertion time.
 type HashRing struct {
-	hashfunc      func([]byte) uint32
+	sync.RWMutex
+
+	hashfunc      func(string) int
 	replicaPoints int
 
-	servers struct {
-		byAddress map[string]bool
-		tree      *redBlackTree
-		checksum  uint32
-		sync.RWMutex
-	}
+	serverSet map[string]struct{}
+	tree      *redBlackTree
+	checksum  uint32
 
 	listeners []events.EventListener
 }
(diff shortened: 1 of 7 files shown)


CorgiMan:
@danielheller @thanodnl @jwolski 
Rebased on master. I think I have addressed all feedback. Enjoy

hashring/hashring.go
```diff
@@ -66,156 +68,175 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
+// HasServer returns wether the HashRing contains the given server.
+func (r *HashRing) HasServer(server string) bool {
+	r.RLock()
+	_, ok := r.serverSet[server]
+	r.RUnlock()
+	return ok
+}
 
-	return server
+// Servers returns all servers contained in the HashRing.
```
thanodnl: Servers -> CopyServers to be valid godoc.

hashring/hashring_test.go
```diff
@@ -23,15 +23,45 @@ package hashring
...
 	"testing"
 
 	"github.com/dgryski/go-farm"
+	"github.com/uber/ringpop-go/events"
 
 	"github.com/stretchr/testify/assert"
 )
 
+// fake event listener
+type dummyListener struct {
```
thanodnl: Could you use mockery here?  I think @dansimau has put a mock for `EventListener` in the events package. Don't know if that is in master yet.

Mocks made with mockery also have a function to count the number of calls that were made to a certain function of an interface.

thanodnl: https://github.com/uber/ringpop-go/pull/59/files#diff-3552bf85474753d5d484a86fce729755

hashring/hashring.go
```diff
@@ -66,156 +68,175 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
-	r.computeChecksum()
+	r.Unlock()
+	return ok
 }
 
-// inserts server replicas into ring
-func (r *HashRing) addReplicas(server string) {
-	r.servers.Lock()
-	r.servers.byAddress[server] = true
+// This function isn't thread-safe, only call it when the HashRing is locked.
```
jwolski: Love that these comments are here.

hashring/hashring.go
```diff
@@ -66,156 +68,175 @@ func (r *HashRing) RegisterListener(l events.EventListener) {
...
-	r.servers.RUnlock()
-
+	r.RLock()
+	count := len(r.serverSet)
+	r.RUnlock()
 	return count
 }
 
-// Lookup returns the owner of the given key.
+// Lookup returns the owner of the given key and wether the HashRing contains
```
jwolski: typo/misspelling: `wether` -> `whether`

hashring/hashring_test.go
```diff
@@ -92,8 +120,25 @@ func TestServerCount(t *testing.T) {
...
 
+func TestCopyServers(t *testing.T) {
+	ring := New(farm.Fingerprint32, 10)
+	assert.Equal(t, 0, ring.ServerCount(), "expected one server to be in ring")
+
+	ring.AddServer("server1")
+	ring.AddServer("server2")
+	ring.AddServer("server3")
+
+	servers := ring.CopyServers()
```
jwolski: I think `Servers()` is a better name. The fact that they're being copied doesn't need to be in the name. It can just be in the documentation.

hashring/hashring_test.go
```diff
@@ -217,8 +263,9 @@ func BenchmarkHashRingLookupN(b *testing.B) {
 
 	b.StartTimer()
 	for n := 0; n < b.N; n++ {
+		k := rand.Intn(100)
```
jwolski: randomness in general seems bad in benchmarks especially when we're talking orders of magnitude. why randomize?

hashring/hashring.go
```diff
@@ -41,16 +42,17 @@ type HashRingConfiguration struct {
 	ReplicaPoints int
 }
 
+// HashRing stores strings on a consistent hash ring. HashRing internally uses
```
jwolski: I think an alternative to the approach of making exported fns thread-safe and non-exported not would be to have two types: `HashRing` and `ConcurrentHashRing` both with the safe interface. `ConcurrentHashRing` would be composed of a `HashRing` and simply delegate fn calls to it. I think that type of approach is closer to the Java approach.

hashring/hashring.go
CorgiMan: thank you

hashring/hashring_test.go
CorgiMan: Good catch.

hashring/hashring_test.go
CorgiMan: That is not yet merged. I will change it at the first convenience.

hashring/hashring.go
CorgiMan: Changed the function name back as Jeff has suggested

hashring/hashring.go
CorgiMan: Interesting indeed.

hashring/hashring.go
CorgiMan: (thumbsup)

jwolski:
lgtm

thanodnl:
lgtm

PR #63, opened 2016-01-09
Fix a potiental goroutine leak by using buffer channel.

sunzl1989:


dansimau:
For the benefit of those without access to Uber's internal systems:

This fixes a bug that caused goroutine leakage on forwarded requests. When a [forwarded request timed out](https://github.com/uber/ringpop-go/blob/6470309/forward/request_sender.go#L119), the goroutine would hang indefinitely, as the parent call would never read the result from the channel.

The fix here creates a buffered channel so that the goroutine in `MakeCall` can always write to it without being blocked. In the case where a timeout occurs, the result will never be read, but the goroutine will be able to return appropriately.

dansimau:
LGTM. :+1:

thanodnl:
LGTM

jwolski:
I feel that this is not addressing the root of the problem. There should be a consistent way of ending a forwarded request whether it's a result of a valid response or a timeout. The way that interact with the underlying should be consistent in either case.

The problem is not necessarily that the channel is unbuffered. The problem is that we are inconsistent in how we consume from the channel.

I understand that this is a quick fix, but I'd prefer to see an alternate solution. Thoughts from @thanodnl @dansimau @sunzl1989?

jwolski:
Following up on my previous comment. This is a quick fix. The non-quick fix will likely be less quick........ We'll merge this and tackle improving how we interact with our forwarding channels in another PR.

@sunzl1989 Thanks for your contrib.

PR #64, opened 2016-01-10
Emit event from AddServer and RemoveServer

CorgiMan:
After putting the AddServer and RemoveServer back in, I forgot to let them emit the RingChanged event. I also put enhanced the tests to check for these events so this won't happen again.

@uber/ringpop

### DIFF
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -64,7 +64,7 @@ func (r *HashRing) emit(event interface{}) {
 	}
 }
 
-// RegisterListener adds a listener that will be sent swim events.
+// RegisterListener adds a listener that will listen for hashring events.
 func (r *HashRing) RegisterListener(l events.EventListener) {
 	r.listeners = append(r.listeners, l)
 }
@@ -110,12 +110,11 @@ func (r *HashRing) computeChecksumNoLock() {
 // AddServer adds a server and its replicas onto the HashRing.
 func (r *HashRing) AddServer(address string) bool {
 	r.Lock()
-
 	ok := r.addServerNoLock(address)
 	if ok {
 		r.computeChecksumNoLock()
+		r.emit(events.RingChangedEvent{[]string{address}, nil})
 	}
-
 	r.Unlock()
 	return ok
 }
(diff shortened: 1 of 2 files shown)


hashring/hashring.go
```diff
@@ -64,7 +64,7 @@ func (r *HashRing) emit(event interface{}) {
 	}
 }
 
-// RegisterListener adds a listener that will be sent swim events.
+// RegisterListener adds a listener that will listens for HashRing events.
```
dansimau: Not sure about this. I actually think this comment was more accurate before. In any case, there is a typo in the updated comment.

hashring/hashring.go
```diff
@@ -197,8 +195,8 @@ func (r *HashRing) addRemoveServersNoLock(add []string, remove []string) bool {
 	}
 
 	if changed {
-		r.emit(events.RingChangedEvent{add, remove})
 		r.computeChecksumNoLock()
+		r.emit(events.RingChangedEvent{add, remove})
```
dansimau: :+1:

hashring/hashring.go
dansimau: Actually, scratch that. I just looked more closely at the events and you're right. A listener does receive all event types. My bad. Just the typo then: `s/listens/listen`! :smile:

hashring/hashring_test.go
```diff
@@ -52,42 +52,45 @@ func (d *dummyListener) HandleEvent(event events.Event) {
...
-	ring.emit(struct{}{})
-	assert.Equal(t, 1, l.events, "expected one event to be emitted")
-}
 
-func TestAddServer(t *testing.T) {
-	ring := New(farm.Fingerprint32, 10)
 	ring.AddServer("server1")
 	ring.AddServer("server2")
-
+	assert.Equal(t, 4, l.EventCount())
```
dansimau: I know it's been mentioned before—and this is just a minor thing—but I think we should move to using generated mocks where possible, for the sake of ease and standardisation.

With the generated mock you could do:

```
// Set up mock event listener with expectations
mockEventListener := &mocks.EventListener{}
mockEventListener.On("HandleEvent", mock.AnythingOfType("RingChangedEvent")).Times(5)
ring.RegisterListener(mockEventListener)

// Your test code here ...

// Assert expectations
mockEventListener.AssertExpectations()    // Will complain if the RingChangedEvent wan't fired 5 times
```

An example is here in this PR:
https://github.com/uber/ringpop-go/blob/tests/swim/swim/ping_test.go#L85

Note that I don't think this is a blocker for landing the PR. Just a nice-to-have.

dansimau:
Code lgtm :+1:

thanodnl:
:+1:

hashring/hashring_test.go
CorgiMan: I agree, but the mockEventListener is not yet in master right? We can merge the changes in there, or clean up when that lands.

CorgiMan:
Thanks for the quick review boys

PR #65, opened 2016-01-11
Don't start disseminating form the join handler

thanodnl:
This PR implements a feature recently implemented in ringpop-node via https://github.com/uber/ringpop-node/pull/217.

It prevents the nodes receiving a join to start gossiping the aliveness of the new node to prevent the node joining the ring before it is pining its self which caused indefinite partitions in the past.

### DIFF
--- a/swim/disseminator.go
+++ b/swim/disseminator.go
@@ -86,6 +86,11 @@ func (d *disseminator) AdjustMaxPropagations() {
 	d.Unlock()
 }
 
+// HasChanges reports whether disseminator has changes to disseminate.
+func (d *disseminator) HasChanges() bool {
+	return len(d.changes) > 0
+}
+
 func (d *disseminator) FullSync() (changes []Change) {
 	d.Lock()
 
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -49,7 +49,7 @@ func (s *BootstrapTestSuite) TearDownTest() {
 
 func (s *BootstrapTestSuite) TestBootstrapOk() {
 	s.peers = genChannelNodes(s.T(), 5)
-	bootstrapNodes(s.T(), append(s.peers, s.tnode)...)
+	bootstrapNodes(s.T(), false, append(s.peers, s.tnode)...)
 	// Reachable members should be s.node + s.peers
 	s.Equal(6, s.node.CountReachableMembers())
 }
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -105,6 +105,8 @@ func (s *BootstrapTestSuite) TestJoinHandlerNotMakingAlive() {
 
 	// test that there are no changes to disseminate after the bootrstrapping of a host
 	for _, peer := range s.peers {
+		_, hasMember := peer.node.memberlist.Member(s.tnode.node.Address())
+		s.False(hasMember, "didn't expect the bootstapping node to appear in the member list of any peers")
 		s.Len(peer.node.disseminator.changes, 0)
 	}
 
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -49,7 +49,7 @@ func (s *BootstrapTestSuite) TearDownTest() {
 
 func (s *BootstrapTestSuite) TestBootstrapOk() {
 	s.peers = genChannelNodes(s.T(), 5)
-	bootstrapNodes(s.T(), false, append(s.peers, s.tnode)...)
+	bootstrapNodes(s.T(), append(s.peers, s.tnode)...)
 	// Reachable members should be s.node + s.peers
 	s.Equal(6, s.node.CountReachableMembers())
 }
@@ -97,7 +97,8 @@ func (s *BootstrapTestSuite) TestBootstrapDestroy() {
 func (s *BootstrapTestSuite) TestJoinHandlerNotMakingAlive() {
 	// get a bootstrapped cluster
 	s.peers = genChannelNodes(s.T(), 3)
-	bootstrapList := bootstrapNodes(s.T(), true, s.peers...)
+	bootstrapList := bootstrapNodes(s.T(), s.peers...)
+	waitForConvergence(s.T(), 500*time.Millisecond, s.peers...)
 
 	s.tnode.node.Bootstrap(&BootstrapOptions{
 		Hosts: bootstrapList,
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -94,6 +94,22 @@ func (s *BootstrapTestSuite) TestBootstrapDestroy() {
 	s.EqualError(chanErr, "node destroyed while attempting to join cluster")
 }
 
+func (s *BootstrapTestSuite) TestJoinHandlerNotMakingAlive() {
+	// get a bootstrapped cluster
+	s.peers = genChannelNodes(s.T(), 3)
+	bootstrapList := bootstrapNodes(s.T(), true, s.peers...)
+
+	s.tnode.node.Bootstrap(&BootstrapOptions{
+		Hosts: bootstrapList,
+	})
+
+	// test that there are no changes to disseminate after the bootrstrapping of a host
+	for _, peer := range s.peers {
+		s.Len(peer.node.disseminator.changes, 0)
+	}
+
+}
+
 func (s *BootstrapTestSuite) TestBootstrapFailsWithNoChannel() {
 	n := &Node{}
 	_, err := n.Bootstrap(nil)
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -104,11 +104,11 @@ func (s *BootstrapTestSuite) TestJoinHandlerNotMakingAlive() {
 		Hosts: bootstrapList,
 	})
 
-	// test that there are no changes to disseminate after the bootrstrapping of a host
+	// test that there are no changes to disseminate after the bootstrapping of a host
 	for _, peer := range s.peers {
 		_, hasMember := peer.node.memberlist.Member(s.tnode.node.Address())
 		s.False(hasMember, "didn't expect the bootstapping node to appear in the member list of any peers")
-		s.Len(peer.node.disseminator.changes, 0)
+		s.False(peer.node.HasChanges(), "didn't expect existing node to have changes")
 	}
 
 }
--- a/swim/test_utils.go
+++ b/swim/test_utils.go
@@ -111,7 +111,7 @@ func memberlistHasMembers(t *testing.T, m *memberlist, members []Member) {
 	}
 }
 
-func bootstrapNodes(t *testing.T, waitForConvergence bool, testNodes ...*testNode) []string {
+func bootstrapNodes(t *testing.T, testNodes ...*testNode) []string {
 	var hostports []string
 
 	for _, tn := range testNodes {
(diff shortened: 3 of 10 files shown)


swim/test_utils.go
```diff
@@ -134,7 +134,7 @@ func memberlistHasMembers(t *testing.T, m *memberlist, members []Member) {
 	}
 }
 
-func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
+func bootstrapNodes(t *testing.T, waitForConvergence bool, testNodes ...*testNode) []string {
```
dansimau: I'm not sure the `waitForConvergence` needs to be part of the bootstrap function, does it?

I think we could create a separate function, e.g. `waitForConvergence(timeout time.Duration, nodes ...*testNode)` and then call that if you have a test that needs this.

I think that has the advantage that you don't need change the existing `bootstrapNodes` calls and those calls are slightly more readable, i.e. you don't have a boolean in the call which requires you to look at the signature to figure out what it is.

swim/test_utils.go
```diff
@@ -146,6 +146,23 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
 		})
 		require.NoError(t, err, "node must bootstrap successfully")
 	}
+
+	if waitForConvergence {
```
dansimau: Nice one! Can we do this using public interfaces?

I realise that no disseminator changes should mean that the ring has converged, but that is an internal implementation detail.

I think we should be looping over the nodes in the ring to ensure they all have the same checksum, as this matches our definition of "converged".

You mentioned yesterday that it would be ideal to call a `ProtocolPeriod` during some tests. Note that can easily be triggered by calling the `/admin/gossip/tick` endpoint.

Finally, I think it would be good to have a timeout here, i.e. if something goes wrong and convergence doesn't occur within a healthy timeframe for the test (e.g. 500ms), then we fail to avoid hanging indefinitely.

swim/test_utils.go
jwolski: This is a good suggestion from @dansimau.

swim/test_utils.go
jwolski: Agree with @dansimau. We should verify an empty dissemination list and membership checksums.

swim/bootstrap_test.go
```diff
@@ -88,6 +88,22 @@ func (s *BootstrapTestSuite) TestBootstrapDestroy() {
...
+	// get a bootstrapped cluster
+	s.peers = genChannelNodes(s.T(), 3)
+	bootstrapList := bootstrapNodes(s.T(), true, s.peers...)
+
+	s.tnode.node.Bootstrap(&BootstrapOptions{
+		Hosts: bootstrapList,
+	})
+
+	// test that there are no changes to disseminate from the peers after bootstrapping the t(est)node
+	for _, peer := range s.peers {
```
jwolski: A more accurate assertion or test would be to verify that when calling the /protocol/join endpoint, the caller (joiner) is not added to membership. Testing whether the disseminator has changes is a roundabout (and potentially frail) way of verifying that the member was not made alive. It relies too much on the interactions between the membership component and the disseminator.

thanodnl: I added a check to make sure the member is not showing up in the membership list. Do you suggest to remove the dissemination test or is that valid to keep here?

jwolski: Keeping the disseminator assertion is OK with me, but I'd prefer an API like `node.PendingChangeCount()` rather than digging into node -> disseminator -> changes.

swim/test_utils.go
```diff
@@ -146,6 +146,52 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
...
+		// execute the Protocol period on all nodes until the dissemination list is exhausted
+		go func(tn *testNode) {
+			for len(tn.node.disseminator.changes) > 0 {
+				tn.node.gossip.ProtocolPeriod()
+			}
+			wg.Done()
+		}(tn)
+	}
+
+	finished := make(chan bool, 1)
```
jwolski: Nice. This a nice implementation of a fork/join-type pattern. Do we need to buffered channel (of `1) here? Can it just be `make(chan bool)`?

thanodnl: If we make this an unbuffered channel the routine that waits for the `WaitGroup` to finish could end up in a state where it could not produce to the channel if the timeout already occurred.

swim/bootstrap_test.go
thanodnl: Ok, will change

swim/test_utils.go
jwolski: Ah, interesting. Good to know.

swim/node_bootstrap_test.go
```diff
@@ -94,6 +94,25 @@ func (s *BootstrapTestSuite) TestBootstrapDestroy() {
...
+
+	s.tnode.node.Bootstrap(&BootstrapOptions{
+		Hosts: bootstrapList,
+	})
+
+	// test that there are no changes to disseminate after the bootrstrapping of a host
+	for _, peer := range s.peers {
+		_, hasMember := peer.node.memberlist.Member(s.tnode.node.Address())
+		s.False(hasMember, "didn't expect the bootstapping node to appear in the member list of any peers")
+		s.Len(peer.node.disseminator.changes, 0)
```
jwolski: How about the API changes here? (e.g. `node.PendingChangeCount()`)

swim/test_utils.go
```diff
@@ -121,6 +123,52 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
...
+	go func() {
+		wg.Wait()
+		finished <- true
+	}()
+
+	select {
+	case <-finished:
+		var checksums []uint32
```
jwolski: It looks like there are some fine opportunities to DRY up this convergence check with the one below. I'd go so far to say that we should start building an API for our test cluster type (e.g. `cluster.HasConverged()`).

swim/test_utils.go
```diff
@@ -121,6 +123,52 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
...
+		for _, node := range testNodes {
+			checksums = append(checksums, node.node.memberlist.Checksum())
+		}
+
+		if len(checksums) == 0 {
+			// no checksums to validate
+			return
```
jwolski: This seems like an error condition that should be surfaced to callers.

swim/node_bootstrap_test.go
```diff
@@ -94,6 +94,25 @@ func (s *BootstrapTestSuite) TestBootstrapDestroy() {
...
+	// get a bootstrapped cluster
+	s.peers = genChannelNodes(s.T(), 3)
+	bootstrapList := bootstrapNodes(s.T(), s.peers...)
+	waitForConvergence(s.T(), 500*time.Millisecond, s.peers...)
+
+	s.tnode.node.Bootstrap(&BootstrapOptions{
+		Hosts: bootstrapList,
+	})
+
+	// test that there are no changes to disseminate after the bootrstrapping of a host
```
jwolski: typo in comment "bootrstrapping" -> "bootstrapping"

swim/test_utils.go
```diff
@@ -121,6 +123,52 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
...
+	return hostports
+}
+
+func waitForConvergence(t *testing.T, timeout time.Duration, testNodes ...*testNode) {
+	var wg sync.WaitGroup
+	for _, tn := range testNodes {
+		wg.Add(1)
+		// execute the Protocol period on all nodes until the dissemination list is exhausted
+		go func(tn *testNode) {
+			for len(tn.node.disseminator.changes) > 0 {
```
jwolski: Exhausting changes seems like a useful API to have, if not at the `swim.Node`, then definitely the test suite API layer. I recommend that we pull out the `swimCluster` type from test_utils and add `ExhaustChanges()` to it.

swim/test_utils.go
```diff
@@ -206,6 +254,36 @@ func (c *swimCluster) Bootstrap() {
 	}
 }
 
+func (c *swimCluster) WaitForConvergence(timeout time.Duration) error {
```
dansimau: `// WaitForConvergence polls the checksums of each node in the cluster and returns when they all match, or until timeout occurs.`

swim/test_utils.go
dansimau: +1

thanodnl: Would love to, but on is iterating over a `[]swim.Node` and the other iterates over `[]testNode` which are different types. I can DRY the loop that checks that all `[]uint32` to be the same, but that is all about it I think. Any thoughts?

swim/node_bootstrap_test.go
thanodnl: Somehow the previous conversation about this got lost. I'm planning to change this indeed.

swim/test_utils.go
dansimau: Perhaps add a method: `testNode.Nodes()`?

dansimau: Oh, ignore that. I was getting confused with `swimCluster`. But yeah, perhaps a function to gather the nodes, given a slice of testNodes?

swim/test_utils.go
```diff
@@ -121,6 +122,44 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
 		})
 		require.NoError(t, err, "node must bootstrap successfully")
 	}
+	return hostports
+}
+
+func waitForConvergence(t *testing.T, timeout time.Duration, testNodes ...*testNode) {
```
thanodnl: Dear reviewers,

I've rewritten this method because it was causing flappy tests. Please have a look at the new version.

dansimau: Should we sleep for 10ms between iterations of the for loop?

thanodnl: I don't think that will change anything here except for the test taking longer. We call `ProtocolPeriod` every iteration so we can move as fast as this method allows us. It blocks till the ping response is received.

swim/test_utils.go
```diff
@@ -121,6 +122,44 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
...
+
+	nodes := testNodesToNodes(testNodes)
+
+	for {
+		select {
+		case <-timeoutCh:
+			t.Errorf("timeout during wait for convergence")
+			return
+		default:
+			// tick all the nodes
```
dansimau: Why are we doing this again? I forgot. Can we explain in the comment?

swim/test_utils.go
dansimau: Ah, cool. Sorry, I thought `ProtocolPeriod()` was async.

swim/disseminator.go
```diff
@@ -86,6 +86,11 @@ func (d *disseminator) AdjustMaxPropagations() {
 	d.Unlock()
 }
 
+// HasChanges reports whether disseminator has changes to disseminate.
```
jwolski: :dancer:

swim/test_utils.go
```diff
@@ -121,6 +122,47 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) {
...
+	for {
+		select {
+		case <-timeoutCh:
+			t.Errorf("timeout during wait for convergence")
+			return
+		default:
+			for _, node := range nodes {
+				node.gossip.ProtocolPeriod()
+			}
+			hasChanges := false
```
jwolski: This code is a bit tricky. I think it can actually be improved with a goto:

```
for {
  START:

  select {
  case <-timeoutCh:
    // existing stuff
  default:
    for _, node := range nodes {
      node.gossip.ProtocolPeriod()
    }

    for _, node := range nodes {
      if node.HasChanges() {
        goto START
      }
    }

    if !nodesConverged(node) {
      // etc
    }
  }
}
```

what do you think?

thanodnl: I was thinking about it just before. I will **NOT** use a goto, but I will look into [label breaks](http://www.goinggo.net/2013/11/label-breaks-in-go.html) and see if it also applies to continue.

jwolski: Go's standard library has gotos to make control flow easier to read especially in cases where you're using booleans to achieve the same thing: https://golang.org/src/math/gamma.go. I can't promise a dinosaur will not come to eat you though.

jwolski:
lgtm

PR #67, opened 2016-01-13
Improve test that verifies correct behavior of a destroyed node durin…

jwolski:
…g bootstrap by making using of channels not sleeps

This test depended upon appropriately timed sleep statements for correct behavior. I've taken out the sleeps and replaced with a error sharing over a channel.

Regarding why this is being done (other than to take variability out), it seems as though that by changing the timing of tests and join sender, the test will fail.

@uber/ringpop

### DIFF
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -74,21 +74,24 @@ func (s *BootstrapTestSuite) TestBootstrapJoinsTimeOut() {
 }
 
 func (s *BootstrapTestSuite) TestBootstrapDestroy() {
-	var lerr lError
+	// Destroy node first to ensure there are no races
+	// in how the goroutine below is scheduled.
+	s.node.Destroy()
+
+	errChan := make(chan error)
 
 	go func() {
 		_, err := s.node.Bootstrap(&BootstrapOptions{
 			Hosts:       fakeHostPorts(1, 1, 1, 10),
 			JoinTimeout: time.Millisecond,
 		})
-		lerr.Set(err)
+		errChan <- err
 	}()
 
-	time.Sleep(2 * time.Millisecond)
-	s.node.Destroy()
-	time.Sleep(2 * time.Millisecond)
-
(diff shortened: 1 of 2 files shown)


dansimau:
Look good.

I think this patch removes all uses of `lError` (which is great) so should we remove it from `swim/test_utils.go`?

```
swim/test_utils.go:
57:type lError struct {
```

jwolski:
sure thing. I'll remove.

thanodnl:
lgtm

swim/node_bootstrap_test.go
```diff
@@ -74,21 +74,20 @@ func (s *BootstrapTestSuite) TestBootstrapJoinsTimeOut() {
...
 		_, err := s.node.Bootstrap(&BootstrapOptions{
 			Hosts:       fakeHostPorts(1, 1, 1, 10),
 			JoinTimeout: time.Millisecond,
 		})
-		lerr.Set(err)
+		errChan <- err
 	}()
 
-	time.Sleep(2 * time.Millisecond)
 	s.node.Destroy()
```
CorgiMan: Is the bootstrap supposed to happen before or after the destroy. The code doesn't look deterministic to me on which one comes first.

CorgiMan: Is the bootstrap supposed to happen before or after the destroy? The code doesn't look deterministic to me on which one comes first.

jwolski: This is a good question. In theory, it's possible for the Bootstrap to be scheduled before the continuation of the test function. I've heard about some guarantees around goroutine scheduling. I'll try to dig up an answer from somewhere and make change if I have to.

CorgiMan: I think if Destroy should go first, we would benefit from moving it up. It makes it easier to understand what we're testing for.

CorgiMan:
I like the channel approach! lgtm after you have addressed my comment

swim/node_bootstrap_test.go
jwolski: Agreed

PR #68, opened 2016-01-13
Introduce exponentially increasing delay after each join attempt

jwolski:
This PR adds a delay in between each join attempt. The delay is a calculation of an exponential backoff with added jitter/fuzz. The maximum potential delay will increase until it is capped at its max. The actual delay will be a random number between the last used max and the new max. Therefore, after the first join attempt, the delay will be between 0 and 100. On the second attempt, 100 and 200. On the third attempt, between 200 and 400 and so on.

cc @uber/ringpop

### DIFF
--- a/swim/join_delayer.go
+++ b/swim/join_delayer.go
@@ -1,0 +1,210 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package swim
+
+import (
+	"errors"
+	"math"
+	"math/rand"
+	"time"
+
+	"github.com/uber-common/bark"
+	"github.com/uber/ringpop-go/util"
+)
+
+var errLoggerRequired = errors.New("joinDelayer requires a logger")
+
+const (
+	defaultInitial = 100 * time.Millisecond
+	defaultMax     = 60 * time.Second
+)
+
+var delayerRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+var defaultRandomizer = delayerRand.Intn
+var defaultSleeper = time.Sleep
+var noDelay = time.Duration(0)
+
+// joinDelayer defines the API of a delayer implementation that
+// applies a delay in between repeated join attempts.
+type joinDelayer interface {
+	delay() time.Duration
+}
+
+// delayRandomizer is a function that returns a random number between
+// 0 and an upper-bound, provided in its first argument.
+type delayRandomizer func(int) int
+
+// delaySleeper is a function that pauses execution for time.Duration.
+type delaySleeper func(time.Duration)
+
+// delayOpts is a struct that houses substitutable parameters of a joinDelayer
+// in order to alter its behavior.
+type delayOpts struct {
+	initial    time.Duration
+	max        time.Duration
+	randomizer delayRandomizer
+	sleeper    delaySleeper
+}
+
+// newDelayOpts creates a delayOpts struct with default values.
+func newDelayOpts() *delayOpts {
+	return &delayOpts{
+		initial:    defaultInitial,
+		max:        defaultMax,
+		randomizer: defaultRandomizer,
+		sleeper:    defaultSleeper,
+	}
+}
+
+// exponentialDelayer applies a delay in between repeated join attempts.
+// The delay increases exponentially and is capped at maxDelay.
+type exponentialDelayer struct {
+	// joiner is the address of the node sending join requests.
+	joiner string
+
+	// logger logs when a maximum delay is reached.
+	logger bark.Logger
+
+	// initialDelay is the fixed portion of the delay by which
+	// the exponential backoff is multiplied.
+	initialDelay time.Duration
+
+	// maxDelay is the maximum delay applied to a join attempt.
+	maxDelay time.Duration
+
+	// maxDelayReached is a flag that is toggled once the delay
+	// applied to a join attempt reaches or exceeds the max delay.
+	maxDelayReached bool
+
+	// nextDelayMin tracks the last used upper-bound for a join attempt delay.
+	// Upon the next join attempt, nextDelayMin will be used as the lower-bound
+	// for that delay. This acts as a shifting window for the bounds of the
+	// exponential backoff.
+	nextDelayMin float64
+
+	// randomizer generates a random delay in between a nextDelayMin
+	// and maxDelay.
+	randomizer delayRandomizer
+
+	// sleeper pauses execution of a join attempt.
+	sleeper delaySleeper
+
+	// numDelays tracks the number of times delay has been called on the
+	// delayer. It's also used as the backoff exponent.
+	numDelays uint
+}
+
+// newExponentialDelayer creates a new exponential delayer. joiner and logger
+// are required. opts is optional.
+func newExponentialDelayer(joiner string, logger bark.Logger,
+	opts *delayOpts) (*exponentialDelayer, error) {
+	if logger == nil {
+		return nil, errLoggerRequired
+	}
+
+	if opts == nil {
+		opts = newDelayOpts()
+	}
+
+	randomizer := opts.randomizer
+	if randomizer == nil {
+		randomizer = defaultRandomizer
+	}
+
+	sleeper := opts.sleeper
+	if sleeper == nil {
+		sleeper = defaultSleeper
+	}
+
+	return &exponentialDelayer{
+		joiner:          joiner,
+		logger:          logger,
+		initialDelay:    opts.initial,
+		nextDelayMin:    0,
+		maxDelayReached: false,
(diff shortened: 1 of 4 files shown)


swim/node_bootstrap_test.go
```diff
@@ -74,21 +74,20 @@ func (s *BootstrapTestSuite) TestBootstrapJoinsTimeOut() {
 }
 
 func (s *BootstrapTestSuite) TestBootstrapDestroy() {
-	var lerr lError
+	errChan := make(chan error)
```
jwolski: NOTES TO REVIEWER: Don't mind these changes. I cherry-picked a commit from https://github.com/uber/ringpop-go/pull/67 so that these tests pass. I'll delete this commit from this PR when that PR is merged.

thanodnl: Now that #67 is merged can we undo this from this PR?

jwolski: of course. i will take care of it.

swim/join_test.go
```diff
@@ -43,6 +48,32 @@ func (s *JoinSenderTestSuite) SetupTest() {
...
+	joinSender, err := newJoinSender(s.node, &joinOpts{
+		discoverProvider: &StaticHostList{bootstrapHosts},
+	})
+	s.NoError(err, "joiner created successfully")
+	return joinSender
+}
+
+func (s *JoinSenderTestSuite) setupTestForDelay() (*joinSender, [6]float64) {
+	joinSender := s.newJoinSenderForDelay()
+	joinSender.delayOpts.initial = 100
```
thanodnl: instead of changing the internal state of the `delayOpts` can we pass in this object to `newJoinSender` via `s.newJoinSenderForDelay`?

swim/join_sender.go
```diff
@@ -162,6 +203,15 @@ func (j *joinSender) parseHosts(hostports []string) {
 	}
 }
 
+func newDelayOpts() *delayOpts {
+	return &delayOpts{
```
thanodnl: Would it make sense to not change the delayOpts after they are passed into the joiner and reuse the defaults object? Currently I only see changes made to `joinsender.delayOpts` during the tests which could also be passed in via the `newJoinSender`.

swim/join_sender.go
```diff
@@ -334,6 +398,68 @@ func (j *joinSender) JoinCluster() ([]string, error) {
 	return nodesJoined, nil
 }
 
+// delayAttempt delays a join attempt by sleeping for an amount of time. The
+// amount of time is computed as an exponential backoff based on the number
+// of join attempts that have been made at the time of the function call;
+// the number of attempts is 0-based. It returns a time.Duration equal to the
+// amount of delay applied and an error if one occurs. If an error occurs, a
+// duration equal to 0 will be returned, signifying no delay has been applied.
+func (j *joinSender) delayAttempt(numAttempts int, randomizer func(int) int,
```
thanodnl: can't we read the `randomizer` and `sleeper` from the `joinSender` instead of passing them in to this member function?

swim/join_test.go
jwolski: good point. i'll clean it up.

swim/join_sender.go
jwolski: we can. I felt as though parameter-injection is a better fit for this function as opposed to injecting them in through the constructor. The dependencies are so specific to this one function that I felt it unnecessary to do hang them off of the `joinSender` type. But, this same thought crossed my mind and I'll try it on for size in the code.

thanodnl: you also use the `initial` and `max` from the `delayOpts` so if you want to keep them injectable I would suggest to pass in the complete `delayOpts` struct to keep them all together.

swim/join_sender.go
```diff
@@ -334,6 +398,68 @@ func (j *joinSender) JoinCluster() ([]string, error) {
 	return nodesJoined, nil
 }
```
CorgiMan: What do you think of putting this logic in a method of an exponentialDelay object?

swim/join_sender.go
```diff
@@ -334,6 +398,68 @@ func (j *joinSender) JoinCluster() ([]string, error) {
...
 
+// delayAttempt delays a join attempt by sleeping for an amount of time. The
+// amount of time is computed as an exponential backoff based on the number
+// of join attempts that have been made at the time of the function call;
+// the number of attempts is 0-based. It returns a time.Duration equal to the
+// amount of delay applied and an error if one occurs. If an error occurs, a
+// duration equal to 0 will be returned, signifying no delay has been applied.
+func (j *joinSender) delayAttempt(numAttempts int, randomizer func(int) int,
+	sleeper func(time.Duration)) (time.Duration, error) {
```
CorgiMan: We can clean up the arguments of this function by having a 
`type SleeperFunc func(time.Duration)`
and possibly a 
`type RandomizerFunc func(int) int`

jwolski: Oh nice! I like it.

swim/join_sender.go
jwolski: yea, I thought about an abstraction like that and decided not to go that far initially. the more I think about it, the more I like it. I'll give it a go.

swim/join_sender.go
jwolski: fair point.

swim/join_sender.go
thanodnl: On a second read, I don't think this is useful. When I was reviewing this I thought it was in a hot code path, but this is only being run on startup.

jwolski: got it. this code is going to undergo some revisioning after you and @CorgiMan gave me some ideas. So stay tuned.

swim/join_delayer.go
```diff
@@ -0,0 +1,203 @@
...
+
+var errAttemptNumberInvalid = errors.New("joinDelayer requires attemptNumber to be greater or equal to 0")
+var errLoggerRequired = errors.New("joinDelayer requires a logger")
+
+const (
+	defaultInitial = float64(100 * time.Millisecond)
+	defaultMax     = float64(60 * time.Second)
+)
+
+var defaultRandomizer = rand.Intn
```
thanodnl: this randomizer uses the default source that has deterministic behavior. Lets make our own [`Rand`](https://golang.org/pkg/math/rand/#New) with a newly seeded [`Source`](https://golang.org/pkg/math/rand/#NewSource). To prevent interference with the `Source` of the application we are embedded in -that application might rely on the fact that random numbers are deterministic- and at the same time prevent all ringpops to use the same 'random' timeouts.

jwolski: It's a little bit more complicated then that. join_sender.go calls `rand.Seed()` and I was relying on that as the defaultRandomizer's seed. But that's non-obvious and error prone. I'll move the call to Seed() to the delayer itself.

jwolski: On second thought, I'll leave the call to Seed() in join_sender.go and create a new instance of type `Rand` with its own source.

thanodnl:
lgtm

swim/join_delayer.go
```diff
@@ -0,0 +1,204 @@
...
+		nextDelayMin:    0,
+		maxDelayReached: false,
+		maxDelay:        opts.max,
+		randomizer:      randomizer,
+		sleeper:         sleeper,
+	}, nil
+}
+
+// nullDelayer is an empty implementation of joinDelayer.
+type nullDelayer struct {
```
CorgiMan: Why is the nullDelayer crammed in between the exponentialDelayer initializer and method?

jwolski: I don't know. I'll move it and make it a bit more organized.

swim/join_delayer.go
```diff
@@ -0,0 +1,204 @@
...
+		randomizer: defaultRandomizer,
+		sleeper:    defaultSleeper,
+	}
+}
+
+// exponentialDelayer applies a delay in between repeated join attempts.
+// The delay increases exponentially and is capped at maxDelay.
+type exponentialDelayer struct {
+	// joiner is the address of the node sending join requests.
+	joiner string
```
CorgiMan: I am wondering if we could abstract everything join completely away from the delayer. It seems to me like the only thing we use join for is logging. We could make make this function emit events and do the logging in a handler that lives in the joiner. It doesn't seem like this is important right now but if we need different delayers I think we should make it more general.

swim/join_delayer.go
```diff
@@ -0,0 +1,204 @@
...
+// delayRandomizer is a function that returns a random number between
+// 0 and an upper-bound, provided in its first argument.
+type delayRandomizer func(int) int
+
+// delaySleeper is a function that pauses execution for time.Duration.
+type delaySleeper func(time.Duration)
+
+// delayOpts is a struct that houses substitutable parameters of a joinDelayer
+// in order to alter its behavior.
+type delayOpts struct {
```
CorgiMan: We use the full word Options in other places in the code. And would configuration be a more correct name?

swim/join_delayer.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// Delay delays a join attempt by sleeping for an amount of time. The
+// amount of time is computed as an exponential backoff based on the number
+// of join attempts that have been made at the time of the function call;
+// the number of attempts is 0-based. It returns a time.Duration equal to the
+// amount of delay applied and an error if one occurs. If an error occurs, a
+// duration equal to 0 will be returned, signifying no delay has been applied.
+func (d *exponentialDelayer) delay(attemptNumber int) (time.Duration, error) {
+	if attemptNumber < 0 {
```
CorgiMan: We could also clamp or panic if < 0. Not a bit deal though.

swim/join_delayer.go
jwolski: This is a good point, but I'll punt on the idea. If we have use for a delayer elsewhere, let's do that refactoring then. Right now, we aren't going to need it.

swim/join_delayer.go
jwolski: Truth. Good suggestion. I'll leave it as is though.

swim/join_delayer.go
jwolski: :+1:

swim/join_delayer.go
CorgiMan: Agreed

swim/join_delayer.go
```diff
@@ -0,0 +1,203 @@
...
+type delayRandomizer func(int) int
+
+// delaySleeper is a function that pauses execution for time.Duration.
+type delaySleeper func(time.Duration)
+
+// delayOpts is a struct that houses substitutable parameters of a joinDelayer
+// in order to alter its behavior.
+type delayOpts struct {
+	initial    float64
+	max        float64
```
CorgiMan: We can improve type safety by making initial and max be of type time.Duration.

swim/join_delayer.go
CorgiMan: Btw, why isn't attemptNumber part of the delayer, and raised every time delay is called?

swim/join_delayer.go
jwolski: The types that are needed for actual computation are floats so I chose to go with them instead even though from an API perspective time.Duration is better. Let me fix.

swim/join_delayer.go
jwolski: This is a good question. I like the idea of taking attemptNumber out of the signature.

CorgiMan:
Tests are failing

CorgiMan:
Looks good!

PR #69, opened 2016-01-15
Dedupe destinations before forwarded request is retried

jwolski:
There exists a bug in Ringpop's forwarding mechanism that will return a "divergence error" if any request containing multiple keys is ever retried. This happens to be the case because LookupKeys does not dedupe destinations when a lookup is attempted just prior to the retry.

@uber/ringpop

### DIFF
--- a/forward/request_sender.go
+++ b/forward/request_sender.go
@@ -33,6 +33,10 @@ import (
 	"github.com/uber/tchannel-go/raw"
 )
 
+// errDestinationsDiverged is an error that is returned from AttemptRetry
+// if keys that previously hashed to the same destination diverge.
+var errDestinationsDiverged = errors.New("key destinations have diverged")
+
 // A requestSender is used to send a request to its destination, as defined by the sender's
 // lookup method
 type requestSender struct {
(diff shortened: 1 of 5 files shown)


test/mocks/ringpop.go
```diff
@@ -8,6 +8,7 @@ import "github.com/uber/ringpop-go/events"
 import "github.com/uber/ringpop-go/forward"
 
 import "github.com/uber/ringpop-go/swim"
+
```
jwolski: Regenerating the mocks with `make mocks` added this line in.

forward/request_sender.go
```diff
@@ -203,8 +213,8 @@ func (s *requestSender) AttemptRetry() ([]byte, error) {
 
 	dests := s.LookupKeys(s.keys)
 	if len(dests) > 1 || len(dests) == 0 {
```
thanodnl: can we make this check easier via `len(dests) != 1`?

test/mocks/ringpop.go
thanodnl: where did this came from? different version of `mockery`?

forward/request_sender.go
jwolski: sure.

test/mocks/ringpop.go
jwolski: possibly. which version are we supposed to use?

thanodnl: turns out that we use the same version and that I am generating the white line as well now. Probably something to do with a change in the source file that this generates from. Nothing to do here.

thanodnl:
lgtm

forward/request_sender.go
```diff
@@ -230,14 +240,30 @@ func (s *requestSender) RerouteRetry(destination string) ([]byte, error) {
...
+	destSet := make(map[string]struct{})
 	for _, key := range keys {
 		dest, err := s.sender.Lookup(key)
 		if err != nil {
+			// TODO Do something better than swallowing these errors.
 			continue
 		}
-		dests = append(dests, dest)
+
+		if _, ok := destSet[dest]; ok {
```
CorgiMan: This seems unnecessary

PR #70, opened 2016-01-18
Feature/module logger

severb:
The pull request adds a module logger implementation and uses it to log SWIM messages only above WARN level. Additional module loggers can be added in a similar way.


util/module_logger_test.go
```diff
@@ -0,0 +1,258 @@
...
+func (suite *ModuleTestSuite) SetupTest() {
+	suite.dl = &dummyLogger{}
+	suite.ml = NewModuleLogger(suite.dl)
+	suite.ml.SetModuleLevel("testdebug", DebugLevel)
+	suite.ml.SetModuleLevel("testpanic", PanicLevel)
+	suite.ml.SetModuleLevel("testoff", OffLevel)
+}
+
+func (suite *ModuleTestSuite) assertMsg(msgLevel Level, args ...interface{}) {
+	suite.dl.assertMsg(suite.T(), msgLevel, args...)
```
jwolski: Is it possible to replace these methods by using testify's mock AssertCalled and AssertNotCalled methods?

util/module_logger_test.go
```diff
@@ -0,0 +1,258 @@
...
+	m2.Debug("m2")
+	suite.assertNoMsg()
+	m2.Info("m2")
+	suite.assertMsg(InfoLevel, "m2")
+}
+
+// Make sure the logger returned by WithField is still wrapped.
+func (suite *ModuleTestSuite) TestWithField() {
+	l := suite.ml.GetLogger("testpanic")
+	new_l := l.WithField("field", 1)
```
jwolski: Do not use snake-case.

util/module_logger_test.go
```diff
@@ -0,0 +1,258 @@
...
+	dlf := new(dummyLogFields)
+	new_l := l.WithFields(dlf)
+	assert.Equal(suite.T(), suite.dl.fields, dlf)
+	new_l.Fatal("Fatal Msg", 1, 2)
+	suite.assertNoMsg()
+	new_l.Panic("Panic Msg", 1, 2)
+	suite.assertMsg(PanicLevel, "Panic Msg", 1, 2)
+}
+
+func (suite *ModuleTestSuite) TestPreconditions() {
```
jwolski: The convention we've been using is to use `s` for the test suite variable and not `suite`.

util/module_logger_test.go
severb: Yeah, I saw when writing the tests for options that there is a Logger mock already; I'll rewrite this to use that instead.

util/module_logger_test.go
severb: Old habits die hard

util/module_logger_test.go
severb: :+1:

util/module_logger_test.go
```diff
@@ -0,0 +1,258 @@
...
+	assert := assert.New(suite.T())
```
jwolski: The Suite struct embeds `*assert.Assertions` and this gives access to all of its methods. I advise you to change:

```
assert.NoError
assert.Error
```

to 

```
s.NoError
s.Error
```
```

File: util/module_logger_test.go (hunk 2 above)
Comment (jwolski) [inline util/module_logger_test.go]: :)

File: util/module_logger_test.go (hunk 1 above)
Comment (jwolski) [inline util/module_logger_test.go]: cool

**Comment by dansimau** [comment]
General feedback on this PR: it feels like we are reimplementing logrus. Why don't we just use logrus?

**Comment by jwolski** [comment]
@dansimau I'd be in favor of using an existing thing. @severb Can you look to see what it'd take to make use of what logrus already provides and what changes would be necessary to make Ringpop use it? I'm a little bit worried about the scope this'd add, but the added scope might be worthwhile if we can avoid having to do a custom logging wrapper in Ringpop.

**Comment by severb** [comment]
Logrus provides a min. level per logger but it doesn't have any multi-logger configuration/factory implementation. We can use different logrus loggers for each module.

File: ringpop.go [hunk 5]
```diff
@@ -89,8 +89,7 @@ type Ringpop struct {
 		sync.RWMutex
 	}
 
-	logger log.Logger
-	log    log.Logger
+	moduleLogger *util.ModuleLogger
```
dansimau: Any reason why we would name this `moduleLogger` over something simpler like `logger` or just `log`?

util/module_logger.go
```diff
@@ -0,0 +1,260 @@
+package util
```
dansimau: It feels to me like the module logger should be in its own package. This would help us keep strong separation of concerns, particularly if we want to move or release the module logger later as a standalone tool.

util/module_logger.go
```diff
@@ -0,0 +1,260 @@
...
+
+import (
+	"fmt"
+	log "github.com/uber-common/bark"
+)
+
+// The level definition is stolen from logrus.
```
dansimau: Should we just import and use the logrus levels?

ringpop.go
severb: I think it hints that it's more than a simple logger; but I don't like ModuleLogger either.

util/module_logger_test.go
```diff
@@ -0,0 +1,253 @@
...
+import (
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/suite"
+	"github.com/uber-common/bark"
+	"github.com/uber/ringpop-go/test/mocks"
+	"github.com/uber/ringpop-go/util"
+	"testing"
+)
+
+type MockLogger struct {
```
dansimau: What does a `MockLogger` do that's different to `mocks.Logger`? Can we document this a comment? Also, does it need to be exported (I am thinking not)?

util/module_logger.go
severb: Good idea, I'll move it in its own module.

util/module_logger.go
severb: The logrus levels are in reverse order and don't implement OffLevel. The problem with the reverse order is that an uninitialized level defaults to PanicLevel.

util/module_logger_test.go
severb: Sure, I'll add a comment and not export it.

util/module_logger.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// ModuleLogger is a flat sub-logger configuration and factory. Each sub-logger
+// can be restricted to emit messages above a minimum level or use a different
+// underlying logger. The configuration can be changed on the fly even for
+// existing sub-loggers.
+// Method calls that return other loggers return instead a ModuleLogger
+// instance. This simplifies porting existing code unaware of module logging.
+type ModuleLogger struct {
+	l      func(*RestrictedLogger) *RestrictedLogger
```
dansimau: What is `l`? Can we use a more descriptive name? And/or comment on its purpose might be a good idea.

severb: Sure, I'll add a comment. It's actually to work around golang's lack of virtual methods :cry:

options.go
```diff
@@ -120,8 +121,29 @@ func HashRingConfig(c *hashring.Configuration) Option {
...
+// with a level less than minLevel are silenced. By default, an unconfigured
+// module outputs all messages.
+func ModuleLevel(name string, minLevel rutil.Level) Option {
+	return func(rp *Ringpop) error {
+		rp.moduleLogger.SetLevel(name, minLevel)
+		return nil
+	}
+}
+
+// ModuleLogger is used to configure different loggers per module.
```
dansimau: Can we include examples here? Note that this godoc is important, since it will be displayed in the ringpop-go documentation and seen by our customers.

severb: :+1:

options.go
```diff
@@ -120,8 +121,29 @@ func HashRingConfig(c *hashring.Configuration) Option {
...
+			r.moduleLogger.SetLogger(rutil.RootModule, l)
+		}
+		return nil
+	}
+}
+
+// ModuleLevel is used to configure different log levels per module. Messages
+// with a level less than minLevel are silenced. By default, an unconfigured
+// module outputs all messages.
+func ModuleLevel(name string, minLevel rutil.Level) Option {
```
dansimau: I feel like maybe the logging configuration should be passed as a struct.

Functional options give us a couple of advantages:

* We can take in runtime objects as options (e.g. logger created at runtime)
* Options can be optional

In the case of some configuration, I think we should just take a struct if we're only receiving config values. For example, note the `HashRingConfig` option which takes a `hashring.Configuration` type.

It will help us to have a more typed configuration when we want to take configuration from other places later on (e.g. from admin endpoints).

The way I imagine the configuration for the logging is something like:

Logging config for each logger:
```
package modulelogger

type LoggerConfig struct {
	Level int
}
```

Ringpop would have its own logging configuration, with an individual log config for each logger in the project:
```
package ringpop

type LoggerConfiguration struct {
	swim    modulelogger.LoggerConfig
	replica modulelogger.LoggerConfig
	gossip  modulelogger.LoggerConfig
}
```

The logging configuration can be passed in during Ringpop's construction, via a `LogConfig` option:
```
func LogConfig(c *LoggerConfiguration) Option {
	return func(r *Ringpop) error {
		...
	}
}
```

options.go
```diff
@@ -195,11 +217,16 @@ func defaultHashRingOptions(r *Ringpop) error {
 	return HashRingConfig(defaultHashRingConfiguration)(r)
 }
 
+func defaultSWIMLevel(r *Ringpop) error {
```
dansimau: If we use a single config struct for the logging config, we can just change this to `defaultLogConfiguration` and include logging config for all loggers (not just swim).

util/module_logger.go
```diff
@@ -0,0 +1,260 @@
...
+				}
+				rl = &RestrictedLogger{
+					logger: l,
+					min:    level}
+			}
+			return ml.l(rl)
+		}}
+}
+
+func (ml *ModuleLogger) WithField(key string, value interface{}) log.Logger {
```
dansimau: Please can we leave a comment explaining what this method is doing? Also note that it is exported, so the comment will show up in the documentation.

util/module_logger.go
```diff
@@ -0,0 +1,260 @@
...
+	return &ModuleLogger{
+		levels: ml.levels,
+		logs:   ml.logs,
+		l: func(rl *RestrictedLogger) *RestrictedLogger {
+			return ml.l(rl).WithField(key, value)
+		}}
+}
+
+func (ml *ModuleLogger) WithFields(fields log.LogFields) log.Logger {
```
dansimau: Same as above; needs comment.

util/module_logger_test.go
```diff
@@ -0,0 +1,253 @@
...
+	return ml.Logger.On(methodName, format, arguments)
+}
+
+func newMockLogger() *MockLogger {
+	return &MockLogger{Logger: &mocks.Logger{}}
+}
+
+type ModuleLoggerSuite struct {
+	suite.Suite
+	ml *util.ModuleLogger
```
dansimau: Note that the `ml` name for "module logger" got me confused with the mock logger when reading the tests. Maybe `ml` should just be `logger` or `baseLogger` or something.

ringpop.go
dansimau: Is the type enough of a hint of this? I'd prefer to keep the name simpler, if possible.

util/module_logger.go
dansimau: That sucks. Let's do our own then, but perhaps we don't need to mention logrus, particular if we have added/changed the levels.

dansimau:
As talked about this morning, we're also missing a lot of loggers. Grab me if you need a help creating a list of loggers. We're going to have to go through the code base and implement these. Is that something you expect to be part of this PR?

options.go
severb: I agree that making the config static makes it easy to see what and how can be changed.

We should include a `log.Logger` field in `LoggerConfig` for per module loggers and a `defaultLogger` to `LoggerConfiguration`.

Do you think it's worthwhile to keep `Logger`, `ModuleLevel` and `ModuleLogger` for one off updates at runtime?

util/module_logger.go
severb: Yes, sorry, the implementation is a bit opaque, I'll add a comment trying to explain what's going on.

The method is exported because it's part of `log.Logger` interface, and it's basically a wrapper.

severb:
I've started to group all existing log messages in modules but I think it's best to track this effort and and the addition of new messages separately since this PR is already starting to get hard to follow.

ringpop.go
severb: Maybe `mLog`? It looks a bit like Hungarian notation, but I think calling methods outside of `log.Logger` on a variable called `log` may be a bit surprising.

util/module_logger.go
severb: :+1:

dansimau:
@severb Cool, makes sense.

util/module_logger_test.go
severb: Good catch!

options.go
dansimau: Maybe we should have another chat to go through it. You'll need to explain to me what a "per module" logger is, and why we would want to pass one at runtime.

modulelogger/examples_test.go
```diff
@@ -0,0 +1,127 @@
+package modulelogger
```
jwolski: can you call the package logging? calling it modulelogger is kind of just an implementation detail of the logging facility.

modulelogger/module_logger.go
```diff
@@ -0,0 +1,278 @@
+package modulelogger
+
+import (
+	"fmt"
+	log "github.com/uber-common/bark"
+)
+
+type Level uint8
```
jwolski: Can you split `Level` `String()` `ParseLevel()` out into a `logging/level.go` file?

modulelogger/module_logger.go
```diff
@@ -0,0 +1,278 @@
...
+		levels: make(map[string]Level),
+		logs:   make(map[string]log.Logger),
+		getLogger: func(rl *RestrictedLogger) *RestrictedLogger {
+			return rl
+		}}
+	ml.SetLogger(RootModule, logger)
+	return ml.Logger(RootModule)
+}
+
+type RestrictedLogger struct {
```
jwolski: can you split `RestrictedLogger` and `ModuleLogger` out into its own files?

options_test.go
```diff
@@ -61,9 +62,10 @@ func (s *RingpopOptionsTestSuite) TestDefaults() {
 	testRingpop := &Ringpop{}
 	defaultStatter(testRingpop)
 	defaultLogger(testRingpop)
+	defaultSWIMLevel(testRingpop)
 	defaultHashRingOptions(testRingpop)
 
-	s.Equal(testRingpop.logger, rp.logger)
+	//s.Equal(testRingpop.moduleLogger, rp.moduleLogger)
```
jwolski: eh? what happened here?

modulelogger/module_logger_test.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func newMockLogger() *mockLogger {
+	return &mockLogger{Logger: &mocks.Logger{}}
+}
+
+type ModuleLoggerSuite struct {
+	suite.Suite
+	// unit under test
+	uut *modulelogger.ModuleLogger
```
jwolski: can you rename `uut` to something like `logger`? abbreviations/acronyms are hard to remember. `uut` is uncommon as far as I'm aware.

modulelogger/module_logger_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+func (s *ModuleLoggerSuite) SetupTest() {
+	s.log = newMockLogger()
+	s.uut = modulelogger.New(s.log)
+	err := s.uut.SetLevel("testFatal", modulelogger.FatalLevel)
+	s.Require().NoError(err)
+}
+
+func (s *ModuleLoggerSuite) TestPassThrough() {
+	s.log.On("Debug", "d msg", 1)
```
jwolski: please DRY this up. There's lot of repetition.

modulelogger/module_logger_test.go
```diff
@@ -0,0 +1,257 @@
...
+	withFieldLogger.On("Debug", "d msg")
+	withFieldsLogger.On("Warn", "w msg")
+	s.uut.WithField("key", "value").Debug("d msg")
+	s.uut.WithFields(fields).Warn("w msg")
+	s.Exactly(s.uut.Fields(), fields)
+	s.log.AssertExpectations(s.T())
+}
+
+func (s *ModuleLoggerSuite) TestLazyFields() {
+	l := s.uut.Logger("test")
```
jwolski: Also dry this up.

modulelogger/module_logger.go
```diff
@@ -0,0 +1,278 @@
...
+// can be restricted to emit messages above a minimum level or use a different
+// underlying logger. The configuration can be changed on the fly even for
+// existing sub-loggers.
+// Method calls that return other loggers return instead a ModuleLogger
+// instance. This simplifies porting existing code unaware of module logging.
+type ModuleLogger struct {
+	// This is as a poor man's virtual method; all method calls from
+	// log.Logger interface delegate to the result of this function.
+	// Usually, this function is actually a closure, chain calling other
+	// closures, until the unterlying logger wrapped in a RestrictedLogger
```
jwolski: typo `underlying`

severb:
I'm closing this in favor of https://github.com/uber/ringpop-go/pull/79.

PR #74, opened 2016-01-19
Ws ping reqs bug fix

CorgiMan:
Restructure ping-reqs and put them into a indirectPing -> (reached bool, errs []errors) function
The test is said to be inconclusive if **all** nodes that should ping the target are unreachable.

Add tests for all seven possible ping-req scenarios.

Add a ListenerFunc type to the swim/events.go that turns a function into a object that implements EventListener.

### DIFF
--- a/swim/node.go
+++ b/swim/node.go
@@ -236,11 +236,14 @@ func (n *Node) Incarnation() int64 {
 
 func (n *Node) emit(event interface{}) {
 	for _, listener := range n.listeners {
-		go listener.HandleEvent(event)
+		listener.HandleEvent(event)
 	}
 }
 
-// RegisterListener adds a listener that will receive swim events.
+// RegisterListener adds an EventListener to the node. When a swim event e is
+// emitted, l.HandleEvent(e) is called for every registered listener l.
+// Attention, all listeners are called synchronously. Be careful with
+// registering blocking and other slow calls.
 func (n *Node) RegisterListener(l EventListener) {
 	n.listeners = append(n.listeners, l)
 }
@@ -333,7 +336,7 @@ type BootstrapOptions struct {
 	// Minimum number of nodes to join to satisfy a bootstrap.
 	JoinSize int
 
-	// Maximum time to attempt joins before the entire boostrap process times
+	// Maximum time to attempt joins before the entire bootstrap process times
 	// out.
 	MaxJoinDuration time.Duration
 
@@ -466,30 +469,26 @@ func (n *Node) pingNextMember() {
 	}
 
 	// ping failed, send ping requests
-	var errs []error
-	for response := range sendPingRequests(n, member.Address, n.pingRequestSize, n.pingRequestTimeout) {
-		switch res := response.(type) {
-		case *pingResponse:
-			if res.Ok {
-				n.log.WithField("target", member.Address).Info("ping request target reachable")
-				n.memberlist.Update(res.Changes)
-				return
-			}
+	target := member.Address
+	targetReached, errs := indirectPing(n, target, n.pingRequestSize, n.pingRequestTimeout)
 
-			n.log.WithField("target", member.Address).Info("ping request target unreachable")
-			n.memberlist.MakeSuspect(member.Address, member.Incarnation)
-			return
-
-		case error:
-			errs = append(errs, res)
-		}
+	// if all helper nodes are unreachable, the indirectPing is inconclusive
+	if len(errs) == n.pingRequestSize {
+		n.log.WithFields(log.Fields{
+			"target":    target,
+			"errors":    errs,
+			"numErrors": len(errs),
+		}).Warn("ping request inconclusive due to errors")
+		return
 	}
 
-	n.log.WithFields(log.Fields{
-		"target":    member.Address,
-		"errors":    errs,
-		"numErrors": len(errs),
-	}).Warn("ping request inconclusive due to errors")
+	if !targetReached {
+		n.log.WithField("target", target).Info("ping request target unreachable")
+		n.memberlist.MakeSuspect(member.Address, member.Incarnation)
+		return
+	}
+
+	n.log.WithField("target", target).Info("ping request target reachable")
 }
 
 // GetReachableMembers returns a slice of members currently in this node's
--- a/swim/ping_request_sender.go
+++ b/swim/ping_request_sender.go
@@ -82,6 +82,9 @@ func (p *pingRequestSender) SendPingRequest() (*pingResponse, error) {
 	var res pingResponse
 	select {
 	case err := <-p.MakeCall(ctx, &res):
+		if err == nil {
+			p.node.memberlist.Update(res.Changes)
+		}
 		return &res, err
 
 	case <-ctx.Done(): // call timed out
@@ -112,6 +115,29 @@ func (p *pingRequestSender) MakeCall(ctx json.Context, res *pingResponse) <-chan
 	return errC
 }
 
+// indirectPing is used to check if a target node can be reached indirectly.
+// The indirectPing is performed by sending a specifiable amount of ping
+// requests nodes in n's membership.
+func indirectPing(n *Node, target string, amount int, timeout time.Duration) (reached bool, errs []error) {
+	resCh := sendPingRequests(n, target, amount, timeout)
+
+	// wait for responses from the ping-reqs
+	for result := range resCh {
+		switch res := result.(type) {
+		case *pingResponse:
+			if res.Ok {
+				return true, errs
+			}
+			// If the ping to the target was not-ok we want to wait for more results.
+
+		case error:
+			errs = append(errs, res)
+		}
+	}
+
+	return false, errs
+}
+
 // sendPingRequests sends ping requests to the target address and returns a channel
 //containing the responses. Responses can be one of type:
 //  (1) error:          if the call to peer failed
@@ -137,6 +163,8 @@ func sendPingRequests(node *Node, target string, size int, timeout time.Duration
 		wg.Add(1)
 
 		go func(peer Member) {
+			defer wg.Done()
+
 			p := newPingRequestSender(node, peer.Address, target, timeout)
 
 			p.node.log.WithFields(log.Fields{
@@ -146,19 +174,28 @@ func sendPingRequests(node *Node, target string, size int, timeout time.Duration
 
 			var startTime = time.Now()
 			res, err := p.SendPingRequest()
+
 			if err != nil {
-				resC <- err
-			} else {
-				node.emit(PingRequestsSendCompleteEvent{
-					Local:    node.Address(),
-					Target:   target,
-					Peers:    peerAddresses,
-					Duration: time.Now().Sub(startTime),
+				node.emit(PingRequestSendErrorEvent{
+					Local:  node.Address(),
+					Target: target,
+					Peers:  peerAddresses,
+					Peer:   peer.Address,
 				})
-				resC <- res
+
+				resC <- err
+				return
 			}
 
-			wg.Done()
+			node.emit(PingRequestsSendCompleteEvent{
+				Local:    node.Address(),
+				Target:   target,
+				Peers:    peerAddresses,
+				Peer:     peer.Address,
+				Duration: time.Now().Sub(startTime),
+			})
+
+			resC <- res
 		}(*peer)
 	}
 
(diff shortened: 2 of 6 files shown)


swim/ping_request_sender.go
```diff
@@ -112,6 +115,26 @@ func (p *pingRequestSender) MakeCall(ctx json.Context, res *pingResponse) <-chan
 	return errC
 }
 
+func indirectHealthCheck(n *Node, target string, size int, timeout time.Duration) (bool, []error) {
```
jwolski: call it indirectPing. Let's not introduce a new term when there doesn't need to be.

swim/node.go
```diff
@@ -218,7 +218,7 @@ func (n *Node) Incarnation() int64 {
 
 func (n *Node) emit(event interface{}) {
 	for _, listener := range n.listeners {
-		go listener.HandleEvent(event)
```
dansimau: Have we now just made all events synchronous? What is the impact of this?

CorgiMan:
Hold your horses on the review boys, there'll come some changes.

swim/node.go
CorgiMan: The only listeners that are registered are for stats and for hashring updates.

I think the hashring updates are better of being in sync, because we don't want the hashring to be out of sync with the membership.

CorgiMan:
Ready for review again @uber/ringpop

swim/ping_request_test.go
```diff
@@ -119,3 +123,250 @@ func (s *PingRequestTestSuite) TestTimesOut() {
...
+// status of the second ping is unknown.
+func TestIndirectPing1(t *testing.T) {
+	block := make(chan bool)
+
+	tnodes := genChannelNodes(t, 4)
+	defer destroyNodes(tnodes...)
+
+	sender, helper1, helper2, target := tnodes[0], tnodes[1], tnodes[2], tnodes[3]
+	bootstrapNodes(t, sender, helper1, helper2, target)
+	waitForConvergence(t, time.Second/2, sender, helper1, helper2, target)
```
dansimau: `500*time.Millisecond`?

dansimau:
Just a general piece of feedback: I found some of the ping tests took a lot of effort to read and understand. I feel that some more comments explaining what is going on would help the reader figure it out faster.

As a general rule, if it takes some effort to design the right pattern (as I know these ping/timings did), it will take some effort for the reader to comprehend. In these situations, being liberal with code comments  can significantly reduce the amount of time the reader needs to get up to speed with what the code is doing.

dansimau:
Oh, forgot to say: other than that, the code looks good! :smile:

swim/node.go
jwolski: Does this change have anything to do with the bugfix?

CorgiMan: I'm blocking upon receiving certain events to synchronize the ping reqs in the tests in the right way.

swim/ping_request_sender.go
```diff
@@ -112,6 +115,29 @@ func (p *pingRequestSender) MakeCall(ctx json.Context, res *pingResponse) <-chan
 	return errC
 }
 
+// indirectPing is used to check if a target node can be reached indirectly.
+// The indirectPing is performed by sending a specifiable amount of ping
+// requests nodes in n's membership.
+func indirectPing(n *Node, target string, amount int, timeout time.Duration) (reached bool, errs []error) {
```
jwolski: these return values need not be named: `(reached bool, errs []error)` -> `(bool, []error)`

CorgiMan: I think it increases the readability since one might think the first return value could mean isHealthy.
isHealthy has a different meaning than reached when the indirectPing is inconclusive.

I think the renaming of this function from indirectHealthCheck to indirectPing, has helped the interpretability of return value, so I'm okay with removing it. 

I'd still like your thoughts on why we benefit from removing the names though.

swim/node.go
danielheller: I think we should have a clear definition of this contract.  Generally, handling events async seems  safer—it allows the event handler implementation to act freely without concern for slowing down other mission-critical paths.  Of course, that does require careful reasoning about how tightly two systems in question are coupled.  If they do need to be tightly coupled to the point of making  operations in the two subsystems atomic, then I think event handling might not be the right pattern.  If synchronous event handling is the right pattern, that contract had better be extremely clearly documented...

danielheller:
I think the commit message here is a step in the right direction, but could be made more informative/accessible at very little cost in brevity, e.g. "Ping Reqs are defined to fail only if all N requests which are fanned out fail; this change addresses a bug which makes us treat the ping req as failed if *any* of the requests fail."

swim/node.go
dansimau: > If they do need to be tightly coupled to the point of making operations in the two subsystems atomic, then I think event handling might not be the right pattern. 

+1 Good observation.

swim/ping_request_sender.go
```diff
@@ -112,6 +115,29 @@ func (p *pingRequestSender) MakeCall(ctx json.Context, res *pingResponse) <-chan
...
+	for result := range resCh {
+		switch res := result.(type) {
+		case *pingResponse:
+			if res.Ok {
+				return true, errs
+			}
+			// If the ping to the target was not-ok we want to wait for more results.
+
+		case error:
+			errs = append(errs, res)
```
danielheller: Should we have default case with a panic here, or perhaps use an anonymous struct that we treat as a union?

swim/ping_request_sender.go
```diff
@@ -82,6 +82,9 @@ func (p *pingRequestSender) SendPingRequest() (*pingResponse, error) {
 	var res pingResponse
 	select {
 	case err := <-p.MakeCall(ctx, &res):
+		if err == nil {
+			p.node.memberlist.Update(res.Changes)
```
danielheller: Is this an unrelated bug that you discovered in the course of this change, i.e. that we aren't applying changes gossiped in ping reqs?  Should this be a separate diff with its own tests?

CorgiMan:
Noted, thanks for the insight. It is not entirely correct though. The bug is that the indirect check has the same result as the first ping-req that comes back (positive or negative). We talked about this and I misspoke then, hence the confusion.

swim/node.go
CorgiMan: Thanks for your thoughts Dan, you haven't been the only one who has expressed worries about the safety of synchronizing the events and it has made me realize my intuition on the observer pattern has been a bit off.

At this moment, events are used for loosely coupled interactions like statting, and strongly coupled interactions updating the hashring.

I strongly believe the hashring should be synchronized with the membership. I'm all for using a different pattern for membership updates anyway.

For testing it is very useful to synchronize the events like I'm doing here. As Simmo and Nils (I believe) have suggested before, having a separate RegisterSynchronized method can help us here.

Since both sync and async events are used in the code and there is no semantic distinction at this point, I'll make a quick writeup on some possible enhancements on the event system on Monday.

swim/ping_request_sender.go
CorgiMan: This was handled in the pingNextMember before, which worked since thats the only caller of sendPingRequests. Handling it there is an accident waiting to happen if we would ever send ping-reqs for another reason. I believe it's should be part of this diff because it's part of refactoring the ping-reqs mechanism.

swim/ping_request_sender.go
CorgiMan: I'm a fan of having a struct as a union here. I wanted to refactor only the things I had to in this diff. I'll be happy to refactor it in another diff, please let me know what you think.

swim/ping_request_sender.go
jwolski: Only reason I suggested removing them is because of the way I've seen Go's named return values previously used. They're used as a replacement actually returning values in various parts of a method and not to serve as documentation.

swim/ping_request_test.go
```diff
@@ -119,3 +123,250 @@ func (s *PingRequestTestSuite) TestTimesOut() {
...
+		cont <- true
+	}))
+
+	reached, errs := indirectPing(sender.node, targetHostPort, 2, time.Second)
+	assert.True(t, reached, "expected that target is reached")
+	assert.Len(t, errs, 0, "expected no errors from the helper nodes")
+}
+
+// The first ping-req doesn't reach the helper node. The second ping-req
+// does and the ping reached the target. The indirectPing therefor return
```
jwolski: `therefore`

swim/ping_request_test.go
```diff
@@ -119,3 +123,250 @@ func (s *PingRequestTestSuite) TestTimesOut() {
...
+	assert.Len(t, errs, 1, "expected one connection error from the helper nodes")
+}
+
+// Test where target is not bootstrapped and responds with not-readies,
+// indirectPing cannot reach the target and returns false.
+func TestIndirectPing5(t *testing.T) {
+	tnodes := genChannelNodes(t, 4)
+	defer destroyNodes(tnodes...)
+
+	// don't boostrap the target
```
jwolski: `bootstrap`

swim/ping_request_test.go
```diff
@@ -119,3 +123,250 @@ func (s *PingRequestTestSuite) TestTimesOut() {
...
+	helper2.channel.Close()
+	waitForTChannelClosed(sender.channel, helper1.node.Address())
+	waitForTChannelClosed(sender.channel, helper2.node.Address())
+
+	reached, errs := indirectPing(sender.node, targetHostPort, 2, time.Second)
+	assert.False(t, reached, "expected that target is unreachable")
+	assert.Len(t, errs, 2, "expected only errors from the helper nodes")
+}
+
+// onPingRequestReceive can be registered on an EventListener and fires the
```
jwolski: `onPingRequestComplete`

swim/ping_request_test.go
```diff
@@ -119,3 +123,250 @@ func (s *PingRequestTestSuite) TestTimesOut() {
...
+
+// Test where helper nodes are unreachable, indirectPing is inconclusive
+// and returns false to indicate that the target hasn't been reached.
+func TestIndirectPing7(t *testing.T) {
+	tnodes := genChannelNodes(t, 4)
+	defer destroyNodes(tnodes...)
+
+	bootstrapNodes(t, tnodes...)
+	waitForConvergence(t, time.Second/2, tnodes...)
+	sender, helper1, helper2, target := tnodes[0], tnodes[1], tnodes[2], tnodes[3]
```
jwolski: Can we make waitForTChannelClosed part of the testNode type?

```
helper1.closeAndWait(sender.tchannel)
```

I think it could help DRY things up a bit.

swim/ping_request_sender.go
CorgiMan: The go tour suggests otherwise: "These names should be used to document the meaning of the return values." https://tour.golang.org/basics/7

They indeed also add some other functionality, which I use a bit off here. (using errs).

swim/ping_request_test.go
CorgiMan: I should have acked this earlier

swim/ping_request_test.go
CorgiMan: ack

swim/ping_request_test.go
CorgiMan: ack

swim/ping_request_test.go
CorgiMan: Sure, I don't mind repeating myself in tests, but I don't mind changing it either

swim/ping_request_test.go
CorgiMan: Thanks

swim/ping_request_sender.go
jwolski: the tour.golang statement is a bit ambiguous.. but ok you win this time.

dansimau: Also see:
https://golang.org/doc/effective_go.html#named-results
https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

jwolski: Interesting. Effective Go makes it much more explicit. I really lose now.

CorgiMan:
@dansimau @jwolski @danielheller  I have addressed your feedback
(Typos, time.Second/2 -> 500*time.Millisecond, and DRYing up with closeAndWait function on testNode)

dansimau:
LGTM. Integration tests have landed in ringpop-go since this PR.

Is there a test in there for this fix? Is it passing?

CorgiMan:
There is not test for this yet. I'll see if I can add it.

CorgiMan:
It's not trivial to implement, let's leave it for another day

jwolski:
lgtm

PR #75, opened 2016-01-19
Fix swimCluster leaking channels

dansimau:
swimCluster was not closing channels created while instantiating nodes.
The fix is to keep track of the channels and close them when one calls
`swimCluster.Destroy`.

@uber/ringpop


benfleis:
Fixes on my machine. Diff lgtm.

jwolski:
yep, the solution works on my machine as well. No matter the open file limits (256+). ship it.

PR #78, opened 2016-01-22
Wait until generated incarnation number of joining node is fresh so t…

CorgiMan:
…hat it will not coincide with the incarnation of previous join

### DIFF
--- a/swim/clock.go
+++ b/swim/clock.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package swim
+
+import (
+	"time"
+)
+
+// Clock is the interface that wraps the Now method.
+type Clock interface {
+	Now() time.Time
+}
+
+// systemClock uses the system time to implement the Clock interface.
+type systemClock struct{}
+
+// Now returns the current system time.
+func (c systemClock) Now() time.Time {
+	return time.Now()
+}
+
+// NowInMillis is a utility function that call Now on the clock and converts it
+// to milliseconds.
+func NowInMillis(c Clock) int64 {
+	return c.Now().UnixNano() / int64(time.Millisecond)
+}
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -24,7 +24,6 @@ import (
 	"errors"
 
 	log "github.com/uber-common/bark"
-	"github.com/uber/ringpop-go/util"
 	"github.com/uber/tchannel-go/json"
 	"golang.org/x/net/context"
 )
@@ -122,7 +121,7 @@ func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
 }
 
 func (n *Node) adminJoinHandler(ctx json.Context, req *emptyArg) (*Status, error) {
-	n.memberlist.MakeAlive(n.address, util.TimeNowMS())
+	n.memberlist.Reincarnate()
 	return &Status{Status: "rejoined"}, nil
 }
 
--- a/swim/handlers_test.go
+++ b/swim/handlers_test.go
@@ -114,11 +114,18 @@ func (s *HandlerTestSuite) TestAdminLeaveJoinHandlers() {
 	s.Equal(4, s.testNode.node.CountReachableMembers())
 
 	// Test join handler brings it back to 5
+	s.testNode.node.clock = mockClock(time.Now().Add(time.Millisecond))
 	status, err = s.testNode.node.adminJoinHandler(s.ctx, &emptyArg{})
 	s.NoError(err, "calling handler should not result in error")
 	s.Equal(&Status{Status: "rejoined"}, status)
 	s.Equal(5, s.testNode.node.CountReachableMembers())
 }
+
+// mockClock implements the Clock interface with a constant Now method.
+type mockClock time.Time
+
+// Now returns the time that is stored in the type.
+func (c mockClock) Now() time.Time { return time.Time(c) }
 
 // TestRegisterHandlers tests that registerHandler always succeeds.
 func (s *HandlerTestSuite) TestRegisterHandlers() {
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -175,6 +175,12 @@ func (m *memberlist) GetMembers() (members []Member) {
 	return
 }
 
+// Reincarnate sets the status of the node to Alive and updates the incarnation
+// number. It adds the change to the disseminator as well.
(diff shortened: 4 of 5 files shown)


CorgiMan:
@uber/ringpop

CorgiMan:
Jeff added a clock interface that is passed to the joinHandler so that we can use a custom clock for testing.

jwolski:
One comment. I would take the Clock type out of the handlers file since it's generally useful. Other than that lgtm.

CorgiMan:
Sure thing

dansimau:
I hate to be a killjoy, but I find the mock clock stuff to be a little weird to read.

Why is the `Clock` interface defined in `handlers.go`? It seems if we are using some kind of clock then the interface is relevant to the entire swim package and therefore should be used all over the place.

I find the addition of many lines of code just to save sleeping for one ms a little overkill. Though, I would find it easier to accept if it was an abstraction used throughout the code, rather than just in this one handler.

Finally, I've never mocked time before so this code looks a little obscure without the context of our discussion. It might be worth commenting `Clock` to explain why this is happening.

swim/handlers.go
```diff
@@ -121,9 +121,22 @@ func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
 	return &ping{Checksum: n.memberlist.Checksum()}, nil
 }
 
-func (n *Node) adminJoinHandler(ctx json.Context, req *emptyArg) (*Status, error) {
-	n.memberlist.MakeAlive(n.address, util.TimeNowMS())
-	return &Status{Status: "rejoined"}, nil
+type Clock interface {
```
benfleis: I feel uneasy at adding a Clock interface in the middle of `swim/handlers.go` -- my feeling is that if we add a mechanism for mock/virtual clocks, that it should come from the top -- Ringpop->Node->handlers.

I recognize that it's a localized problem atm... but don't like the queasy feeling.

@jwolski @CorgiMan whatcha think? too much?

jwolski: i dont think it should be in `handlers.go` either. It's a generally useful abstraction.

jwolski:
@dansimau @CorgiMan @benfleis Now's not the time introduce the Clock type everywhere, but it is the time to stop making our test suite slower by continuing to rely on the passage-of-time to exercise test cases. It's our responsibility that when new code is written or old code refactored to make sure we proliferate the use of the Clock abstraction.

swim/handlers.go
```diff
@@ -121,9 +121,22 @@ func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
...
+
+type globalClock struct{}
+
+func (c globalClock) Now() time.Time { return time.Now() }
+
+type handler func(json.Context, *emptyArg) (*Status, error)
+
+func (n *Node) adminJoinHandler(clock Clock) handler {
+	return func(ctx json.Context, req *emptyArg) (*Status, error) {
+		tInMS := clock.Now().UnixNano() / int64(time.Millisecond)
```
jwolski: I'd go so far to say that `type Clock` could expose a `IncarnationNumber()` method which hides the conversion of time to milliseconds.

dansimau:
I agree with adding the `clock.go`. Not sure how I feel about having a clock dependency on something like, say, a request handler.

The fact that the incarnation number uses the clock has nothing to do with the handler function. It seems like we're adding a dependency to the handler function to make it testable because our internal implementation of something unrelated to the handler is less-than-ideal.

CorgiMan:
Time is truly and unintuitively a dependency, the problem is that we are not used to think of time that way because we experience it as something unchanging. The fact that only tests use it doesn't make it less of a dependency.

dansimau:
Yes, but it should be on the Node or something. It is not a direct dependency of the handler. It's in implicit dependency, via some other abstraction.

dansimau:
To elaborate:

The problem for me is that the responsibility of getting the next incarnation should not be that of the request handler. So, the API is wrong. Something better would be:

```
n.memberlist.MakeAlive(n.address, n.memberlist.nextIncarnation())
```

A better API could add a method to the memberlist that is responsible for generating the next incarnation number. This way, you're not leaking the implementation detail—I.e. that the incarnation number is currently using the clock—to the handler function, and thus not adding an indirect dependency (of `time`) to a handler function that has nothing to do with it.

I would argue that adding this weird indirect dependency is actually worse than sleeping for 1ms.

Some alternative solutions:

* Improve the API of `MakeAlive` so that there is some kind of incarnationNumber generator, which the tests can mock in the right place under the hood.
* Add a global clock that can be mocked in tests
* Just sleep for 1ms

My fear is that we end up injecting a clock into random functions and methods all over the place.

jwolski:
lgtm lol

swim/clock.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type Clock interface {
+	Now() time.Time
+	NowInMillis() int64
+}
+
+type systemClock struct{}
+
+func newSystemClock() Clock {
```
CorgiMan: I don't love how this returns a Clock and not a systemClock. Converting to the interface type doesn't need a cast anyway.

Also, we don't really need the constructor since we can just call &systemClock{}

swim/clock.go
```diff
@@ -0,0 +1,44 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package swim
+
+import (
+	"time"
+)
+
+type Clock interface {
```
CorgiMan: We need some comments in the file, I'll add some

swim/clock.go
```diff
@@ -0,0 +1,44 @@
...
+	NowInMillis() int64
+}
+
+type systemClock struct{}
+
+func newSystemClock() Clock {
+	return &systemClock{}
+}
+
+func (c systemClock) Now() time.Time {
```
CorgiMan: I'll add some comments here

swim/clock.go
```diff
@@ -0,0 +1,44 @@
...
+	return time.Now()
+}
+
+func (c systemClock) NowInMillis() int64 {
```
CorgiMan: I'll add some comments here

swim/clock.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"time"
+)
+
+type Clock interface {
+	Now() time.Time
+	NowInMillis() int64
+}
+
+type systemClock struct{}
```
CorgiMan: I'll add some comments here

swim/handlers_test.go
```diff
@@ -114,12 +114,21 @@ func (s *HandlerTestSuite) TestAdminLeaveJoinHandlers() {
...
+	s.testNode.node.clock = mockClock(time.Now().Add(time.Millisecond))
 	status, err = s.testNode.node.adminJoinHandler(s.ctx, &emptyArg{})
 	s.NoError(err, "calling handler should not result in error")
 	s.Equal(&Status{Status: "rejoined"}, status)
 	s.Equal(5, s.testNode.node.CountReachableMembers())
 }
 
+type mockClock time.Time
+
+func (c mockClock) Now() time.Time { return time.Time(c) }
```
CorgiMan: I'll add some comments here

swim/handlers_test.go
```diff
@@ -114,12 +114,21 @@ func (s *HandlerTestSuite) TestAdminLeaveJoinHandlers() {
...
+
+func (c mockClock) NowInMillis() int64 {
```
CorgiMan: I'll add some comments here

swim/handlers_test.go
```diff
@@ -114,12 +114,21 @@ func (s *HandlerTestSuite) TestAdminLeaveJoinHandlers() {
...
 
 	// Test join handler brings it back to 5
+	s.testNode.node.clock = mockClock(time.Now().Add(time.Millisecond))
 	status, err = s.testNode.node.adminJoinHandler(s.ctx, &emptyArg{})
 	s.NoError(err, "calling handler should not result in error")
 	s.Equal(&Status{Status: "rejoined"}, status)
 	s.Equal(5, s.testNode.node.CountReachableMembers())
 }
 
+type mockClock time.Time
```
CorgiMan: I'll add some comments here

swim/memberlist.go
```diff
@@ -175,6 +175,10 @@ func (m *memberlist) GetMembers() (members []Member) {
 	return
 }
 
+func (m *memberlist) Reincarnate() []Change {
```
CorgiMan: I'll add some comments here

swim/node.go
```diff
@@ -154,6 +162,7 @@ type Node struct {
 
 	logger log.Logger
 	log    log.Logger // wraps the provided logger with a 'local: address' field
+	clock  Clock
```
CorgiMan: I'll add some comments here saying that this clock is used to generate the incarnation numbers

CorgiMan:
lgtm too lololol

dansimau:
Summary of this pull request:
This pull request

LAND IT!

jwolski:
lgtm

![thumbsup]
(https://33.media.tumblr.com/d5d88fbb9eb22578e67556f7ca4c501e/tumblr_mvewnudGLK1r55ed0o2_400.gif)

CorgiMan:
I'll squash and land

PR #79, opened 2016-01-22
Feature/logger

severb:
A simplified module/named logger.

There are 3 major components:
* A `LogFacility` used to configure module log levels.
* A `LogFactory` used to create named loggers for each module in ringpop.
* A `Logger` which extends the bark.Logger interface with an aditional `Trace()`/`Tracef()` methods.


logger/interface.go
```diff
@@ -0,0 +1,33 @@
...
+
+package logger
+
+import (
+	"github.com/uber-common/bark"
+)
+
+// Similar to bark.Logger but with an extra Trace method.
+type Logger interface {
+	bark.Logger
```
severb: On a second thought, this won't work as expected: after a `WithFields` call, `Trace` won't be available anymore.

Two solutions come to mind:
* Drop the additional `Trace` and `Tracef` methods.
* Copy the `bark.Logger` interface and change `WithField` and `WithFields` to return `Logger` instead of `bark.Logger` followed by `s/bark.Logger/Logger/` in the codebase.

logger/factory.go
```diff
@@ -0,0 +1,84 @@
...
+	if named, ok := lf.cache[name]; ok {
+		named.setLevel(level)
+	} else {
+		lf.cache[name] = &namedLogger{
+			logger: lf.logger,
+			min:    level,
+		}
+	}
+	return nil
+}
```
severb: The implementation is shortsighted as it doesn't update loggers created by `WithField` or `WithFields`. For example:
```go
named := factory.Logger("named")
fielded := named.WithField("field", 1)
factory.SetLevel("named", Warning) // updates named but not fielded
```

logger/factory.go
```diff
@@ -0,0 +1,84 @@
...
+func NewLoggerFactory(l bark.Logger) *LoggerFactory {
+	return &LoggerFactory{
+		logger: l,
+		cache:  make(map[string]*namedLogger),
+	}
+}
+
+// SetLevel sets the minimum level for a named logger. A named logger emits
+// messages with a level equal to or greater than this level.
+func (lf *LoggerFactory) SetLevel(name string, level Level) error {
```
dansimau: I find it weird that `SetLevel` is a method of the factory. Shouldn't it be a method of the logger itself?

logger/factory.go
```diff
@@ -0,0 +1,84 @@
...
+	if level < lowestLevel {
+		return fmt.Errorf("level must be higher than %s", lowestLevel)
+	}
+	if level > highestLevel {
+		return fmt.Errorf("level must be lower than %s", highestLevel)
+	}
+	if named, ok := lf.cache[name]; ok {
+		named.setLevel(level)
+	} else {
```
dansimau: This seems slightly magic to me. I think if you try to set the level of something that doesn't exist, you should get an error, rather than a new logger being implicitly created. The latter behaviour could lead to subtle bugs yet I'm not sure what value it gives us.

logger/factory.go
dansimau: That doesn't seem like a problem to me, as long as the documentation is clear. `fielded` is a new logger, right? So why would changed the level of "named" affect that?

The only thing that seems a little weird is from my comment above: that the factory even has anything to do with the setting of levels. It feels to me like it would be more consistent to just do this on the loggers themselves.

Then in the above example you gave, the levels are both set in the same way:

```
named := factory.Logger("named")
fielded := named.WithField("field", 1)

named.SetLevel(Warning) // Updates the level of named
fielded.SetLevel(Warning) // Updates the level of fielded
```

logger/factory.go
```diff
@@ -0,0 +1,84 @@
...
+			min:    level,
+		}
+	}
+	return nil
+}
+
+// Logger returns a named logger. If no level was previously set for this
+// named logger it defaults to Warn.
+func (lf *LoggerFactory) Logger(name string) Logger {
+	// XXX: fix races
```
dansimau: I think we should fix the races before landing this...

logger/factory.go
```diff
@@ -0,0 +1,84 @@
...
+		named := &namedLogger{
+			logger: lf.logger,
+			min:    defaultNotConfiguredNamedLogger,
+		}
+		lf.cache[name] = named
+		return named
+	}
+}
+
+// Override the bark logger implementation with another one.
```
dansimau: Let's be specific in this comment to say that we're going to also update the logger on all the loggers in the factory. I think that's an important detail. Also a minor nit: we don't need to say that it's a bark logger in the comment. The type definition tells us that.

logger/factory_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+package logger
+
+import (
+	"github.com/stretchr/testify/suite"
+	"github.com/uber/ringpop-go/test/mocks"
+	"testing"
+)
+
+type LoggerFactorySuite struct {
```
dansimau: Nice one! These tests are really readable.

logger/interface.go
dansimau: You're right. Though the latter idea won't work on its own because then we won't implement `bark.Logger` anymore, which is a key requirement. I have a couple of ideas for how we solve it though. We should discuss on Monday.

I'd argue Trace is a nice-to-have though, so we should only keep it in if it's not going to be a headache.

logger/level.go
```diff
@@ -0,0 +1,89 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package logger
+
+import "fmt"
+
+type Level uint8
```
dansimau: Needs comment.

logger/level.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (level Level) String() string {
```
dansimau: Needs doc comment

logger/level.go
```diff
@@ -0,0 +1,89 @@
...
+		return Debug, nil
+	case "trace":
+		return Trace, nil
+	}
+
+	var l Level
+	return l, fmt.Errorf("not a valid Level: %q", lvl)
+}
+
+const (
```
dansimau: How come these levels are the opposite way to [logrus levels](https://github.com/Sirupsen/logrus/blob/master/logrus.go#L57)?

logger/named.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (rl *namedLogger) setLevel(level Level) {
+	rl.min = level
+}
+
+func (rl *namedLogger) setLogger(logger bark.Logger) {
+	rl.logger = logger
+}
+
+func (rl *namedLogger) shouldOutput(level Level) bool {
```
dansimau: Nice API here; produces readable code later on.

logger/named_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+import (
+	"github.com/stretchr/testify/suite"
+	"github.com/uber-common/bark"
+	"github.com/uber/ringpop-go/test/mocks"
+	"strings"
+	"testing"
+)
+
+type NamedLoggerSuite struct {
```
dansimau: Nice tests again, very readable

logger/level.go
dansimau: Also each level needs a godoc comment. And why is `unset` not exported?

logger/ringpop.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if conf.Damping != unset {
+		if err := rl.factory.SetLevel(dampingLogger, conf.Damping); err != nil {
+			return fmt.Errorf("Config.Damping: %v", err)
+		}
+	}
+	return nil
+}
+
+func (rl *RingpopLogger) Ringpop() Logger {
```
dansimau: All of these should have godoc comments

logger/ringpop.go
```diff
@@ -0,0 +1,136 @@
...
+
+	Ringpop    Level
+	Gossip     Level
+	Suspicion  Level
+	Ring       Level
+	Membership Level
+	Damping    Level
+}
+
+// RingpopLogger is a factory for ringpop named loggers. Each named logger can
```
dansimau: I like the clear separation here of the ringpop vs default factory implementation. Makes it very easy to understand what loggers we have in ringpop.

options.go
```diff
@@ -115,13 +114,24 @@ func HashRingConfig(c *hashring.Configuration) Option {
...
 }
 
-// Logger is used to specify a bark-compatible logger that will be used for
-// all Ringpop logging. If a logger is not provided, one will be created
-// automatically.
-func Logger(l log.Logger) Option {
+// Logger is used to specify a bark-compatible logger that will be used for all
+// Ringpop logging and the log levels for all modules. If a logger is not
+// provided, one will be created automatically.
+func Logger(conf logger.Options) Option {
```
dansimau: This breaks backwards-compatibility, which I'd rather not do.

Could we solve this by having `Logger` and `LoggingConfig` as separate options? It matches the `HashRingConfig` naming pattern. And we could just deprecate `Logger` later, if we really want.

dansimau:
Looking very nice. A lot more straightforward than previous PR. Thanks for persevering. I think it's good that we fret over this stuff for a new implementation of something.

My only other comment is that I tried to run the tests but it wouldn't compile:

```
$ go test -v
# github.com/uber/ringpop-go/logger
./factory.go:32: cannot use restricted (type *restrictedLogger) as type *bark.Logger in return argument:
	*bark.Logger is pointer to interface, not interface
./factory.go:39: cannot use restricted (type *restrictedLogger) as type *bark.Logger in return argument:
	*bark.Logger is pointer to interface, not interface
```

Also recommend running `golint`. There are a few suggestions in there which I'd pay attention to.

Nice work. :+1:

logger/ringpop.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"fmt"
+	"github.com/Sirupsen/logrus"
+	"github.com/uber-common/bark"
+	"io/ioutil"
+)
+
+// Options is used to set the ringpop logger implementation and the minimum
+// output level for each named logger.
+type Options struct {
```
dansimau: Should `Options` be `Configuration`? We should figure out a convention here. We have uses of both in various packages.

My vote would be for `Configuration` since it's just a struct; even though the name is longer, it's a less-overloaded term and shows a clear differentiation between a simple struct of config versus the more functional-style options.

logger/factory.go
severb: Not really, we want a centralized facility to configure all named loggers. Consider this:
```go
// in some function
xLogger := factory.Logger("x")
// ... use xLogger

// in another function
xLoggerAgain := factory.Logger("y")
// ... use xLoggerAgain
```
With the `SetLevel` method on the factory we can silence both `xLogger` and `xLoggerAgain` from an admin handler:
```go
factory.SetLevel("x", Off)
```

I'd say it would make sense have the `SetLevel` on the logger if we could somehow have a single logger for each module (something like `Ringpop.Gossip`, `Ringpop.Membership`, etc.). Unfortunately this won't work because of how extra metadata is built using `WithField` and `WithFields`.

logger/factory.go
severb: I think this is fine, otherwise you wouldn't be able to configure the factory a priori:
```go
factory := NewLoggerFactory(l)
factory.SetLevel("x", Debug)
factory.SetLevel("y", Warn)
// ... and later
xLogger := factory.Logger("x")
```

logger/factory.go
severb: Setting the level on an individual logger kind of defeats the purpose of having the factory. The levels is what we want to abstract by removing the `name: Level` mapping from user's sight. Another thing to keep in mind is that all loggers created by calling `WithField`/`WithFields` need to preserve the notion of name and level restriction, otherwise we won't be able to change the configuration for all loggers belonging to the same name.

Consider this (ringpop) usecase:
```go
named := factory.Logger("named")
named.WithField("a", 1).Debug("msg")
named.WithField("b", 2).Debug("msg")
```

If the level update isn't propagated to all loggers, then a subsequent `factory.SetLevel("x", Off)` won't silence the messages above.

I have an improved version that fixes propagation but I'm still working on the tests. Stay tuned.

logger/factory.go
severb: Agree

logger/factory.go
severb: :+1:

logger/factory_test.go
severb: I agree that they are easy to read but at the same time I'd say they are very brittle. Any change to the caching strategy, field rename or even to the `namedLogger` will break them. I usually prefer a more black-box approach for robust tests, but that usually translates in a lot of extra time spent on writing them.

logger/interface.go
severb: We don't really have to implement `bark.Logger`. The only place we care about `bark.Logger` is when an external logger is sent to us: `ringpop.New(..., Logger(logger.Options{logger: barkImplementation}))`. Inside ringpop we can use whatever interface we want for loggers.

logger/level.go
severb: Initially it was because I wanted the Level to default to Debug when unset, but I don't think the level direction matters. The levels should be refereed to by name/const.

`unset` is not exported because it has no meaning to the outside world and it's an error to call `factory.SetLevel("name", unset)`.

logger/named.go
severb: :metal:

logger/ringpop.go
severb: Yeah, I don't really have any strong preference. It used to be called `Config` a few commits ago. :smile:

options.go
severb: Yes, that's a great idea!

logger/factory_test.go
dansimau: I don't really understand what you mean. Why are they brittle?

In the example you give, if you change the caching strategy, from what I can tell in the tests below, nothing will break, because you haven't changed the contract of the `LoggerFactory.Logger` method.

Any change to field names will most certainly break, though. Fortunately it's just a find-replace fix which is usually quick. However, where possible, we should avoid using internals (like `min` below) and instead test through public APIs.

logger/factory.go
dansimau:
```
factory.SetLevel("x", Off)
```
 is equivalent to:

```
factory.Logger("x").SetLevel(Off)
```

So you can silence both your `xLogger` and `xLoggerAgain` with the same effort.

However, the latter makes a lot more sense to me: the level itself is not really a concern of the factory. The factory never stores it. The factory doesn't even care about the level; it's only pertinent to the logger. So why is the API not on the logger?

logger/factory.go
dansimau: That's a good point. It clearly does give us value then :smile:

logger/interface.go
dansimau: OK great. That sounds good

logger/level.go
dansimau: > but I don't think the level direction matters

Ha! If it doesn't matter, can we leave it the way it was when we took it from logrus? :)

> The levels should be refereed to by name/const

*Should* is the key word there.

> unset is not exported because it has no meaning to the outside world and it's an error to call factory.SetLevel("name", unset).

Good point. I forgot there is no longer a `GetLevel`.

logger/factory.go
severb: Oh, I see what you're saying. There are a few things to keep in mind:
* The factory interface makes no promises that `factory.Logger("x")` always returns the same instance.
* Is not enough to propagate the level change only on the logger returned by `factory.Logger("x")`. The level change must be propagated to all loggers subsequently created by `WithField`/`WithFields` calls. (i.e. `logger := factory.Logger("x").WithField("key", value)`)
* Once the level propagation fix lands the API will make more sense because the levels are stored in the factory itself and `namedLoggers` instances forward messages back to the factory. This arranges the code in a way that propagation (which turns out is difficult to achieve) is unnecessary.

But I do get your point, and for the current implementation I agree with it.

logger/factory_test.go
severb: I'll argue that as long as the internal details/implementation of a struct changes but the observable behavior for the outside world is unchanged, no test should fail. This property of tests is essential for code evolution and keeping false positives low.

logger/level.go
severb: With some extra boilerplate in this file we can go from *sholud* to *must* by adding an interface and not exporting `Level`. Maybe that's the way to go.

logger/factory_test.go
dansimau: I agree. But I don't understand how that statement relates to the readability of the tests.

I think you should just take the compliment ;)

severb: Haha, thanks!

logger/level.go
severb: Slightly changed the implementation to disallow levels to be created from outside the package.

severb:
Sorry for the noise, but in order to grow an extra `Trace()` log message I had to change the `bark.Logger` interface everywhere in the code.
This change should also include a mock for `logger.Logger` but I need help generating one. Until then, I fixed all tests that were failing by wrapping the `bark.Logger` mocks.

ringpop.go
```diff
@@ -89,8 +90,9 @@ type Ringpop struct {
 		sync.RWMutex
 	}
 
-	logger log.Logger
-	log    log.Logger
+	logFacility logger.LogFacility
```
jwolski: I think this is a bit of a smell. With the logging work, it's not any clearer why (at a glance) there is the distinction between `logger` and `log`, but now we have a 3rd nebulous concept in `logFacilty`. What's the distinction between all 3 of these? Do we need all of them?

logger/facility.go
```diff
@@ -0,0 +1,211 @@
...
+	Log(*namedMsg)
+	Logf(*namedfMsg)
+}
+
+type loggerFacilityFields struct {
+	recv   logReceiver
+	fields Fields
+}
+
+func (lff *loggerFacilityFields) Log(msg *namedMsg) {
```
jwolski: I'm a little concerned about these short-lived structs we're creating and how much garbage we might be creating especially at sites that log at the debug/trace levels. Is there a way to defer the creation of a `namedMsg` until we actually evaluate the log level?

ringpop.go
severb: `.log` and `.logger` were there from the beginning, as an early attempt to implement module logging from what I understand. There is no difference between them, and from what I remember only one gets the most use.

`.logFacility` is new, and isn't used anywhere yet. For now, to keep the code running, `log == logger == logFacility.Ringpop()` where `logFacility.Ringpop()` is a named logger.

jwolski: Let's get rid of one of `log` or `logger`.

logger/facility.go
severb: This is a good point. `namedMsg` is just a plain struct used move data. We can change `Log`'s signature to accept same data a `namedMsg` carries as arguments.

ringpop.go
severb: :+1:

logger/level.go
```diff
@@ -0,0 +1,107 @@
...
+	Warn
+	// Error log level
+	Error
+	// Fatal log level
+	Fatal
+	// Panic log level
+	Panic
+	// Off log level, turns off all logging.
+	// Use this carefuly as is it skips Panic calls.
+	Off
```
severb: I wonder if this is safe to have.

akshayjshah:
# :angel: I come in peace :angel: 

Ringpop friends,

I hear you like logging! Over here in RPC, we've been a bit dissatisfied with bark, logrus, and our go-common wrappers - they lack a few conveniences while also being relatively slow and garbage-producing. I've started to sketch out some ideas for a more performant structured logger, and I'm hoping to pull in all the secret scaling sauce that the node team have built into their logging libraries (defer work until the message is actually written out, run-time log level changes, very few allocations, and even reservoir sampling).

By all means, go ahead with this work...but I'd love to talk to you about your use case. If at all possible, I'd like to find/write a library that'll work for all of us down the road.

logger/level.go
jwolski: I think it's fine.

logger/facility.go
```diff
@@ -0,0 +1,211 @@
...
+func (lf *loggerFacility) WithFields(fields Fields) Factory {
+	return &loggerFacilityFields{
+		recv:   lf,
+		fields: fields,
+	}
+}
+
+func (lf *loggerFacility) shouldOutput(name string, level level) bool {
+	minLevel := defaultNotConfiguredNamedLogger
+	if confLevel, ok := lf.levels[name]; ok {
```
jwolski: access to the `levels` map should be made thread-safe.

logger/ringpop.go
```diff
@@ -0,0 +1,159 @@
...
+	Ring Level
+	// Minimum level to which the Membership logger is restricted.
+	Membership Level
+	// Minimum level to which the Damping logger is restricted.
+	Damping Level
+}
+
+// LogFactory provides a convenient way to produce named loggers and pass
+// metadata as fields between separate parts of the codebase.
+type LogFactory interface {
```
jwolski: Can you briefly explain how a dev would go about introducing a new logger into the system? I'm trying to determine if these conveniences are extraneous.

To me it looks like they'll have to:
1) Add a field to the Options struct for the new logging level
2) Expose the Logger in the LogFactory interface
3) Add a logger constant for its name
4) Add a condition to `Update()` to evaluate whether its Option has been specified
5) Implement a ringpopFacility logger method.
6) Implement a ringpopFactory logger method

Is that right?

logger/ringpop.go
```diff
@@ -0,0 +1,159 @@
...
+	// Ringpop retutrns the logger used to log messages outside of any module.
```
jwolski: Next question:

I see there are two ways of creating/accessing loggers: 1) through strongly-type LogFactory interface methods (like Ringpop() and Gossip()) and 2) through a weaker-typed Logger method (and providing a string)

Am I right? 

I'd rather provide only 1 way. Is there a need for both (other than for perceived flexibility)?

logger/ringpop.go
severb: Yes, that's a complete list.

Note that if we give up this static wrapper, including the static `Options` struct, adding a new named logger would require zero additional code.

logger/ringpop.go
severb: You're right, there's no need for both. `Factory` is more convenient because you can create named loggers on the fly, `LogFactory` on the other hand has the benefit of being static. I think I was unsure on the rate we plan to use new named loggers and I didn't knew which one to keep.

severb:
Close in favor of https://github.com/uber/ringpop-go/pull/83

PR #80, opened 2016-01-25
add ‘setup’ target to make, update README to match

benfleis:
Instructions to setup for local dev were incorrect. This updates them, and adds a make target to simplify.

@uber/ringpop

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean clean-mocks testpop mocks out test test-integration test-unit
+.PHONY: clean clean-mocks testpop mocks out setup test test-integration test-unit
 
 SHELL = /bin/bash
 
@@ -7,8 +7,10 @@ export PATH := $(shell pwd)/scripts/travis/thrift-gen-release/linux-x86_64:$(PAT
 
 # go commands should use the Godeps/_workspace
 GODEPS := $(shell pwd)/Godeps/_workspace
-OLDGOPATH := $(GOPATH)
-export GOPATH = $(GODEPS):$(OLDGOPATH)
+USER_GOPATH := $(GOPATH)
+export GOPATH = $(GODEPS):$(USER_GOPATH)
+
+DEV_DEPS = github.com/uber/tchannel-go/thrift/thrift-gen github.com/vektra/mockery/...
 
 out:	test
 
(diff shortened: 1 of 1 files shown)


dansimau:
Thanks for sorting this. It's been a pain for a few people.

Only thing I don't like about it is that it's not portable (it assumes Mac OS X due to the dependency on `brew`). For that reason, it's not really the Done Thing™ to put the installation of dependencies in a Makefile.

Then again, we're already abusing Makefiles in the first place...

benfleis:
Point taken, and I considered that as well, but still prefer a small step forward, ie, no worse than it already was.

I think it's not out of bounds when the depends are stable and easy; go specific deps are already forcibly structured and consistent. I could also do sth like this, to keep it arch neutral, but still very convenient:

```bash
setup:
    which thrift-gen | grep -q / || echo "No thrift-gen in your PATH. Please install it! (brew install thrift ?)" >&2
    go get $(DEV_DEPS)  
```

CorgiMan:
Thanks Ben, lgtm

dansimau:
I guess I could bend on the go deps, since that's cross-platform (e.g. `go get [...]`). I think we should still list the thrift requirement in the README, for the sake of people on other platforms (i.e. Linux).

I feel that for us to try and automatically install thrift, would should make it conditional on the platform, e.g.:

```
PLATFORM := $(shell uname)

setup:
ifeq "$(PLATFORM)" "Darwin"
    brew install thrift || { echo "Unable to install thrift" >&2; exit 1; }
endif
```

Then we:

 * Make it automatic for people on OS X (i.e. by doing `make setup`)
 * Don't break for people on other platforms, but they need to do a manual install of thrift

thanodnl:
I like the idea of at least testing the platform before calling `brew`. I'm just wondered if we need to add support for `ubuntu` as well, or if we will cross that bridge when somebody needs it.

dansimau:
LGTM, but we should test this again after rebasing off master. The way GOPATH works in the Makefile has changed. We should make sure `go get` is going to work properly, because I didn't test this...

thanodnl:
lgtm

benfleis:
Finally merging in. The README change got accidentally folded into master directly, but since it's a README, no harm, no foul, I hope :)

PR #81, opened 2016-01-27
Ensure non-zero exit on compile fail

dansimau:
This fixes an issue where tests with import errors were passing because
the number of failed tests was equal to 0 (because no tests ran).

@uber/ringpop


jwolski:
oops! lgtm

PR #82, opened 2016-01-27
Ws defensive piggybacking

CorgiMan:
Three commits, I recommend looking at them separately.

1. Removes the bumping of piggyback counters out of issueChange function. IssueAsSender now returns a callback that raises the pb counters, this callback is supposed to be called when the ping or ping-req has succeeded. IssueAsReceiver now increments the pb counters when called, because it is not easy to find out wether a response will reach the client.

2. Removes newPingRequest and just use the literal expression &pingRequest{...}

3. Removes filter from issueChanges because it was only used in issueAsReceiver and added a lot of unnecessary complexity.

### DIFF
--- a/swim/disseminator.go
+++ b/swim/disseminator.go
@@ -41,6 +41,9 @@ type pChange struct {
 type disseminator struct {
 	node    *Node
 	changes map[string]*pChange
+
+	// maxP indicates how many times a change is disseminated. It is declared
+	// in the swim paper as piggybackfactor * log(# nodes).
 	maxP    int
 	pFactor int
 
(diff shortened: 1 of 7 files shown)


swim/disseminator.go
```diff
@@ -109,63 +109,88 @@ func (d *disseminator) FullSync() (changes []Change) {
 	return changes
 }
 
-func (d *disseminator) IssueAsSender() []Change {
-	return d.issueChanges(nil)
+// IssueAsSender collects all changes a node needs when sending a ping or
+// ping-req. The second return value is a callback that raises the piggyback
+// counters of the given changes.
+func (d *disseminator) IssueAsSender() (changes []Change, bumpPiggybackCounters func([]Change)) {
+	return d.issueChanges(), d.bumpPiggybackCounters
```
thanodnl: Can we return a `func()` here that takes the changes from `d.issueChanges()` and feeds them into `d.bumpPiggybackCounters`. This makes it less error prone on the calling site after the response is received.

swim/disseminator.go
```diff
@@ -109,63 +109,88 @@ func (d *disseminator) FullSync() (changes []Change) {
...
 }
 
-func (d *disseminator) issueChanges(filter func(*pChange) bool) (changes []Change) {
+// filterChangesFromSender returns changes that didn't originate at the sender.
+// Attention, this function reorders the underlaying input array.
+func (d *disseminator) filterChangesFromSender(cs []Change, source string, incarnation int64) []Change {
+	for i := 0; i < len(cs); i++ {
+		if incarnation == cs[i].SourceIncarnation && source == cs[i].Source {
+			d.node.emit(ChangeFilteredEvent{cs[i]})
+			cs[i], cs[len(cs)-1] = cs[len(cs)-1], cs[i]
```
thanodnl: There is no reason to swap the `cs[i]` to the end of the list if your next step is to remove it. It is enough to `cs[i] = cs[len(cs)-1]`.

CorgiMan: Gotcha. We don't have need for it now, but I prefer to not overwrite changes in the underlying array, which you still have access to. I figured that reordering is not a concern since the order is non-deterministic to begin with. This saves us some memory allocation.

If you're unaware of how go does slices internally http://blog.golang.org/go-slices-usage-and-internals.

It explains that if you do:
b := a[:10]
where a has length 20. Slice a still exists and still has length 20. b just has access to the first 10 elements of a. Mutating elements of b will therefor mutate a. 

There are some subtleties in where you next do:
b = append(b, 999)
b now needs to grow since it's a its max capacity. The go runtime allocates a bigger array somewhere and copies the contents to it. If we now mutate elements of b, a is no longer mutated.

thanodnl:
lgtm

swim/disseminator.go
```diff
@@ -109,63 +109,95 @@ func (d *disseminator) FullSync() (changes []Change) {
...
+	for _, change := range changes {
+		c, ok := d.changes[change.Address]
+		if !ok {
+			continue
+		}
 
-	filter := func(c *pChange) bool {
-		return senderIncarnation == c.SourceIncarnation &&
-			senderAddress == c.Source
+		c.p += 1
```
dansimau: Not your fault and not a blocker for merging: `p` could be named better.

swim/disseminator.go
```diff
@@ -109,63 +109,95 @@ func (d *disseminator) FullSync() (changes []Change) {
...
+		if c.p >= d.maxP {
```
dansimau: Not your fault and not a blocker for merging: `maxP` is poorly documented.

swim/disseminator.go
CorgiMan: I agree, let's leave it for now

swim/disseminator.go
```diff
@@ -109,63 +109,95 @@ func (d *disseminator) FullSync() (changes []Change) {
...
+			delete(d.changes, c.Address)
+		}
 	}
+	d.Unlock()
+}
 
-	changes := d.issueChanges(filter)
+// IssueAsReceiver collects all changes a node needs when responding to a ping
+// or ping-req. Unlike IssueAsSender, IssueAsReceiver automatically increments
+// the piggyback counters. The second return value indicates whether a full sync
```
dansimau: Good work adding comments. This has helped me a lot.

I'd be even more liberal with this. E.g.:

> Unlike IssueAsSender, IssueAsReceiver automatically increments
> the piggyback counters.

Great highlight of an important detail. I'd take it one step further and explain _why_ it does this.

Anyway, this is not a blocker for merging.

swim/disseminator.go
```diff
@@ -109,63 +109,95 @@ func (d *disseminator) FullSync() (changes []Change) {
...
-		d.node.emit(FullSyncEvent{senderAddress, senderChecksum})
+	changes = d.issueChanges()
 
-		d.node.log.WithFields(log.Fields{
-			"localChecksum":  d.node.memberlist.Checksum(),
-			"remote":         senderAddress,
-			"remoteChecksum": senderChecksum,
-		}).Info("full sync")
+	// filter out changes that came from the sender previously
+	changes = d.filterChangesFromSender(changes, senderAddress, senderIncarnation)
```
dansimau: Nice

swim/disseminator_test.go
```diff
@@ -120,19 +176,132 @@ func (s *DisseminatorTestSuite) TestChangesDeleted() {
...
+	bumpPiggybackCounters()
 	s.Len(changes, 1, "expected one change to be issued")
 	s.Equal(1, s.d.changes[address].p, "expected propagations for change to be 1")
 
-	changes = s.d.IssueAsSender()
+	changes, bumpPiggybackCounters = s.d.IssueAsSender()
+	bumpPiggybackCounters()
 	s.Len(changes, 1, "expected one change to be issued")
-	s.Equal(2, s.d.changes[address].p, "expected propagations for change to be 2")
+	s.Empty(s.d.changes, "expected chandes are cleared after 2 propagations")
```
dansimau: s/chandes/changes

swim/disseminator_test.go
```diff
@@ -120,19 +176,132 @@ func (s *DisseminatorTestSuite) TestChangesDeleted() {
...
+	bumpPiggybackCounters()
 	s.Empty(changes, "expected no changes to be issued")
+	s.Empty(s.d.changes, "expected chandes are cleared after 2 propagations")
 
 	_, ok := s.d.changes[address]
 	s.False(ok, "expected change to be deleted")
+
+	// make sure bumping pb counters of removed changes is ok
+	oldBumpPBC()
+	s.Empty(s.d.changes, "expected chandes are cleared after 2 propagations")
```
dansimau: s/chandes/changes

CorgiMan: yes..

swim/disseminator.go
CorgiMan: I added a comment

swim/ping_handler.go
```diff
@@ -37,10 +37,11 @@ func handlePing(node *Node, req *ping) (*ping, error) {
 
 	node.memberlist.Update(req.Changes)
 
-	changes, fs := node.disseminator.IssueAsReceiver(req.Source, req.SourceIncarnation, req.Checksum)
+	changes, fullSync :=
```
dansimau: :+1:

swim/ping_request_sender.go
```diff
@@ -39,17 +39,6 @@ type pingRequest struct {
 	Changes           []Change `json:"changes"`
 }
 
-// newPingRequest creates a new ping request to the specified target.
```
dansimau: This patch is literally a reversal of a patch I made which introduced a constructor for `pingRequest`.

Personally, I find the lack of constructor here more difficult to understand, as without it I have to concern myself with every field of the `pingRequest`, even though only one or two fields should be set. `Source`, `SourceIncarnation`, `Checksum` should never be set manually, right? They are always going to refer to the node's address, incarnation number and checksum respectively.

swim/disseminator.go
```diff
@@ -40,6 +40,9 @@ type pChange struct {
 type disseminator struct {
 	node    *Node
 	changes map[string]*pChange
+
+	// maxP indicates how many times a change is disseminated. It is declared
```
dansimau: Nice!

swim/disseminator.go
dansimau: Oh awesome. Thanks, that helped!

dansimau:
LGTM :+1:

PR #83, opened 2016-01-27
Module logger

severb:
A simple module logger implementation.

### DIFF
--- a/logging/facility.go
+++ b/logging/facility.go
@@ -1,0 +1,169 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package logging
+
+import (
+	"fmt"
+	"github.com/uber-common/bark"
+	"sync"
+)
+
+// Facility is a collection of named loggers that can be configured
+// individually.
+type Facility struct {
+	logger bark.Logger
+	levels map[string]Level
+
+	mu sync.RWMutex
+}
+
+// NewFacility creates a new log facility with the specified logger as the
+// underlying logger. If no logger is passed, a no-op implementation is used.
+func NewFacility(log bark.Logger) *Facility {
+	if log == nil {
+		log = NoLogger
+	}
+	return &Facility{
+		logger: log,
+		levels: make(map[string]Level),
+	}
+}
+
+// SetLevels is like SetLevel but for multiple named loggers.
+func (f *Facility) SetLevels(levels map[string]Level) error {
+	for logName, level := range levels {
+		if level < Fatal {
+			return fmt.Errorf("cannot set a level above %s for %s", Fatal, logName)
+		}
+	}
+	// Prevent changing levels while a message is logged.
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	for logName, level := range levels {
+		f.levels[logName] = level
+	}
+	return nil
+}
+
+// SetLevel sets the minimum severity level for a named logger. All messages
+// produced by a named logger with a severity lower than the one set here are
+// silenced.
+// In most logger implementations, both Fatal and Panic stop the execution.
+// This means that those levels only produce a single, and often important,
+// message. Because of that, it's an error to pass a level above Fatal to this
+// method.
+func (f *Facility) SetLevel(logName string, level Level) error {
+	if level < Fatal {
+		return fmt.Errorf("cannot set a level above %s for %s", Fatal, logName)
+	}
+	// Prevent changing levels while a message is logged.
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.levels[logName] = level
+	return nil
+}
+
+// SetLogger sets the underlying logger. All log messages produced, that are
+// not silenced, are propagated to this logger.
+func (f *Facility) SetLogger(log bark.Logger) {
+	// Prevent changing the logger while a message is logged.
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.logger = log
+}
+
+// Logger returns a new named logger.
+func (f *Facility) Logger(logName string) bark.Logger {
+	return &namedLogger{
+		name:      logName,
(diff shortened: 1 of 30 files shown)


logging/facility.go
```diff
@@ -0,0 +1,169 @@
...
+// individually.
+type Facility struct {
+	logger bark.Logger
+	levels map[string]Level
+
+	mu sync.RWMutex
+}
+
+// NewFacility creates a new log facility with the specified logger as the
+// underlying logger. If no logger is passed, a no-op implementation is used.
```
dansimau: :+1: great

logging/facility.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+// SetLevel sets the minimum severity level for a named logger. All messages
+// produced by a named logger with a severity lower than the one set here are
+// silenced.
+// In most of logger implementations, both Fatal and Panic stop the execution.
+// This means that those levels only produce a single, and often important,
+// message. Because of that, it's an error to pass a level above Fatal to this
+// method.
```
dansimau: Excellent documentation :+1:

dansimau: `s/In most of logger implementations/In most logger implementations`

logging/facility.go
```diff
@@ -0,0 +1,169 @@
...
+// bark.Logger instance bound to a specific name.
+func (f *Facility) Log(logName string, wantLevel Level, fields bark.Fields, msg []interface{}) {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+	if setLevel, ok := f.levels[logName]; ok {
+		// For me this is confusing, so here's a clarification.
+		// The levels are consecutive integers starting from 0 defined in this order:
+		// Panic=0, Fatal=1, Error=2, Warning=3, etc...
+		// For the condition to make more sense, consider a named
+		// logger set to Fatal and an incoming Error message.
```
dansimau: Good commenting

logging/level.go
```diff
@@ -0,0 +1,90 @@
...
+	// Error log level
+	Error
+	// Warn log level
+	Warn
+	// Info log level
+	Info
+	// Debug log level
+	Debug
+
+	maxLevel = int(^uint8(0))
```
dansimau: Clever, but this could just be:

```
maxLevel = 255
```

If you changed the type of `Level` to something other than `uint8`, you'd still need to change this line too. So why not just simplify.

Also might be worth commenting on maxLevel and explaining why it exists.

logging/level.go
```diff
@@ -0,0 +1,90 @@
...
+		return Warn, nil
+	case "info":
+		return Info, nil
+	case "debug":
+		return Debug, nil
+	}
+
+	level, err := strconv.Atoi(lvl)
+	if level > maxLevel {
+		err = fmt.Errorf("Invalid level value: %s", lvl)
```
dansimau: nit: by convention errors are lowercase in Golang `¯\_(ツ)_/¯`

dansimau:
This PR is in good shape. I feel like we've come to a really good place. It's simple but also leaves the door open for future improvements.

Let's see some tests then rope someone else in for review and hopefully we can land this. :raised_hands:

logging/level.go
severb: Whoops, will change.

logging/level.go
severb: I agree. As a side note, `^int(0)` is much more useful since the size can vary.

logging/facility.go
severb: Thanks!

logging/facility.go
severb: :no_mouth:

severb:
Thanks for the help and feedback. Tests and updates for all log messages are on the way.

CorgiMan:
In assertions in tests can we add description strings on why we make the assertions?
s.Error(err, "expected .. to fail when ...")

severb:
@CorgiMan yes, good idea!

severb:
@CorgiMan and @dansimau I included your feedback.

The latest commit updated all log messages in the library to use named loggers. This touched a lot of files and broke some tests, which I fixed. Please have a look.

logging/level_test.go
```diff
@@ -0,0 +1,58 @@
...
+		s.Equal(gotLevel, level,
+			"Converting a Level to a string and back should produce the same Level.")
+	}
+}
+
+func (s *LogLevelTestSuite) TestInvalidLevel() {
+	levels := []string{"", "1234", "abc"}
+	for _, level := range levels {
+		_, err := Parse(level)
+		s.Error(err,
```
dansimau: nit: I think this is unusual. In Go you just put it the whole thing on one line. The temptation to put this on a new line is a hangover from Python. I don't think it makes the code more readable.

logging/named_test.go
```diff
@@ -0,0 +1,120 @@
...
+		expected := &recv{
+			LogName:   "name",
+			WantLevel: c.level,
+			Fields:    bark.Fields{"key": 1},
+			Format:    "",
+			Msg:       []interface{}{"msg"},
+		}
+
+		s.Equal(s.recv, expected,
+			"Log messages should be forwarded unmodified.")
```
dansimau: Also here

logging/named_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+		*s.recv = recv{} // reset
+
+		// logger.Debugf("format", "msg")
+		c.logMethf(s.logger, "format", "msg")
+
+		expected.Format = "format"
+		s.Equal(s.recv, expected,
+			"Log messages should be forwarded unmodified.")
```
dansimau: And here

logging/named_test.go
```diff
@@ -0,0 +1,120 @@
...
+	l := s.logger.WithField("a", 1).WithFields(bark.Fields{"b": 2, "a": 3})
+	l.Debug("msg")
+	expected := &recv{
+		LogName:   "name",
+		WantLevel: Debug,
+		Fields:    bark.Fields{"key": 1, "a": 3, "b": 2},
+		Msg:       []interface{}{"msg"},
+	}
+	s.Equal(s.recv, expected,
+		"Multiple fields should be merged.")
```
dansimau: etc

dansimau:
LGTM. Made some comments on code formatting but aside from that everything looks magical.

I'd say we should deploy this on ringpop-go (our test daemon) and implement @thanodnl's log messages for the failure testing. It will help us flesh out any bugs.

Also having a quick look at the test coverage:
```
$ go test -coverprofile=/tmp/cov.out
PASS
coverage: 93.3% of statements
ok  	github.com/uber/ringpop-go/logging	0.016s
```

I think we could do tests for the default facility and noop logger and get this to 100%.

If we're gonna do that let's try and do this quickly though. We don't want this to drag on.

severb:
I inlined the assertion messages and improved the coverage slightly. I didn't test the noop logger because there's little value in doing so and there's not much I can assert :smile:

Thanks for the review!

logging/nologger.go
```diff
@@ -0,0 +1,47 @@
...
+func (l noLogger) Infof(format string, args ...interface{})            {}
+func (l noLogger) Warn(args ...interface{})                            {}
+func (l noLogger) Warnf(format string, args ...interface{})            {}
+func (l noLogger) Error(args ...interface{})                           {}
+func (l noLogger) Errorf(format string, args ...interface{})           {}
+func (l noLogger) Fatal(args ...interface{})                           {}
+func (l noLogger) Fatalf(format string, args ...interface{})           {}
+func (l noLogger) Panic(args ...interface{})                           {}
+func (l noLogger) Panicf(format string, args ...interface{})           {}
+func (l noLogger) WithField(key string, value interface{}) bark.Logger { return l }
```
dansimau: Does this create a new copy of the logger? Should we return a pointer to the same logger instead?

severb: It does create a copy on each method call but since it's an empty struct *I expect* it to be at least as efficient as passing a pointer.

dansimau: What about garbage collection?

dansimau:
>  I didn't test the noop logger because there's little value in doing so

I think it would be a reasonable test to assert that by default log messages _aren't_ forwarded. That would also cover the noLogger.

logging/nologger.go
severb: The size of an empty struct instance is zero, I don't think there is any GC involved.

dansimau: OK, cool. :+1:

severb:
They are, it's just that the noop logger drops them.

jwolski:
lgtm

PR #84, opened 2016-01-27
Add test for error on bootstrap

dansimau:
Also remove testPop as it's not used.

Cause... why not `¯\_(ツ)_/¯`

@uber/ringpop

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -187,15 +187,16 @@ func (rp *Ringpop) identity() (string, error) {
 // r.channelIdentityResolver resolves the hostport identity from the current
 // TChannel object on the Ringpop instance.
 func (rp *Ringpop) channelIdentityResolver() (string, error) {
-	hostport := rp.channel.PeerInfo().HostPort
-	// Check that TChannel is listening. By default, TChannel listens on an
-	// ephemeral host/port. The real port is then assigned by the OS when
-	// ListenAndServe is called. If the hostport is 0.0.0.0:0, it means
-	// TChannel is not yet listening and the hostport cannot be resolved.
-	if hostport == "0.0.0.0:0" {
-		return "", fmt.Errorf("unable to resolve valid listen address (TChannel hostport is %s)", hostport)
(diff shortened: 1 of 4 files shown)


ringpop_test.go
```diff
@@ -549,6 +549,18 @@ func (s *RingpopTestSuite) TestEmptyJoinListCreatesSingleNodeCluster() {
...
 
+func (s *RingpopTestSuite) TestErrorOnChannelNotListening() {
+	ch, err := tchannel.NewChannel("test", nil)
+	s.Require().NoError(err)
+
+	rp, err := New("test", Channel(ch))
+	s.Require().NoError(err)
+
+	nodesJoined, err := rp.Bootstrap(&swim.BootstrapOptions{})
+	s.Error(err)
```
jwolski: Is it possible to add a stricter assert? What kind of error should we expect back?

thanodnl:
lgtm

ringpop.go
```diff
@@ -188,15 +188,15 @@ func (rp *Ringpop) identity() (string, error) {
 // r.channelIdentityResolver resolves the hostport identity from the current
 // TChannel object on the Ringpop instance.
 func (rp *Ringpop) channelIdentityResolver() (string, error) {
-	hostport := rp.channel.PeerInfo().HostPort
+	peerInfo := rp.channel.PeerInfo()
 	// Check that TChannel is listening. By default, TChannel listens on an
 	// ephemeral host/port. The real port is then assigned by the OS when
 	// ListenAndServe is called. If the hostport is 0.0.0.0:0, it means
```
Motiejus: Comment-nitpick: looking at `connection.go:53`: it should say "ends with `:0`" instead (this can happen if service is listening, e.g., on loopback address, and specifying the zero port). Code is fine though.

dansimau:
@Motiejus improved the comment, lemme know what you think :blush:

Motiejus:
LGTM

PR #85, opened 2016-01-28
don't disseminate join list

CorgiMan:
After join, clear all nonlocal changes from the disseminator

### DIFF
--- a/swim/disseminator.go
+++ b/swim/disseminator.go
@@ -179,3 +179,24 @@ func (d *disseminator) RecordChange(change Change) {
 	d.changes[change.Address] = &pChange{change, 0}
 	d.Unlock()
 }
+
+func (d *disseminator) ClearChange(c Change) {
+	d.Lock()
+	delete(d.changes, c.Address)
+	d.Unlock()
+}
+
+func (d *disseminator) ChangeByAddress(address string) (Change, bool) {
+	d.Lock()
+	pc, ok := d.changes[address]
+	c := pc.Change
+	d.Unlock()
+	return c, ok
+}
+
+func (d *disseminator) ChangeCount() int {
+	d.Lock()
+	c := len(d.changes)
+	d.Unlock()
+	return c
+}
--- a/swim/disseminator.go
+++ b/swim/disseminator.go
@@ -186,20 +186,20 @@ func (d *disseminator) ClearChange(c Change) {
 	d.Unlock()
 }
 
-func (d *disseminator) ChangeByAddress(address string) (Change, bool) {
-	d.Lock()
-	pc, ok := d.changes[address]
+func (d *disseminator) ChangesByAddress(address string) (Change, bool) {
+	d.RLock()
+	change, ok := d.changes[address]
 	var c Change
 	if ok {
-		c = pc.Change
+		c = change.Change
 	}
-	d.Unlock()
+	d.RUnlock()
 	return c, ok
 }
 
-func (d *disseminator) ChangeCount() int {
-	d.Lock()
+func (d *disseminator) ChangesCount() int {
+	d.RLock()
 	c := len(d.changes)
-	d.Unlock()
+	d.RUnlock()
 	return c
 }
--- a/swim/disseminator.go
+++ b/swim/disseminator.go
@@ -189,7 +189,10 @@ func (d *disseminator) ClearChange(c Change) {
 func (d *disseminator) ChangeByAddress(address string) (Change, bool) {
 	d.Lock()
 	pc, ok := d.changes[address]
-	c := pc.Change
+	var c Change
+	if ok {
+		c = pc.Change
+	}
 	d.Unlock()
 	return c, ok
 }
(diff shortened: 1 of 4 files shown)


swim/join_sender.go
```diff
@@ -388,7 +388,11 @@ func (j *joinSender) JoinGroup(nodesJoined []string) ([]string, []string) {
 					failed = true
 					break
 				}
+
 				j.node.memberlist.Update(res.Membership)
+				for _, change := range res.Membership {
```
jwolski: This is a clever way of solving this problem. I was thinking that it could/should be solved by making the initial membership-set more explicit in the API, but for the sake of time, I'm OK with this. I will ask you to place a comment about the ClearChange loop saying why we're doing this.

swim/join_test.go
```diff
@@ -212,3 +212,15 @@ func (s *JoinSenderTestSuite) TestCustomDelayer() {
...
 	suite.Run(t, new(JoinSenderTestSuite))
 }
+
+func (s *JoinSenderTestSuite) TestDontDisseminateJoinList() {
+	tnodes := genChannelNodes(s.T(), 5)
+	bootstrapNodes(s.T(), tnodes...)
+
+	for i, tnode := range tnodes {
+		s.Len(tnodes[i].node.memberlist.GetMembers(), i+1, "expected that next node joined all previous ones")
+		s.Len(tnodes[i].node.disseminator.changes, 1, "expected to only disseminate yourself")
```
jwolski: can you make this Disseminator API calls? e.g. (`ChangeByAddress()` and `ChangeCount()`)

swim/join_sender.go
CorgiMan: Yeah I've wondered about the same thing. I figured this is the most obvious way to do it since it's reusing the elements that are already built. Will add comment though

swim/join_test.go
CorgiMan: Yes

swim/disseminator.go
```diff
@@ -179,3 +179,24 @@ func (d *disseminator) RecordChange(change Change) {
...
+
+func (d *disseminator) ClearChange(c Change) {
+	d.Lock()
+	delete(d.changes, c.Address)
+	d.Unlock()
+}
+
+func (d *disseminator) ChangeByAddress(address string) (Change, bool) {
+	d.Lock()
+	pc, ok := d.changes[address]
```
jwolski: better name for `pc` plz

swim/join_sender.go
```diff
@@ -388,7 +388,17 @@ func (j *joinSender) JoinGroup(nodesJoined []string) ([]string, []string) {
 					failed = true
 					break
 				}
+
+				// We add the join list to the membership with the Update
+				// function. However, as a side effect, Update adds changes to
+				// the disseminator as well. Since we don't want to disseminate
+				// the potentially very large join lists, we clear all the
+				// changes from the disseminator, except for the one change
```
jwolski: Good comment :+1:

swim/disseminator.go
CorgiMan: The struct is called pChange, which is a terrible name, but since it is, I think this name is fine. Do you have alternatives?

jwolski:
```
change, ok := d.changes[address]
```

swim/disseminator.go
```diff
@@ -179,3 +179,24 @@ func (d *disseminator) RecordChange(change Change) {
...
 }
+
+func (d *disseminator) ClearChange(c Change) {
+	d.Lock()
+	delete(d.changes, c.Address)
+	d.Unlock()
+}
+
+func (d *disseminator) ChangeByAddress(address string) (Change, bool) {
+	d.Lock()
```
thanodnl: From the code I would expect that an `RLock` would be enough here.

swim/disseminator.go
jwolski: Shouldn't you tests for ok here?

```
if piggybackedChange, ok := d.changes[address]; ok {
  d.Unlock()
  change := piggybackedChange.Change
  return change, true
}

return nil, false
```
```

File: swim/disseminator.go [hunk 6]
```diff
@@ -179,3 +179,24 @@ func (d *disseminator) RecordChange(change Change) {
...
+	pc, ok := d.changes[address]
+	c := pc.Change
```
thanodnl: What happens when the read is not `ok`? I would expect a `nil pointer dereference` here. Lets wrap this with a `if ok { ... }`

swim/disseminator.go
```diff
@@ -179,3 +179,24 @@ func (d *disseminator) RecordChange(change Change) {
...
+	d.Unlock()
+	return c, ok
+}
+
+func (d *disseminator) ChangeCount() int {
+	d.Lock()
```
thanodnl: `RLock`?

swim/disseminator.go
CorgiMan: Yup, fixed that in next commit

jwolski:
lgtm

thanodnl:
lgtm

PR #86, opened 2016-02-01
Add locks in disseminator.HasChanges method

CorgiMan:


jwolski:
lgtm

swim/disseminator.go
```diff
@@ -88,7 +88,10 @@ func (d *disseminator) AdjustMaxPropagations() {
 
 // HasChanges reports whether disseminator has changes to disseminate.
 func (d *disseminator) HasChanges() bool {
-	return len(d.changes) > 0
+	d.Lock()
+	r := len(d.changes) > 0
```
jwolski: `r` is not a very good name. Please rename.

CorgiMan: I believe it's quite idiomatic to have 1 letter variable names, especially in small scope functions like this.

PR #87, opened 2016-02-01
Add readlock and use better variable name as per Jeff's feedback

CorgiMan:


jwolski:
lgtm, but broken tests? go ahead and merge if it's a flapper.

PR #90, opened 2016-02-02
Fix invalid return code on partial compile error

dansimau:
This better be the last time I do this.

@uber/ringpop

### DIFF
--- a/test/go-test-prettify
+++ b/test/go-test-prettify
@@ -1,74 +1,104 @@
 #!/bin/bash
 #
-# Parse the output from `go test -v` and prints a summary at the end. Also only
-# outputs errors by default unless -v is specified.
+# Runs `go test -v`, parses the output and prints a summary at the end. Also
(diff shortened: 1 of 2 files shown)


test/go-test-prettify
```diff
@@ -1,74 +1,104 @@
...
+					echo "$line"
+				fi
+				;;
+
+			# Skip and ignore printing junk
+			PASS|FAIL|\?*)
+				;;
+
+			*)
+				# Output unknown lines
```
Motiejus: Maybe you meant the opposite, i.e. output unknown lines _only_ in verbose mode?

Motiejus: Sorry, I missed that you're outputting them below.

Motiejus:
LGTM

thanodnl:
lgtm

test/go-test-prettify
dansimau: :+1:

PR #91, opened 2016-02-02
Wait longer for flushing the buffer in unit tests

Motiejus:
This test flaps on travis-ci.

This is easy to reproduce by giving CPU some work next to executing this test,
e.g. a few cpuburn processes on the same core.

### DIFF
--- a/swim/update_rollup_test.go
+++ b/swim/update_rollup_test.go
@@ -23,6 +23,7 @@ package swim
 import (
 	"testing"
 	"time"
+	"fmt"
 
 	"github.com/stretchr/testify/suite"
 	"github.com/uber/ringpop-go/util"
@@ -99,11 +100,23 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
 	})
 
 	s.NotNil(s.r.FlushTimer(), "expected timer to be set")
+
+	// This is a race condition which never got triggered yet. If the scheduler
+	// decides to yield at this comment and proceed with flushing the buffer,
+	// assertion below will fail. Oops.
 	s.Len(s.r.Buffer(), 2, "expected two updates to be in buffer")
 
-	time.Sleep(2 * time.Millisecond)
+	// Flushing the buffer used to take <2ms on an otherwise idle machine. On
+	// travis-ci we might have a significantly less CPU shares, so wait longer.
+	sleepUpTo := 100 // Milliseconds
+	for i := 0; i < sleepUpTo; i++ {
+		if len(s.r.Buffer()) == 0 {
+			break
+		}
+		time.Sleep(time.Millisecond)
+	}
 
-	s.Empty(s.r.Buffer(), "expected buffer to be flushed")
+	s.Empty(s.r.Buffer(), fmt.Sprintf("expected buffer to be flushed after %d ms", sleepUpTo))
 	s.NotNil(s.r.FlushTimer(), "expected timer to be renewed")
 }
 


swim/update_rollup_test.go
```diff
@@ -99,11 +100,25 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
...
+	sleepUpTo := 100 // Milliseconds
+	for i := 0; i < sleepUpTo; i++ {
+		if len(s.r.Buffer()) == 0 {
+			break
+		}
+		time.Sleep(time.Millisecond)
+	}
 
-	s.Empty(s.r.Buffer(), "expected buffer to be flushed")
+	s.Empty(
```
dansimau: Just a note that I've found multiline statements like this to be less common in Golang, unless the line is really long.

The only thing that's said on it is:
https://github.com/golang/go/wiki/CodeReviewComments#line-length

Motiejus:
@dansimau I agree with you; updated the commit to have a single line there.

swim/update_rollup_test.go
```diff
@@ -99,11 +100,23 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
 	})
 
 	s.NotNil(s.r.FlushTimer(), "expected timer to be set")
+
+	// This is a race condition which never got triggered yet. If the scheduler
```
dansimau: Can you induce this race condition? Is it practically possible to induce this? If not, I'd say there is no race condition and perhaps the comment is slightly confusing. If there is a race condition, we should file a task and fix it.

Also for clarity, it might help to prefix the comment with "TODO:" as my first impression was that the commented line was actually fixing it or something.

swim/update_rollup_test.go
```diff
@@ -99,11 +100,23 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
...
+	// decides to yield at this comment and proceed with flushing the buffer,
+	// assertion below will fail. Oops.
 	s.Len(s.r.Buffer(), 2, "expected two updates to be in buffer")
 
-	time.Sleep(2 * time.Millisecond)
+	// Flushing the buffer used to take <2ms on an otherwise idle machine. On
```
dansimau: I'm unsure if the test is actually meant to assert that the buffer gets flushed after the `flushInterval`. We should check what contract `updateRollup` is providing for flushing and if this matters.

If the contract is strict (i.e. flushing *MUST* occur after the `flushInterval`), then the test is genuinely failing. I think it would be a pretty bad contract to provide though and impossible to guarantee.

Let's have a quick chat with @jwolski and try to find out how precise the flushes need to be for an `updateRollup`.

swim/update_rollup_test.go
```diff
@@ -99,11 +100,23 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
...
+	// travis-ci we might have a significantly less CPU shares, so wait longer.
+	sleepUpTo := 100 // Milliseconds
```
dansimau: This is a little misleading. 100 is the number if iterations of the loop, not the time. The call to `Buffer()` in the loop has a read lock, so I'd say it's entirely possible for the actual time spent to go past 100 milliseconds.

swim/update_rollup_test.go
```diff
@@ -99,11 +100,23 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
...
+	for i := 0; i < sleepUpTo; i++ {
+		if len(s.r.Buffer()) == 0 {
+			break
+		}
+		time.Sleep(time.Millisecond)
```
jwolski: See the use of a mockClock in swim/handler_test.go. Let's not depend on the passing of wall-clock time with time.Sleeps.

swim/update_rollup_test.go
jwolski: The purpose of update_rollup.go is to collect membership updates and report them to the logging system after a period of at least X (`flushInterval`) milliseconds of "quite time" (aka the cluster convering).

I think this test is making sure that after 1ms passes, the "buffer is flushed."

The contract is not that strict. I'm not sure what type of promises Go's timers make. I'd imagine it something like "sometime after, but not earlier than the time you specify." That is OK enough for update_rollup.go too.

swim/update_rollup_test.go
jwolski: +1 for finding a potential race
+1 for Dans remarks; put a TODO and let's make another post-it for addressing the race. I think this test is addressing 2 different concerns and that it needs to be split up into 2 tests: 1) Flushing empties the buffer, 2) Enabling the background flusher flushes buffer after flushInterval. To prove #2 one can either separate the concerns of what it means to flush the buffer and what it means to trigger flushing after an interval OR, and more simply, just raise an event (invoke a delegate) after the flushing occurs and assert the event was emitted (or delegate called) but not asserting that the buffer is actually empty.

Motiejus:
Obsoleted by #93

PR #92, opened 2016-02-05
add ring.checksum* stats for partition detection

benfleis:
This adds 2 stats useful in external partition detection/monitoring.

@uber/ringpop


options.go
```diff
@@ -177,6 +183,25 @@ func IdentityResolverFunc(resolver IdentityResolver) Option {
 	}
 }
 
+// Installs RingChecksumStatPeriod, which determines the period between
+// emissions of the stat 'ring.checksum-periodic'. Using a value <=0 will
+// disable emissino of this stat. Using a value in (0, 10ms) will return an
```
dansimau: s/emissino/emission

options.go
```diff
@@ -177,6 +183,25 @@ func IdentityResolverFunc(resolver IdentityResolver) Option {
 	}
 }
 
+// Installs RingChecksumStatPeriod, which determines the period between
```
dansimau: documentation nitpick: "Installs" is a weird word here in the context of the godoc (e.g. for a user). It might make more sense just to say _"RingChecksumStatPeriod determines the period between ..."_

dansimau: Also, that would fix this minor golint note:

```
$ golint options.go 
options.go:186:1: comment on exported function RingChecksumStatPeriod should be of the form "RingChecksumStatPeriod ..."
```

Which is a result of this convention:

_"The first sentence should be a one-sentence summary that **starts with the name being declared**."_
https://golang.org/doc/effective_go.html#commentary

Yay, code formatting conventions! :dancers:

dansimau:
Looks good, but how do we test this?

options.go
benfleis: ack

options.go
benfleis: ack, and thanks for the info. @dansim ought we include linting in our pre-commit hooks?

benfleis:
configuration tests added in coming commit. there will be a separate commit w/ actual Ticker handling.

options.go
dansimau: Yeah, I want to. And add it to Travis too. First step is to add a `make lint`...

ringpop.go
```diff
@@ -173,11 +174,44 @@ func (rp *Ringpop) init() error {
...
 	return nil
 }
 
+// Starts periodic timers in a single goroutine. Can be turned back off via
+// stopTimers. At present, only 1 timer exists, to emit ring.checksum-periodic.
+func (rp *Ringpop) startTimers() {
+	if rp.tickers != nil {
+		return
+	}
+	rp.tickers = make(chan *time.Ticker, 1) // 1 == max number of ticker
```
thanodnl: Why using a channel here instead of a slice?

benfleis: [discussed offline] - is it more gopherish to use channel here? I found at least one go blogger claiming as much. Either is possible.

benfleis:
ok, v2 with testing!
@dansimau @CorgiMan @Motiejus

options.go
```diff
@@ -77,6 +84,19 @@ func checkOptions(rp *Ringpop) []error {
 
 // Runtime options
 
+// Clock is used to set the Clock mechanism. Passing nil will use the system
+// clock. Testing harnesses will typically replace this with a mocked clock.
+func Clock(c clock.Clock) Option {
+	return func(r *Ringpop) error {
+		if c == nil {
```
dansimau: Why would we want/allow them to pass `nil`? Shouldn't passing nil be an error?

I think the system clock be the default option and then this functional option becomes optional.

options.go
```diff
@@ -177,8 +197,38 @@ func IdentityResolverFunc(resolver IdentityResolver) Option {
...
+		r.config.RingChecksumStatPeriod = period
+		return nil
+	}
+}
+
 // Default options
 
+// RingChecksumStatPeriodNever defines a "period" which disables
+// ring.checksum-periodic stat emission.
+const RingChecksumStatPeriodNever = time.Duration(-1)
```
dansimau: Does this constant belong here? I feel like it relates to stats, not options. Should it be in `ringpop.go` or in a new `stats.go`?

options.go
```diff
@@ -177,8 +197,38 @@ func IdentityResolverFunc(resolver IdentityResolver) Option {
...
+
+// RingChecksumStatPeriodDefault defines the default emission period for the
+// ring.checksum-periodic stat.
+const RingChecksumStatPeriodDefault = time.Duration(5 * time.Second)
```
dansimau: Is this constant really required? It is only used in one place in the code base, a few lines later. Is it going to be used somewhere else in the future?

dansimau: Ah, I should pay more attention. I see now it's in the tests as well :blush:

ringpop.go
```diff
@@ -160,7 +164,9 @@ func (rp *Ringpop) init() error {
 	rp.subChannel = rp.channel.GetSubChannel("ringpop", tchannel.Isolated)
 	rp.registerHandlers()
 
-	rp.node = swim.NewNode(rp.config.App, address, rp.subChannel, nil)
+	rp.node = swim.NewNode(rp.config.App, address, rp.subChannel, &swim.Options{
+		Clock: clock.New(),
```
dansimau: Are we not using the clock passed to Ringpop? Am I reading this wrong?

options.go
CorgiMan: Does this mean that we now have two ways to set a default value. I think calling it with nil should cause a panic or return an error.

CorgiMan:
is it okay to use a repo that is failing to build (according to readme) and has experimental status?

options.go
CorgiMan: I see that Dan was here first..

options.go
CorgiMan: In stead of having a negative value, is it better to have a seperate Option for this?
func NoRingChecksumStat() Opstion

options.go
benfleis: The problem is slightly tricky -- both a swim.Node, and ringpop.Ringpop want to be able to have sensible defaults. Options is defined at a level outside swim. Unless we want swim to refer to ringpop.options for its own defaults, using nil allows each of them to do the sane thing here. Perhaps there's a more elegant way of handling this in go, but modularity makes this slightly trickier, afaict.

Modularity aside, I think nil is a valid default for "use the system clock", such that a user then has no need to know how to instantiate one of them, similar to a sane global default.

@CorgiMan @dansimau

options.go
benfleis: @dansimau There is no current notion of 'stats' across the board. It might be sensible to create one, but I went more minimal.

@CorgiMan Also possible, but then it's no longer a pass through from a config file. Do we have a pattern/example for this type of thing? I would rather follow the existing pattern then invent my own...

options.go
benfleis: It could still go elsewhere. Again, we don't seem to have a clear path on where options should live. I welcome guidance here; if making a stats.go is the right thing, then so be it.

ringpop.go
benfleis: good catch. fixing.

benfleis:
@CorgiMan In short, I am okay with it -- it's little different than us writing our own. If we're concerned, the better thing is to fix+submit.

Motiejus:
I see the travis build is failing on a test I`m working on; feel free to cherry-pick this commit to fix it: https://github.com/uber/ringpop-go/commit/60c5437ce8f75d05e199e7874ab9f3d0ab15277c

The commit is based on top of your branch as of now, so should be ff-merge.

options.go
dansimau: They are using different options mechanisms, though.

The functional option gives us precisely this advantage. We can differentiate between `nil` and not having anything passed. There is no reason why we would ever want to accept a nil clock, so I think doing so explicitly should be an error.

If you want the default clock, you don't need to specify this option at all.

For the swim node, it's not as relevant, since the package is internal (well, it's meant to be and/or will be as soon as we get around to it) and I think that if we're taking in a Clock from the user from Ringpop we need to pass that on in swim as well. So we're always going to be passing the Ringpop clock.

options.go
dansimau: I think it's okay to make `0` mean "disabled", as long as it's clear in the doc comment. I can see the reason for the constant, but I think it actually makes it more complicated than it needs to be.

This is what disabling the checksum would look like with the constant:

```go
rp := ringpop.New("myapp",
    ringpop.Channel(ch),
    ringpop.RingChecksumStatPeriod(ringpop.RingChecksumStatPeriodNever))
```

Without it:

```go
rp := ringpop.New("myapp",
    ringpop.Channel(ch),
    ringpop.RingChecksumStatPeriod(0))   // If you really really need to do this, you can
```

As a side note, I don't see a strong use-case for ever disabling this anyway.

options.go
dansimau: Meh, I don't know... I think defaults pertaining to options are fine in options. But constants in general sound like they are default across the code base, so not really relating to options. Maybe it should go in `ringpop.go`? But then again, it's crowded in there. I'll leave it up to you to decide...

benfleis: we have no current "defaults.go" or "config.go" which specifies these sorts of things -- they are scattered about. Both are reasonable IMO, and since there is no clear precedent, I will leave it where it is. If we start having more of these, then it is trivial to move it to a better location.

options.go
benfleis: 0 is a magic number, even if a common one. You can use any value <= 0, and it will work just fine. Our code ought to use symbolic constants instead of magic numbers wherever possible; clients/users can do it as desired.

I corrected the original use of 0 in the code, replacing it with the symbol. The rest I will leave as-is.

benfleis:
@CorgiMan btw, tests pass locally. It appears to be an issue w/ the drone.io setup.

options.go
```diff
@@ -77,6 +84,19 @@ func checkOptions(rp *Ringpop) []error {
 
 // Runtime options
 
+// Clock is used to set the Clock mechanism.  Testing harnesses will typically
+// replace this with a mocked clock.
+func Clock(c clock.Clock) Option {
+	return func(r *Ringpop) error {
+		if c == nil {
+			return errors.New("Clock is required")
```
dansimau: By convention, error messages in Golang do not start with a capital letter (unless you're referring to a name I guess). Just an FYI, not a blocker, if that's what you intended.

dansimau:
LGTM :+1:

options.go
dansimau: > Our code ought to use symbolic constants instead of magic numbers wherever possible; clients/users can do it as desired.

That makes sense. But it probably makes sense to define `RingChecksumStatPeriodNever` as the same thing we're telling users to use: `0`.

Otherwise we're testing with `time.Duration(-1)` but telling our users to use `0`.

I see the tests are testing the `<= 0` case though, which is a good thing. Nice one.

options.go
benfleis: fixed, along w/ neighbors

swim/node.go
```diff
@@ -66,7 +67,7 @@ func defaultOptions() *Options {
 		RollupFlushInterval: 5000 * time.Millisecond,
 		RollupMaxUpdates:    250,
 
-		Clock: &systemClock{},
+		Clock: nil,
```
CorgiMan: isn't this supposed to be clock.New()

swim/node.go
```diff
@@ -99,7 +102,7 @@ func mergeDefaultOptions(opts *Options) *Options {
 		def.PingRequestSize)
 
 	if opts.Clock == nil {
-		opts.Clock = def.Clock
+		opts.Clock = clock.New()
```
CorgiMan: And here:
if opts.Clock != nil { ... }
Am I misunderstanding something?

swim/node.go
```diff
@@ -76,6 +77,8 @@ func mergeDefaultOptions(opts *Options) *Options {
 	def := defaultOptions()
 
 	if opts == nil {
+		// typically only in tests
+		def.Clock = clock.New()
```
CorgiMan: I don't get why we do this. Why do we set it if the user supplies no options?

swim/update_rollup_test.go
```diff
@@ -101,7 +103,7 @@ func (s *RollupTestSuite) TestFlushesAfterInterval() {
 	s.NotNil(s.r.FlushTimer(), "expected timer to be set")
 	s.Len(s.r.Buffer(), 2, "expected two updates to be in buffer")
 
-	time.Sleep(2 * time.Millisecond)
+	s.node.clock.(*clock.Mock).Add(2 * time.Millisecond)
```
CorgiMan: I think its a bit cleaner if we didn't need the cast
s.clock.Add(2*..)
This means need to add the field to the testSuite. 
Not a blocker though

dansimau: It's a good point. A related question is: how come the mock clock implementation doesn't just replace `Sleep` with the `Add`? That would make sense to me, because then the mock clock would work transparently, wouldn't it?

CorgiMan: I think that rename only makes sense if sleep is part of the interface, which could be a good idea.

swim/node.go
CorgiMan: I think having to init an object here is the cost of using a struct for options. If it is the default option I think we should set it here, not somewhere else. If this is not efficient enough, I think we should consider reworking the design.

swim/update_rollup_test.go
benfleis: Agreed on both counts. IMO, Sleep should be an alias for Add in this case. But I have no intention of updating the lib right now for this case.

dansimau:
LGTM :+1:

CorgiMan:
LgTm

PR #93, opened 2016-02-05
Split TestFlushesAfterInterval, remove Timer.sleep

Motiejus:
TestFlushesAfterInterval flaps on travis-ci due to relying on wall clock.

1. Create TestFlushingEmptiesTheBuffer, which tests FlushBuffer.
2. Remove Timer.sleep from TestFlushesAfterInterval by mocking the timer.

### DIFF
--- a/swim/clock.go
+++ b/swim/clock.go
@@ -27,6 +27,7 @@ import (
 // Clock is the interface that wraps the Now method.
 type Clock interface {
 	Now() time.Time
+	AfterFunc(time.Duration, func()) *time.Timer
 }
 
 // systemClock uses the system time to implement the Clock interface.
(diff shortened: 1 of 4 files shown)


swim/clock.go
```diff
@@ -27,6 +27,7 @@ import (
 // Clock is the interface that wraps the Now method.
```
dansimau: Comment is out-of-date

swim/clock.go
```diff
@@ -37,6 +38,10 @@ func (c systemClock) Now() time.Time {
 	return time.Now()
 }
 
+func (c systemClock) AfterFunc(d time.Duration, f func()) *time.Timer {
```
dansimau: Should have a godoc comment. Perhaps you can redirect people to `time.AfterFunc` though, to avoid duplicating the description.

swim/clock.go
Motiejus: Good catch, updated in R2.

swim/clock.go
Motiejus: OK.

dansimau:
RE: discussions earlier today, should we replace these mocks _now_ with a time mocking library? I feel like it's better to try dropping this in now: https://github.com/benbjohnson/clock

dansimau:
Fixed in #92.

PR #94, opened 2016-02-05
Make leave respect incarnation numbers

CorgiMan:
1. make the update rules more transparent.
2. fixes bug: If a member is in the leave state, only updates where the member
   is made alive are applied. This means that updates where the member changes
   state to the Suspect, Faulty or Leave state are ignored, even if the
   the updates should be applied due to having a bigger incarnation number.
   So if node.state==Leave, the following changes are now applied: when change.state == Suspect | Faulty | Leave and change.IncNo>node.IncNo

Please ignore changes to the generated mocks dir. Only member.go and member_test.go are actual changes.

### DIFF
--- a/swim/member.go
+++ b/swim/member.go
@@ -31,14 +31,14 @@ const (
 	// Alive is the member "alive" state
 	Alive = "alive"
 
+	// Suspect is the member "suspect" state
+	Suspect = "suspect"
+
 	// Faulty is the member "faulty" state
 	Faulty = "faulty"
 
 	// Leave is the member "leave" state
 	Leave = "leave"
-
-	// Suspect is the member "suspect" state
-	Suspect = "suspect"
 )
 
 // A Member is a member in the member list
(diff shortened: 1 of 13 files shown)


swim/member.go
```diff
@@ -70,46 +70,46 @@ func shuffle(members []*Member) []*Member {
...
-
-func (m *Member) localSuspectOverride(local string, change Change) bool {
-	return m.Address == local && change.Status == Suspect
+	if m.Address != local {
+		return false
+	}
+	return change.Status == Faulty || change.Status == Suspect
+}
+
+func statePrecedence(s string) int {
```
Motiejus: Shall we use it like this:
```
const (
  Alive0 = 0
  Suspect1 = 1
  Faulty = 2
  Leave = 3
)
```
and use them in this function?

thanodnl: In the end I think it will be better to use this:

```go
type State int

const (
    Alive State = iota
    Suspect
    Faulty
    Leave
)
```

But lets keep it like this for now.

swim/member_test.go
```diff
@@ -0,0 +1,121 @@
...
+	// the state (incNum, status pairs) slice is generated with ever
+	// increasing precendence, changes with index j override members
+	// with index i if and only if j > i.
+	for i, s1 := range s.states {
+		for j, s2 := range s.states {
+			m := newMember(s.localAddr, s1)
+			c := newChange(s.localAddr, s2)
+			expected := j > i
+			got := m.nonLocalOverride(c)
+			s.Equal(expected, got, "expected override iff j > i")
```
thanodnl: what does `iff` stand for here

CorgiMan: If and only if. It might be only known by math nerds...

thanodnl:
lgtm

swim/member.go
dansimau: +1 what @thanodnl suggested

dansimau:
LGTM

PR #95, opened 2016-02-05
don't remove local member from the changes when adding join list to m…

CorgiMan:
…embership

### DIFF
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -283,6 +283,22 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 	return applied
 }
 
+// AddJoinList adds the list to the membership with the Update
+// function. However, as a side effect, Update adds changes to
+// the disseminator as well. Since we don't want to disseminate
+// the potentially very large join lists, we clear all the
+// changes from the disseminator, except for the one change
+// that refers to the make-alive of this node.
+func (m *memberlist) AddJoinList(list []Change) {
+	applied := m.Update(list)
+	for _, member := range applied {
+		if member.Address == m.node.Address() {
+			continue
+		}
+		m.node.disseminator.ClearChange(member.Address)
+	}
+}
+
 // gets a random position in [0, length of member list)
 func (m *memberlist) getJoinPosition() int {
 	l := len(m.members.list)
--- a/swim/memberlist_test.go
+++ b/swim/memberlist_test.go
@@ -66,6 +66,34 @@ func (s *MemberlistTestSuite) SetupTest() {
 	}
 }
 
+func (s *MemberlistTestSuite) TestAddJoinList() {
+
+	joinList := s.changes
+
+	// We add all the members of this node to the joinList because we want
+	// the joinList to contain the change for the local member.
+	mems := s.node.disseminator.FullSync()
+	for _, mem := range mems {
+		joinList = append(joinList, mem)
+	}
+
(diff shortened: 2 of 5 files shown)


swim/memberlist.go
```diff
@@ -283,6 +283,22 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 	return applied
 }
 
+// AddJoinList adds the list to the membership with the Update
```
dansimau: Excellent comment.

swim/memberlist_test.go
```diff
@@ -66,6 +66,34 @@ func (s *MemberlistTestSuite) SetupTest() {
...
+	}
+
+	// We make sure that the joinList contains a change for the local member.
+	var localMember *Change
+	for _, mem := range joinList {
+		if mem.Address == s.node.Address() {
+			localMember = &mem
+		}
+	}
+	s.NotNil(localMember, "expected joinList to contain a change for the local member")
```
dansimau: Every time I see `NotNil`, I read it as `NotNils`. And I think: "What has Nils done _this_ time?!" :smirk:

dansimau:
I don't fully understand the specific details of this bug, but I picked up that the naming of methods was confusing and I see that this patch corrects it.

LGTM from me, but it would be good to get someone to deep dive into the behaviour and give a LGTM from that perspective.

swim/memberlist.go
```diff
@@ -283,6 +283,22 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
...
+// function. However, as a side effect, Update adds changes to
+// the disseminator as well. Since we don't want to disseminate
+// the potentially very large join lists, we clear all the
+// changes from the disseminator, except for the one change
+// that refers to the make-alive of this node.
+func (m *memberlist) AddJoinList(list []Change) {
+	applied := m.Update(list)
+	for _, member := range applied {
```
thanodnl: before this change we iterated over `res.Membership` and which would be `list` here, is there a specific reason that we change this to iterate only over the applied changes?

swim/memberlist.go
```diff
@@ -283,6 +283,22 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
...
+		if member.Address == m.node.Address() {
+			continue
+		}
+		m.node.disseminator.ClearChange(member.Address)
+	}
+}
```
severb: This is not ideal. Instead of calling `Update`, which does more than we need, and then cleanup the bad side effects, can we instead have a method that only does what is necessary? `Update` can also call the same function as part of its implementation.

swim/memberlist.go
CorgiMan: Well, not applied changes will not end up in the changes anyway so there would be nothing to remove.

swim/memberlist.go
CorgiMan: The change you are talking about is quite involved. I like it but this PR is focused to just fix the bug. Changing the flow of the code could have some unexpected repercussions.

I think we should keep your suggestion in mind and apply it later though.

severb: I agree, I've been looking at what this change would involve and it's not a trivial refactor. This is a small smell that we can live with.

swim/memberlist_test.go
thanodnl: **Not**hing ;)

thanodnl:
lgtm

PR #96, opened 2016-02-08
update tchannel-go to v1.0.2

CorgiMan:
tickcluster-testing looks good
Need to deploy on test cluster


Motiejus:
lgtm

severb:
I got a full sync loop in tickcluster on this branch.

severb:
Can't reproduce it neither on this branch on on master anymore. I do see an increased number of full syncs, but I think that's expected because the maxP fix is not included in the branch nor in master.

severb:
Let's Get The Money!

PR #97, opened 2016-02-08
Set a better default for maxP and fix maxP value after bootstrap.

severb:
Two things going on:
* `maxP` should never be lower than 15, considering the math used to adjust it.
* after bootstrap, the initial value for `maxP` doesn't reflect the ring size.

### DIFF
--- a/swim/node_bootstrap_test.go
+++ b/swim/node_bootstrap_test.go
@@ -195,6 +195,23 @@ func (s *BootstrapTestSuite) TestMalformedJSONFile() {
 	s.Error(err, "invalid character 'T' looking for beginning of value", "should fail to unmarhsal JSON")
 }
 
+func (s *BootstrapTestSuite) TestDisseminationCounterAfterBootstrap() {
+	s.peers = genChannelNodes(s.T(), 10)
+	bootstrapList := bootstrapNodes(s.T(), s.peers...)
+	waitForConvergence(s.T(), 5*time.Second, s.peers...)
+
+	s.Equal(s.node.disseminator.pFactor, s.node.disseminator.maxP, "Initial maxP value should equal pFactor")
+	s.node.Bootstrap(&BootstrapOptions{
+		Hosts:   bootstrapList,
+		Stopped: true,
+	})
+	waitForConvergence(s.T(), 5*time.Second, append(s.peers, s.tnode)...)
+
+	// there are 11 nodes in total, log10(11) rounded up is 2
+	expected := 2 * s.node.disseminator.pFactor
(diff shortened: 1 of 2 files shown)


severb:
It's worth noting that the if condition I removed doesn't appear in the node codebase.

CorgiMan:
That's a useful comment, thanks :+1:

swim/node_bootstrap_test.go
```diff
@@ -195,6 +195,22 @@ func (s *BootstrapTestSuite) TestMalformedJSONFile() {
...
 
+func (s *BootstrapTestSuite) TestDisseminationCounterAfterBootstrap() {
+	s.peers = genChannelNodes(s.T(), 10)
+	bootstrapList := bootstrapNodes(s.T(), s.peers...)
+	waitForConvergence(s.T(), 5*time.Second, s.peers...)
+
+	s.node.Bootstrap(&BootstrapOptions{
+		Hosts:   bootstrapList,
+		Stopped: true,
+	})
```
CorgiMan: Can we test here that maxP of s.node is not smaller than pFactor?

CorgiMan:
lgtm except the one comment. Have you tested with tickcluster etc? Are there effects on number of full syncs? I'll rebase your work on mine and test it on the cluster as soon as Ben's change has landed

swim/node_bootstrap_test.go
severb: yeap!

severb:
@CorgiMan Good question. Here's how the full syncs look now (each cell has the number of observed full syncs):

cluster size (nodes) | 1st run | 2nd run | 3rd run | 4th run | 5th run
---------------------------|-----------|------------|----------|-----------|----------
3|0|0|0|0|0
5|1|0|0|0|0
15|0|3|1|0|0
45|15|2|1|6|2

I'd say it's a good improvement over what we used to see before.

dansimau:
How did you gather these metrics? Can you post the command/script? It would be interesting for all to see.

severb:
@dansimau I've run tickcluster with different cluster sizes and counted the number of full sync log messages printed until the cluster converged. It's all manual :coffee:

dansimau:
MANUAL!??? I'm disgusted. :wink:

thanodnl:
lgtm

CorgiMan:
lgtm, before merging, let me deploy on ringpopd-go

severb:
Thanks! I'll wait with the merge.

CorgiMan:
Let's get the Merge!

PR #98, opened 2016-02-09
Feature/lint

dansimau:
This adds a lint target to the Makefile that does `golint` and `go vet`. It also adds the lint check to Travis.

This PR also fixes most of the lint and go vet errors already existing in the project.

For the remaining ones that we're aware of but are unable to address right away, there is an exclude file which turns the lint errors into warnings. This way we can benefit from the linters right away.

E.g.:

```
$ make lint
WARNING [excluded]: ./swim/disseminator.go:107: range var member copies Lock: swim.Member
WARNING [excluded]: ./swim/member.go:53: address passes Lock by value: swim.Member
WARNING [excluded]: ./swim/member.go:57: incarnation passes Lock by value: swim.Member
WARNING [excluded]: ./swim/memberlist.go:128: Pingable passes Lock by value: swim.Member
WARNING [excluded]: ./swim/stats_test.go:101: range var member copies Lock: swim.Member
WARNING [excluded]: ./swim/test_utils.go:135: range var expected copies Lock: swim.Member
$ 

(exit code 0)
```

This also adds linting to the pre-commit hook as part of `make setup`.

@uber/ringpop


.travis.yml
```diff
@@ -20,8 +22,6 @@ env:
 matrix:
   allow_failures:
     - env: RUN=test/update-coveralls
-    # Temporarily allow the integration tests to fail until we fix them
-    - env: RUN="make test-integration"
```
Motiejus: Is this still needed, even with serial integ tests?

dansimau: Sorry. I just rebased off master and pushed again. Take another look here. Thanks!

Makefile
```diff
@@ -12,7 +12,17 @@ export GOPATH = $(GODEPS):$(USER_GOPATH)
...
+		   github.com/golang/lint/golint
+
+# Automatically gather packages
+PKGS = $(shell find . -type d -maxdepth 3 \
+	! -path '*/.git*' \
+	! -path '*/_*' \
+	! -path '*/Godeps*' \
+	! -path '*/test*' \
+	! -path '*/examples*' \
+)
```
Motiejus: Why not `/test*` and `/exmaples*`? Do we have a good reason for them to be not linted?

dansimau: `/test/` yes - it includes mocks that have lint errors that are outside our control. Since they're only test utilities they're less important.

`/examples` - this one is harder to justify excluding. I checked again and actually the lint errors from the examples are mostly in generated code. So I've updated the PR to exclude the generated code (instead of the entire examples directory).

Makefile
```diff
@@ -23,6 +33,20 @@ clean-mocks:
...
+
+	@for pkg in $(PKGS); do \
+		{ \
+			 find $$pkg -maxdepth 1 -mindepth 1 -type f -name '*.go' \
+				-exec go tool vet -printfuncs Logf:3 {} + 2>&1 ; \
+		} | tee -a /tmp/lint.log ; \
+	done;
+
+	@[ ! -s /tmp/lint.log ]
+
```
Motiejus: My suggestion:

```patch
diff --git a/Makefile b/Makefile
index ad406ba..e27b6f8 100644
--- a/Makefile
+++ b/Makefile
@@ -34,18 +34,7 @@ clean-mocks:
 	rm -rf test/thrift/pingpong/
 
 lint:
-	@:>/tmp/lint.log
-
-	@-golint ./... | grep -Ev '(test|gen-go)/' | tee -a /tmp/lint.log
-
-	@for pkg in $(PKGS); do \
-		{ \
-			 find $$pkg -maxdepth 1 -mindepth 1 -type f -name '*.go' \
-				-exec go tool vet -printfuncs Logf:3 {} + 2>&1 ; \
-		} | tee -a /tmp/lint.log ; \
-	done;
-
-	@[ ! -s /tmp/lint.log ]
+	@./scripts/lint $(PKGS)
 
 mocks:
 	test/gen-testfiles
diff --git a/scripts/lint b/scripts/lint
new file mode 100755
index 0000000..343f42d
--- /dev/null
+++ b/scripts/lint
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Lints given Go packages (via argv) and returns correct exit status.
+
+set -eu
+
+if [ $# -eq 0 ]; then
+    echo "Usage: $0 package1 ..."
+    exit 1
+fi
+
+OUTPUT=$(
+golint ./... | grep -Ev '(test|gen-go)/'
+for pkg in $@; do
+    find $pkg -maxdepth 1 -mindepth 1 -type f -name '*.go' \
+        -exec go tool vet -printfuncs Logf:3 {} + 2>&1 ;
+done;
+)
+
+if [ -n "${OUTPUT}" ]; then
+    echo "${OUTPUT}"
+    exit 1
+fi
+
+exit 0
```

Makefile
```diff
@@ -33,6 +57,8 @@ setup:
  		exit 1; \
 	fi
 
+	ln -sf ../../scripts/pre-commit .git/hooks/pre-commit
```
Motiejus: I might want to overwrite my `pre-commit`... Maybe we shouldn't force re-overwrite it in that case.

Makefile
dansimau: The thing I don't like about this script is that it takes a package (or packages) as an argument, yet runs `golint` across `./...`.

There are several solutions, but for the sake of not sinking more time into this I'm just gonna go with what we've got.

I've updated the PR to output to `lint.log` and will clean it up afterwards.

We can mark this as something to be improved in the future but I think for now we should just move on.

Thank you for suggestions though; you're quite right about the /tmp thing.

Makefile
dansimau: Quite right. Updated.

swim/disseminator.go
```diff
@@ -137,7 +137,7 @@ func (d *disseminator) bumpPiggybackCounters(changes []Change) {
 			continue
 		}
 
-		c.p += 1
+		c.p++
```
CorgiMan: is this a lint thing? I like the other one better. But don't mind having it like this either

swim/gossip_test.go
```diff
@@ -97,11 +97,11 @@ func (s *GossipTestSuite) TestUpdatesArePropagated() {
 
 	s.g.ProtocolPeriod()
 
-	expectedMembers := []Member{
-		Member{Address: "127.0.0.1:3003", Status: Alive},
-		Member{Address: "127.0.0.1:3004", Status: Faulty},
-		Member{Address: "127.0.0.1:3005", Status: Suspect},
-		Member{Address: "127.0.0.1:3006", Status: Leave},
+	expectedMembers := []*Member{
```
CorgiMan: Is this because of lock copy?

CorgiMan: Ah I see, because we change the receiver to pointer receiver we need pointers here

swim/disseminator.go
dansimau: Yes, it's a lint thing.
https://github.com/golang/lint/blob/master/lint.go#L1231

swim/gossip_test.go
dansimau: Yep.

swim/member.go
```diff
@@ -50,11 +50,11 @@ type Member struct {
 }
 
 // suspect interface
-func (m Member) address() string {
+func (m *Member) address() string {
```
CorgiMan: I am not opposed to this, but why was this change made?

swim/memberlist.go
```diff
@@ -166,10 +166,10 @@ func (m *memberlist) RandomPingableMembers(n int, excluding map[string]bool) []*
 }
 
 // returns an immutable slice of members representing the current state of the membership
-func (m *memberlist) GetMembers() (members []Member) {
+func (m *memberlist) GetMembers() (members []*Member) {
 	m.members.RLock()
 	for _, member := range m.members.list {
-		members = append(members, *member)
+		members = append(members, member)
```
CorgiMan: We are returning pointers to members protected by the lock. I don't think this is safe

swim/member.go
dansimau:
```
./swim/member.go:53: address passes Lock by value: swim.Member
```

I believe calling `m.address()` previously would result in `m` being copied. `m` has a mutex on it, so `go vet` calls out this behaviour as potentially harmful.

Correct me if I'm wrong, but it seems to be convention in Golang to pass by reference unless copying is necessary. For this reason, we generally use pointer receivers by default.

I am unaware if the previous code was intentional or not. I can't see any reason why these values would need to be copied.

CorgiMan: The member has a lock embedded but it is used nowhere I believe. Best to double check that... And maybe we should change that in different PR

swim/memberlist.go
CorgiMan: Consider removing the lock on member and leaving this as it is. Messing with copying v.s. referencing might result in unexpected behaviour. If we remove the lock without generating compiler errors, it is not used.

dansimau: Good point. Is it unsafe though? We need to answer this question definitively.

There are two things here that have mutexes:

* `m.members`: an anonymous struct on the memberlist. Presumably this lock is for mutating the list.
* `Member`s themselves. Presumably, this lock is for changing the status or incarnation number of the member.

`GetMembers()` returns a new slice containing pointers to the members. I can only see this used in two places. In both cases, data from the member is being read without acquiring a read lock.

One is for generating stats and the other place is for creating a list of changes for a full sync. Might we be OK to just change these places to acquire the read lock on the member before reading data?

swim/member.go
dansimau: Interesting. If we remove the lock, we could satisfy the linter while touching less code elsewhere.

swim/memberlist.go
dansimau: ack

scripts/lint/lint-warn
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+#
+# lint-warn takes the output of `go tool vet` and matches lines against an
+# exclude file. If an output line matches an exclude, it is printed to stdout,
+# otherwise it is printed to stderr.
+#
+# If anything is printed to stderr, the program exits with exit code 1.
+
```
Motiejus: `set -u`?

Motiejus:
LGTM

mbrysa:
LGTM

PR #99, opened 2016-02-09
Fix integration tests

dansimau:
This PR introduces two commits which:
- Changes the integration tests to run in serial. This fixes a lot of failures (but we haven't identified why yet).
- Uses an updated tap-filter to show more useful output with a test fails on Travis.

Thanks!

@uber/ringpop

### DIFF
--- a/test/run-integration-tests
+++ b/test/run-integration-tests
@@ -2,16 +2,11 @@
 #
 # Run integration tests for ringpop-go.
 #
-# Integration tests for different cluster sizes are run in parallel. Success
-# output is suppressed and only the failures are shown.
-#
-# 2015-01-14
-#
 set -eo pipefail
 
 declare project_root="${0%/*}/.."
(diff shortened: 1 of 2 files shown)


test/run-integration-tests
```diff
@@ -130,7 +124,7 @@ run-test-for-cluster-size() {
 }
 
 #
-# Run the integration tests against the testpop binary.
+# Run the integration tests against the testpop binary, in parallel.
```
Motiejus: Add a "TODO: failing as of `date +%F`." Also, do we have an issue in the issue tracker?

Motiejus:
LGTM

test/run-integration-tests
dansimau: Do you mean an issue to parallelise the tests again?

CorgiMan:
STAMP

PR #112, opened 2016-02-23
Fix race in membership and ring consistency

Motiejus:
Without this change, membership and ring could be not in sync. Scenario:
A - current node.
B - alive (according to C).
B - suspect (according to D).
C - thinks B is faulty.
D - thinks B is alive.

At the same time:
1. C gossips A that B is faulty. A removes B to the membership list.
2. D gossips A that B is alive. A adds B from the membership list.

Changes in the ring:
3. The lock on m.members is released, so updates on Ring due to (1) and
   (2) can execute in any order. In our example, the operations are
   reversed:
3.1. A adds B to the ring list due to (2).
3.2. A removes B to the ring list due to (1).

Now B is in the membership list, but not in the ring.

This is the second iteration of the change. First iteration did not
introduce a new lock, but a new invariant: a goroutine needs to lock
both memberlist and disseminator, lock memberlist first, then
disseminator. However, the invariant caused requestAdminStats to fail on
travis-ci, and I couldn't find out why.

### DIFF
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -45,6 +45,14 @@ type memberlist struct {
 		checksum  uint32
 		sync.RWMutex
 	}
+
+	// TODO: rework locking in ringpop-go (see #113). Required for Update().
+
+	// Updates to membership list and hash ring should happen atomically. We
+	// could use members lock for that, but that introduces more deadlocks, so
+	// making a short-term fix instead by adding another lock. Like said, this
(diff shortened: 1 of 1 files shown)


danielheller:
lgtm

danielheller:
Oops, wrong button, sorry Motiejus.  lgtm!

thanodnl:
lgtm

CorgiMan:
lgtm

PR #114, opened 2016-02-23
Make suspicion flexible for other states

severb:
This is a `suspicion` refactor to accommodate upcoming changes for reaping. Right now `suspicion` does time based transitions from `Suspect` to `Faulty`. I want to reuse the same code to support similar transitions from `Faulty` to `TombStone` and from `TombStone` to evicted.


swim/suspicion.go
```diff
@@ -33,62 +33,82 @@ type suspect interface {
 	incarnation() int64
 }
 
+type timerstate struct {
```
dansimau: Why not simply embed time.Timer, add the `state` field and continue to call this a timer?

dansimau:
> Right now suspicion does time based transitions from Suspect to Faulty. I want to reuse the same code to support similar transitions from Faulty to TombStone and from TombStone to evicted.

I think in this case, the code needs to be moved out of "suspicion" and named something different / more generic.

This patch would change the purpose of `suspicion` to do stuff that has nothing to do with suspicion, which I find confusing.

swim/suspicion.go
severb: Good idea

swim/suspicion.go
```diff
@@ -33,62 +33,82 @@ type suspect interface {
...
+	timer *time.Timer
+	state string
```
Motiejus:
```
// at which state the timer was started
```

severb: :+1:

Motiejus:
Looks good, though I also agree with @dansimau comment about embedding `state` in `time.Timer`.

severb:
@dansimau For now, because there is no change in functionality, I think `suspicion` is still a good name. As soon as we start adding multiple transitions, we may need a better name. Do you have any suggestions?

severb:
@dansimau @Motiejus I amended the commit with your suggestions.

dansimau:
RE: our chat, I think `suspicion` needs to be generalised. Otherwise landing this PR as a logical change does not make sense.

swim/suspicion.go
```diff
@@ -33,62 +33,85 @@ type suspect interface {
...
 }
 
+type timer struct {
+	*time.Timer
+	// at which state the timer was started
+	state string
+}
+
 // Suspicion handles the suspicion sub-protocol of the SWIM protocol
 type suspicion struct {
```
thanodnl: Can we come up with a better name for this struct if it will be doing more than the suspicion timer? First that comes to mind is 'delayed state change' but that is a long word. Anyone having a better suggestion?

severb: IIRC I called it `statetimer` originally.

swim/suspicion.go
```diff
@@ -33,62 +33,85 @@ type suspect interface {
...
 	return suspicion
 }
 
-func (s *suspicion) Start(suspect suspect) {
-	s.withLock(func() {
-		if !s.enabled {
-			s.logger.Warn("cannot start suspect period while disabled")
-			return
-		}
+func (s *suspicion) StartSuspect(suspect suspect) {
```
thanodnl: The type `suspect` is also out of place here since it is more than the suspect, it is more of a `subject` of the change.

swim/suspicion.go
thanodnl: That would work for me

severb: You and @dansimau gonna need to armwrestle on the name, he suggested to change it back to `timer`.

swim/suspicion.go
```diff
@@ -33,62 +33,85 @@ type suspect interface {
...
+		default:
+			panic("unknown state")
+		}
 	})
+	s.timers[suspect.address()] = &timer{
+		Timer: t,
+		state: state,
+	}
+
+	s.logger.WithField(state, suspect.address()).Debugf("started member %s period", state)
```
thanodnl: A bit more information in the log message might be nice, something like: `Started state timer for member`.

And is it intentional that you encode the state as a field name in the `WithField`? it might be easier for machines to parse if you add specific fields for `address` and `state`

swim/suspicion.go
thanodnl: I think `statetimer` covers better what it does, it is not any timer imho.

dansimau: I didn't suggest changing the name of the struct. Only to keep it called a "timer" further down in the patch.

swim/suspicion.go
severb: I didn't touch this part, the diff is pretty messed; it may be easier to see the diff with split view.

swim/suspicion.go
severb: :+1:  I understand

swim/state_transitions.go
```diff
@@ -0,0 +1,198 @@
...
+
+// GetTimeout returns the timeout to apply on a transition from state
+func (s *StateTimeouts) GetTimeout(state string) time.Duration {
+	switch state {
+	case Suspect:
+		return s.Suspect
+	default:
+		panic(fmt.Sprintf("unknown state %q", state))
+	}
+}
```
severb: We can get rid of the `switch` here if we pass the timer explicitly in `ScheduleSuspectToFaulty`.

swim/state_transitions.go
```diff
@@ -0,0 +1,198 @@
...
+		node:     n,
+		timeouts: timeouts,
+		timers:   make(map[string]*transitionTimer),
+		enabled:  true,
+		logger:   logging.Logger("stateTransitions").WithField("local", n.Address()),
+	}
+}
+
+// ScheduleSuspectToFaulty starts the suspect timer. After the Suspect timeout the node will be declared faulty
+func (s *stateTransitions) ScheduleSuspectToFaulty(subject subject) {
```
severb: I like this name :+1:. Should we also rename the `Suspect` timeout to `SuspectToFaulty`?

swim/state_transitions.go
```diff
@@ -0,0 +1,198 @@
...
+		transition()
+	})
+
+	s.timers[subject.address()] = &transitionTimer{
+		Timer:      timer,
+		state:      state,
+		transition: transition,
+	}
+
+	s.logger.WithField("member", subject.address()).Debugf("scheduled state transition for member in %q state", state)
```
severb: I think we should put the state in a field and have a static message. It would be extra nice to include both the start state and the target state.

swim/state_transitions.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+type transitionTimer struct {
+	*time.Timer
+
+	// state represents the state the subject was in when the transition was scheduled
+	state string
+
+	// transition contains the transition to execute
+	transition func()
```
severb: I don't see `transition` filed used in this file. Does it help with testing?

swim/state_transitions_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (s *StateTimeoutsTestCase) TestTimeoutForUnknownState() {
+	defer func() {
+		if r := recover(); r == nil {
+			s.Fail("Expected a panic when looking up the timeout for an unknown state")
+		}
+	}()
+
+	// should trigger a panic
+	_ = s.timeouts.GetTimeout("unknown")
```
severb: :clap:

swim/state_transitions.go
thanodnl: It is actually not being used :), I put it on here at first because I named the struct `transitionTimer`. It had a timer already, but no transition. Now that all is implemented I will remove this field again since it is not used.

swim/state_transitions.go
thanodnl: Good one! This way we can get rid of the panic.

swim/state_transitions.go
thanodnl: I don't have an opinion on the name of the field in the `StateTimeouts` struct. But when we expose this variable in the configuration I like `suspectTimeout` or,

```yaml
timeouts:
  suspect: 5000
```

over the `SuspectToFaulty` name.

swim/state_transitions.go
```diff
@@ -0,0 +1,195 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package swim
+
+import (
+	"sync"
+	"time"
+
+	log "github.com/uber-common/bark"
```
dansimau: remove "log" alias and just call this "bark"

swim/state_transitions.go
```diff
@@ -0,0 +1,195 @@
...
+// subject is an interface to define the subject (eg. member) to transition state for
+type subject interface {
+	address() string
+	incarnation() int64
+}
+
+type transitionTimer struct {
+	*time.Timer
+
+	// state represents the state the subject was in when the transition was scheduled
```
dansimau: Excellent comment

swim/state_transitions.go
```diff
@@ -0,0 +1,195 @@
...
+	node *Node
+
+	timeouts StateTimeouts
+	timers   map[string]*transitionTimer
+	enabled  bool
+	logger   log.Logger
+}
+
+// StateTimeouts contains the timeouts for every state before transitioning to the new state
+type StateTimeouts struct {
```
dansimau: I don't fully understand what `StateTimeouts` is. Is struct of configuration values that are set as a one-time thing?

swim/state_transitions.go
dansimau: Why does `stateTransitions` take a `node` in the constructor, but then also the transition methods also take a `subject`? Aren't the `node` from the constructor and the `subject` one and the same?

swim/state_transitions.go
thanodnl: Yes, it is the struct that keeps all the timeouts that gets passed in to the `stateTransitions`. It could be constructed from the config and passed on to the `Node` which uses the merge function to fill in  fields that are not configured with the defaults.

swim/state_transitions.go
```diff
@@ -0,0 +1,195 @@
...
+			"member": subject.address(),
+			"state":  timer.state,
+		}).Debug("stopped scheduled state transition for member")
+	}
+
+	s.Unlock()
+}
+
+// Reenable state transition controller
+func (s *stateTransitions) Reenable() {
```
dansimau: Shouldn't `Reenable` just be called `Enable`?

dansimau: Also, can the comment explain in a bit more detail what is happening if you enable or disable? Like, does that pause the timers and resume them? Why would one actually disable and/or enable the entire set of state transitions (what is the use-case?)

swim/state_transitions.go
dansimau: That's what I came to. Can we make this more explicit (either in a comment, or in the naming)?

swim/state_transitions.go
thanodnl: `Disable` covers in its documentation that it cancels the outstanding timers.

I don't really know why they are in there, but they are called from `Node` on `Start` and `Stop`. It might have to do with enabling and disabling the gossip protocol which you can do from the admin endpoints.

swim/state_transitions.go
thanodnl: Nope, there is a difference between the `node` and a `subject`.

Node is the internal representation of the local node. The subject is the state of a remote peer. It is basically the change that is being gossiped.

severb:
LGT:moneybag:, good job.

swim/state_transitions.go
dansimau: Cool, makes sense.

However, I find it slightly strange that we're defining an interface of type "subject" (the `address()/incarnation()` interface).

Is it our intention that `ScheduleSuspectToFaulty` should be able to accept both a `Member` struct _and_ a `Change` struct?

And if yes, why do we not just have one interface that is used consistently throughout the swim package, with a common name? (e.g. `suspicion.go` defines an interface that is identical to this one).

thanodnl: As discussed in person, this interface came from the original `suspicion.go` file. I had a look to see if I could change it to an actual type but both [`Member`](https://github.com/uber/ringpop-go/blob/feature/reaping/swim/state_transitions_test.go#L111) and [`Change`](https://github.com/uber/ringpop-go/blob/feature/reaping/swim/state_transitions_test.go#L136-L138) are passed into the `ScheduleSuspectToFaulty`.

For now I think it is best to keep this interface as the input for the `stateTransitions`. After writing the `Ringpop - User Data Per Node` document I got some ideas to merge the `Member` and `Change` struct into one. That would be a good moment to revisit if we want to remove this interface.

dansimau: Yeah, makes sense. I think the passing around of these stats/data needs to be cleaned up in the package as a whole. I'm happy to land this as-is. The only suggestion I have is that we could comment on the `subject` interface just to say: "This is so we can accept both a Member and a Change" or something. Might make reading/understanding that a bit easier.

dansimau:
LGTM :+1:

swim/state_transitions.go
thanodnl: Adding a comment about this is a good one, will add it. After that I will squash and merge this PR.

PR #116, opened 2016-02-24
Feature/convergence logs

thanodnl:
Add the logs for convergence measurement to ringpop-go.

@severb can you have a look if I used the `logger` you made correctly?


hashring/hashring.go
```diff
@@ -28,7 +28,9 @@ import (
 	"strings"
 	"sync"
 
+	log "github.com/uber-common/bark"
```
dansimau: Last PR we started reverting from doing this. `bark` should just be called `bark`, especially now that there is a new logging entity (the `logger` package).

swim/memberlist.go
```diff
@@ -79,11 +84,21 @@ func (m *memberlist) ComputeChecksum() {
 	startTime := time.Now()
 	m.members.Lock()
 	checksum := farm.Fingerprint32([]byte(m.GenChecksumString()))
+	oldChecksum := m.members.checksum
 	m.members.checksum = checksum
 	m.members.Unlock()
+
+	if oldChecksum != checksum {
```
dansimau: This creates a new untested code path which reduces coverage. Not saying it's the wrong approach since coverage != code quality, but if we can avoid untested code paths easily, that would be good.

swim/memberlist.go
```diff
@@ -228,6 +243,16 @@ func (m *memberlist) MakeChange(address string, incarnation int64, status string
 		Status:            status,
 		Timestamp:         util.Timestamp(time.Now()),
 	}})
+
+	if len(changes) > 0 {
+		m.logger.WithFields(log.Fields{
+			"update": changes[0],
+		}).Debugf("ringpop member declares other member %s", changes[0].Status)
+
+		m.node.emit(ChangeMadeEvent{changes[0]})
```
dansimau: Is there a test for this new event?

swim/memberlist.go
```diff
@@ -228,6 +243,16 @@ func (m *memberlist) MakeChange(address string, incarnation int64, status string
 		Status:            status,
 		Timestamp:         util.Timestamp(time.Now()),
 	}})
+
+	if len(changes) > 0 {
+		m.logger.WithFields(bark.Fields{
+			"update": changes[0],
+		}).Debugf("ringpop member declares other member %s", changes[0].Status)
+
+		m.node.emit(ChangeMadeEvent{changes[0]})
```
dansimau: Is there a test for this new event?

swim/memberlist.go
```diff
@@ -280,6 +305,17 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 	if len(applied) > 0 {
 		oldChecksum := m.Checksum()
 		m.ComputeChecksum()
+
+		for _, change := range applied {
+			if change.Source != m.node.address {
+				m.logger.WithFields(bark.Fields{
+					"remote": change.Source,
+					// changes in ringpop go do not have an id.
+					// "updateId": change.id,
```
dansimau: Are changes in ringpop going to have an `updateId` in the immediate future? If not, this commented-out line seems like code smell.

hashring/hashring.go
```diff
@@ -76,6 +80,7 @@ func New(hashfunc func([]byte) uint32, replicaPoints int) *HashRing {
 		hashfunc: func(str string) int {
 			return int(hashfunc([]byte(str)))
 		},
+		logger: logging.Logger("ring"),
```
severb: Ideally, the logger has the "local" field set to the node's hostport. In this particular case, it feels wrong to pass that info just to construct the logger, so I'm fine with the way it is.

dansimau: Just curious, but why?

severb: That's a good question, I did it because this is how it was in node. I think the reason for it is to differentiate between log messages from different nodes, for example, in tickcluster.

thanodnl: I like the idea of having the local in every log message from ringpop, because it makes it easy to reason about which member logged a line. But maybe we should move that logic into 1 place instead of passing that field in with every module logger. It also makes it transparent for all the modules in ringpop instead that they all call the WithField on their own logger.

swim/memberlist.go
thanodnl: I think it will actually be good to have this branch tested. If only to make sure the code works and does not dereference things. I will work on testing all log lines.

swim/memberlist.go
thanodnl: Nope, good question. I initially added the event so I could add all the logs to an event handler. Now this event is not even used so I am inclined to remove this event completely from this PR.

hashring/hashring.go
dansimau: @thanodnl +1 maybe something for the future.

When you say "reason about which member logged the line", do you mean, it's easier for a human to see which node it's from given a stream of logs from all nodes?

I would have hoped that log analysis (ELK?) services are able to tell where logs came from.

But you're right, if we do need to log this, I think we should do it in one place. Probably at the root-logger level.

swim/memberlist.go
thanodnl: Good question.

In the shared integration tests between node and go we have marked the `id` on a change as optional so it works on both implementations. So it would be great to add an `id` to go and make it a required field in the integration tests.

That said. This id is not used _anywhere_ so it add little value. With this I think it will not be added soon to ringpop-go.

I will remove this 'code smell' :)

hashring/hashring.go
thanodnl: I'm pretty sure the hostname is visible in log analysis, but it is missing the port number. When we run multiple instances on 1 machine (port 0) it is hard to reason about the identity of the ringpop member that logged the line which I find sometimes very handy.

severb: I agree with what you're saying @thanodnl, but that train has left the station. This is something I wanted too, to be able to pass loggers around and set the "local" field once in one place. Unfortunately, with the current implementation, this is impossible.

dansimau: I don't think the train has left the station (nothing is impossible :wink:)  - I don't even think this is difficult to achieve.

The root-level logger just needs to have a `WithFields` implementation with a way to set the value of the `local` field from outside. The identity resolver can then just set it as soon as it knows it.

Anyway, outside the scope of this PR....

swim/memberlist.go
thanodnl: Turns out the branch is covered during testing: https://coveralls.io/builds/5201500/source?filename=swim%2Fmemberlist.go#L91

dansimau: :+1:

dansimau:
LGTM :+1:

severb:
LGT:moneybag:

PR #118, opened 2016-02-25
Test Go 1.6 on Travis

dansimau:
This PR adds test runs to Travis on Go 1.6.

Also includes these changes:

* Make the lint errors obviously different to the warnings by prepending the line with "ERROR"
* Adds new lint errors to the exclude/ignore list (all lint errors are related to the same issue though; copying `swim.Member` locks)
* Run the coveralls upload only once after the Travis builds have succeeded

PR #120, opened 2016-03-01
Remove `Hosts` and `File` bootstrapOptions

Motiejus:
They can be replaced by implementing a DiscoverProvider instead. We
provide two default implementations:

1. discovery.statichosts (compatible with old `Hosts` attribute).
2. discovery.staticjson (compatible with old `File` attribute).

A separate pull request which fixes `Bootstrap()` not to have the `StaticHosts`-specific workaround will follow.


CHANGES.md
```diff
@@ -1,6 +1,12 @@
 ringpop-go changes
 ==================
 
+v0.x.y (to be released)
```
thanodnl: Love the addition of this to the change log. Was thinking of exactly this so we keep track during development on what gets added. However I was thinking about calling this `Dev` since we don't know the version to be released. What are you thoughts on calling this section something like `Dev` or `Development`?

discovery/staticjson/lib.go
```diff
@@ -18,11 +18,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package swim
+package staticjson
```
thanodnl: not really sure if this package name captures what it does.

1. It is not necessarily static as it reads the file every time you call `Hosts`
2. This name does not tell that it is reading the json from a file

I would suggest `jsonfile` as the package name.

ringpop.go
```diff
@@ -330,11 +331,10 @@ func (rp *Ringpop) setState(s state) {
 // Bootstrap starts communication for this Ringpop instance.
 //
 // When Bootstrap is called, this Ringpop instance will attempt to contact
-// other instances from a seed list provided either in the BootstrapOptions or
-// as a JSON file.
+// other instances from the DiscoverProvider.
 //
 // If no seed hosts are provided, a single-node cluster will be created.
-func (rp *Ringpop) Bootstrap(userBootstrapOpts *swim.BootstrapOptions) ([]string, error) {
+func (rp *Ringpop) Bootstrap(userbootstrapOpts *swim.BootstrapOptions) ([]string, error) {
```
thanodnl: Is there a specific reason that you changed that capitalization of `userBootstrapOpts` ? I find the one with capital `B` easier to read.

ringpop.go
```diff
@@ -346,14 +346,23 @@ func (rp *Ringpop) Bootstrap(userBootstrapOpts *swim.BootstrapOptions) ([]string
 	if err != nil {
 		return nil, err
 	}
+	hosts, err := userbootstrapOpts.DiscoverProvider.Hosts()
+	if err != nil {
+		rp.logger.WithField("error", err).Info("Failed to get boostrap hosts")
```
thanodnl: You are logging an `error` on `info` level this seems counterintuitive.

ringpop.go
```diff
@@ -346,14 +346,23 @@ func (rp *Ringpop) Bootstrap(userBootstrapOpts *swim.BootstrapOptions) ([]string
...
-	// check we're in the bootstrap host list and add ourselves if we're not
+	// Check we're in the bootstrap host list and add ourselves if we're not
 	// there. If the host list is empty, this will create a single-node
 	// cluster.
-	bootstrapOpts := *userBootstrapOpts
-	if len(bootstrapOpts.File) == 0 && !util.StringInSlice(bootstrapOpts.Hosts, identity) {
-		bootstrapOpts.Hosts = append(bootstrapOpts.Hosts, identity)
+	// TODO: move this code to the caller of the Hosts().
+	bootstrapOpts := *userbootstrapOpts
+	switch bootstrapOpts.DiscoverProvider.(type) {
```
thanodnl: I thought we discussed to treat all the `DiscoveryProviders` the same. What happens if there is an other provider that does not report the current node as part of the list?

discovery/statichosts/lib.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Hosts just returns the static list of hosts that was provided to this struct
+// on construction.
+func (p *StaticHostList) Hosts() ([]string, error) {
+	return p.hosts, nil
+}
+
+// New creates a provider with a static list of hosts.
+func New(hosts []string) *StaticHostList {
```
thanodnl: Have you considered using [`hosts ...string`](https://golang.org/ref/spec#Passing_arguments_to_..._parameters) here? It makes it easier to use if you hardcode the hosts in your application since you don't have to initialize the slice your self.

If you happen to already have a slice you can still pass it in to this function via `New(yourHostsLists...)`

CHANGES.md
dansimau: I also did this, see:
https://github.com/uber/ringpop-go/pull/118/files#diff-8b1c3fd0d4a6765c16dfd18509182f9dR4

I called it "0.3.1", as I intend it to be in the next _minor_ release.

discovery/staticjson/lib.go
dansimau: +1 with Niels. I thought the same thing.

ringpop.go
dansimau: +1 Niels. Why was this changed?

discovery/staticjson/lib.go
thanodnl: **UCH** Nils **UCH**

CHANGES.md
severb: I think whatever works as long as we don't have a fixed number. I used "unreleased" before.

Motiejus: I like the idea. `0.DEV`.

discovery/staticjson/lib.go
Motiejus: OK.

ringpop.go
Motiejus: I refactored the whole section, `userbootstrapOpts` and the likes are gone from this file.

dansimau: I think the point is "userbootstrap" is not a word. So if you're doing camel case, it should be "userBootstrap"...

ringpop.go
Motiejus: Removed this message completely in the other incarnation.

ringpop.go
Motiejus: That was the idea to first post a compatible change, and later, in a separate PR, fix that (see the PR message).

I will send an update with this section fixed.

discovery/statichosts/lib.go
Motiejus: Great idea.

Motiejus:
Like discussed with @thanodnl, I`ll add some explicit tests for DiscoverProviders.

CHANGES.md
```diff
@@ -1,6 +1,14 @@
 ringpop-go changes
 ==================
 
+v0.DEV (to be released)
+-----------------------
+
+* Remove `File` and `Host`-based discover providers in favor of
```
dansimau: This is a breaking change, so I think we need to provide an additional section: "release notes" which gives clear instructions on how to migrate from the old API to the new. There is a "release notes" section further down in this readme to get an example from.

Motiejus: Good one, thanks! Added.

thanodnl:
LGTM

dansimau:
LGTM. 

I think we could do some bike shedding over the exact API, e.g. here's another option:

```
import github.com/uber/ringpop-go/discovery

discovery.StaticHostList([]string{"1.1.1.1:1234", "1.1.1.2:1234"})
discovery.JSONFile("/path/to/file")
```

I think this is a nicer/cleaner API, though, I am undecided about whether it's as idiomatic in Go (or whether that even matters!). I can more imagine this type of API from Python (a la Kenneth's "API For Humans" style).

What do you guys think?

Motiejus:
@dansimau While I agree the API would look less to write and potentially prettier, we agreed to put every discovery provider to its own separate package, rather than have a single package for the discovery providers. Merging.

Motiejus:
Squashed all the commits and rebased on top of dev. Will wait for the final round of integ tests, and do a fast-forward merge.

PR #123, opened 2016-03-07
Reaping Faulty Nodes

thanodnl:
This branch is an implementation of reaping faulty nodes.

The basic design is:

1. Add extra status for a node that is called `tombstone`
2. A member that has the `tombstone` state will not be added to the checksum of the membership. The membership list with the tombstone has the same checksum as the membership list without the member. This makes it possible to remove the `tombstone` member from the membership list later on without the possibility for full sync mechanism of ringpop to make the deleted member appear again.
3. If a node joins a cluster where a tombstone is in the member list, or when a node receives an update of a `tombstone` for an unknown member it will be ignored.

These 3 rules together make it possible to 'forget' about a faulty node in a distributed system.

### DIFF
--- a/options.go
+++ b/options.go
@@ -29,6 +29,7 @@ import (
 	"github.com/uber/ringpop-go/hashring"
 	"github.com/uber/ringpop-go/logging"
 	"github.com/uber/ringpop-go/shared"
+	"github.com/uber/ringpop-go/swim"
 )
 
 type configuration struct {
@@ -43,6 +44,9 @@ type configuration struct {
 	// See funcs {Membership,Ring}ChecksumStatPeriod for specifics.
 	MembershipChecksumStatPeriod time.Duration
 	RingChecksumStatPeriod       time.Duration
+
+	// StateTimeouts keeps the state transition timeouts for swim to use
+	StateTimeouts swim.StateTimeouts
 }
 
 // An Option is a modifier functions that configure/modify a real Ringpop
@@ -233,6 +237,45 @@ func RingChecksumStatPeriod(period time.Duration) Option {
 			return errors.New("ring checksum stat period invalid below 10 ms")
 		}
 		r.config.RingChecksumStatPeriod = period
+		return nil
+	}
+}
+
+// SuspectPeriod configures the period it takes ringpop to declare a node faulty
+// after ringpop has first detected the node to be unresponsive to a healthcheck.
+// When a node is declared faulty it is removed from the consistent hashring and
+// stops forwarding traffic to that node. All keys previously routed to that node
+// will then be routed to the new owner of the key
+func SuspectPeriod(period time.Duration) Option {
+	return func(r *Ringpop) error {
+		r.config.StateTimeouts.Suspect = period
+		return nil
+	}
+}
+
+// FaultyPeriod configures the period Ringpop keeps a faulty node in its memberlist.
+// Even though the node will not receive any traffic it is still present in the
+// list in case it will come back online later. After this timeout ringpop will
+// remove the node from its membership list permanently. If a node happens to come
+// back after it has been removed from the membership Ringpop still allows it to
+// join and take its old position in the hashring. To remove the node from the
+// distributed membership it will mark it as a tombstone which can be removed from
+// every members membership list independently.
+func FaultyPeriod(period time.Duration) Option {
+	return func(r *Ringpop) error {
+		r.config.StateTimeouts.Faulty = period
+		return nil
+	}
+}
+
+// TombstonePeriod configures the period of the last time of the lifecycle in of
+// a node in the membership list. This period should give the gossip protocol the
+// time it needs to disseminate this change. If configured too short the node in
+// question might show up again in faulty state in the distributed memberlist of
+// Ringpop.
+func TombstonePeriod(period time.Duration) Option {
+	return func(r *Ringpop) error {
+		r.config.StateTimeouts.Tombstone = period
 		return nil
 	}
 }
--- a/swim/member.go
+++ b/swim/member.go
@@ -39,6 +39,9 @@ const (
 
 	// Leave is the member "leave" state
 	Leave = "leave"
+
+	// Tombstone is the member "tombstone" state
+	Tombstone = "tombstone"
 )
 
 // A Member is a member in the member list
@@ -94,7 +97,7 @@ func (m *Member) localOverride(local string, change Change) bool {
 	if m.Address != local {
 		return false
 	}
-	return change.Status == Faulty || change.Status == Suspect
+	return change.Status == Faulty || change.Status == Suspect || change.Status == Tombstone
 }
 
 func statePrecedence(s string) int {
@@ -107,6 +110,8 @@ func statePrecedence(s string) int {
 		return 2
 	case Leave:
 		return 3
+	case Tombstone:
+		return 4
 	default:
 		panic("invalid state")
 	}
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -107,6 +107,11 @@ func (m *memberlist) GenChecksumString() string {
 	var strings sort.StringSlice
 
 	for _, member := range m.members.list {
+		// Don't include Tombstone nodes in the checksum to avoid
+		// bringing them back to life through full syncs
+		if member.Status == Tombstone {
+			continue
+		}
 		s := fmt.Sprintf("%s%s%v", member.Address, member.Status, member.Incarnation)
 		strings = append(strings, s)
 	}
@@ -129,6 +134,31 @@ func (m *memberlist) Member(address string) (*Member, bool) {
 	m.members.RUnlock()
 
 	return member, ok
+}
+
+// RemoveMember removes the member from the membership list. If the membership has
+// changed during this operation a new checksum will be computed.
+func (m *memberlist) RemoveMember(address string) bool {
+	m.members.Lock()
+	member, hasMember := m.members.byAddress[address]
+	if hasMember {
+		delete(m.members.byAddress, address)
+		for i, lMember := range m.members.list {
+			if member == lMember {
+				// a safe way to remove a pointer from a slice
+				m.members.list, m.members.list[len(m.members.list)-1] = append(m.members.list[:i], m.members.list[i+1:]...), nil
+				break
+			}
+		}
+	}
+	m.members.Unlock()
+
+	if hasMember {
+		// if we changed the membership recompute the actual checksum
+		m.ComputeChecksum()
+	}
+
+	return hasMember
 }
 
 func (m *memberlist) MemberAt(i int) *Member {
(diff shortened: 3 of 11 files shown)


options.go
```diff
@@ -237,6 +241,45 @@ func RingChecksumStatPeriod(period time.Duration) Option {
 	}
 }
 
+// SuspectTimeout configures the period it takes ringpop to declare a node faulty
+// after ringpop has first detected the node to be unresponsive to a healtcheck.
```
dansimau: s/healtcheck/healthcheck :smile:

swim/node.go
```diff
@@ -58,7 +58,9 @@ type Options struct {
 func defaultOptions() *Options {
 	opts := &Options{
 		StateTimeouts: StateTimeouts{
-			Suspect: 5000 * time.Millisecond,
+			Suspect:   5 * time.Second,
+			Faulty:    1 * time.Hour,
```
dansimau: How did you arrive at 1 hour as a default for tombstone? I had it in my mind that it would be more like 24 hours...

options.go
```diff
@@ -237,6 +241,45 @@ func RingChecksumStatPeriod(period time.Duration) Option {
...
+// When a node is declared faulty it is removed from the consistent hashring and
+// stops forwarding traffic to that node. All keys previously routed to that node
+// will then be routed to the new owner of the key
+func SuspectTimeout(timeout time.Duration) Option {
```
dansimau: Good work adding the options. We definitely needed these.

swim/memberlist.go
```diff
@@ -225,6 +230,31 @@ func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
...
+func (m *memberlist) MakeTombstone(address string, incarnation int64) []Change {
+	m.node.emit(MakeNodeStatusEvent{Tombstone})
+	return m.MakeChange(address, incarnation, Tombstone)
+}
+
+func (m *memberlist) Evict(address string) {
+	if m.local.Address == address {
+		// don't evict ourself, this is an assertion that no execution
+		// path should lead to this state
+		panic("can't evict self")
```
dansimau: Panic seems a bit extreme. Is there no other way to handle this?

It seems that in the attempt to "not evict" oneself, we are panicking which will exit! That will certainly lead to an eviction! :wink:

swim/state_transitions_test.go
```diff
@@ -109,12 +109,42 @@ func (s *StateTransitionsSuite) TestSuspectBecomesFaulty() {
...
+	s.stateTransitions.timeouts.Faulty = time.Millisecond
+
+	s.m.MakeFaulty(s.suspect.Address, s.suspect.Incarnation)
+	member, _ := s.m.Member(s.suspect.Address)
+	s.Require().NotNil(member, "expected cannot be nil")
+
+	s.stateTransitions.ScheduleFaultyToTombstone(*member)
+	s.NotNil(s.stateTransitions.timer(member.Address), "expected state transtition timer to be set")
+
+	time.Sleep(5 * time.Millisecond)
```
dansimau: I think we have a mock clock in place in ringpop now. Can we use that here instead of sleeping?

swim/member.go
```diff
@@ -107,6 +110,8 @@ func statePrecedence(s string) int {
 		return 2
 	case Leave:
 		return 3
+	case Tombstone:
```
dansimau: Oh man, are we still doing this? Can we make these states proper integers so we can just do `Tomestone > Leave` etc?

dansimau:
Nice work on this PR dude. Kudos!

swim/memberlist.go
Motiejus: Yep, an error to the logs seems to be more appropriate.

swim/memberlist.go
```diff
@@ -225,6 +230,31 @@ func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
...
+	}
+	m.members.Lock()
+	if member, ok := m.members.byAddress[address]; ok {
+		delete(m.members.byAddress, address)
+		for i, lMember := range m.members.list {
+			if member == lMember {
+				// a safe way to remove a pointer from a slice
+				m.members.list, m.members.list[len(m.members.list)-1] = append(m.members.list[:i], m.members.list[i+1:]...), nil
```
Motiejus: Can we abstract memberlist manipulation to a separate module, or just a list of functions? Because memberlist update code is becoming sprinkled throughout this module...

swim/memberlist.go
```diff
@@ -356,6 +386,11 @@ func (m *memberlist) Apply(change Change) {
 	member, ok := m.members.byAddress[change.Address]
 
 	if !ok {
+		// avoid indefinite tombstones by not creating new nodes
+		// directly in this state
```
Motiejus: I can't find a code path that leads to this code. Could you explain why this is necessary?

swim/node.go
Motiejus: This seems indeed to aggressive. What are we trading off here?

- Large membership (low timeout) list versus.
- Time it takes to resolve a potential partition.

Unless there are more reasons why we want to keep this timeout low, I would keep this value to a few days.

test/run-integration-tests
```diff
@@ -60,12 +61,14 @@ prefix() {
 #
 fetch-ringpop-common() {
     if [ ! -e "$ringpop_common_dir" ]; then
-        run git clone --depth=1 https://github.com/uber/ringpop-common.git "$ringpop_common_dir"
+        run git clone --depth=1 https://github.com/uber/ringpop-common.git "$ringpop_common_dir" --branch "$ringpop_common_branch"
```
Motiejus: Should we instead run the integration test for both languages, and see if we can report an expected failure?

Discussed with @benfleis as well, that strikes us a more reasonable approach.

Motiejus:
Did you consider putting the original RFC to the [official docs][1]? That document would be really well placed to explain the motivation and inner-workings of this change for the non-Uber users of ringpop. The more high-level docs, the better to our users. :-)

[1]: http://ringpop.readthedocs.org/en/latest/running_ringpop.html

swim/state_transitions_test.go
Motiejus: :+1:

swim/node.go
thanodnl: The one hour default I put in here was just a wild guess. I don't see why it is beneficial to keep faulty nodes in the list for that long at the moment since faulty nodes are not used by anything yet.

When we decide we want to keep them longer we can change it while adding a feature that benefits from keeping the faulties around. I'm also happy with bumping the faulty time to a longer period if there is a good argument to keep them around for a longer period.

swim/member.go
thanodnl: There is a lot of work to do to make this a propper type and constant because the names are directly used in logging and in the marshaling and un-marshaling from and to json. I'm very much in favor of doing it, but in a contained PR because it will on its own already touch a lot of code which has nothing todo with what we are adding here.

Lets put a task on the board and properly plan it.

swim/memberlist.go
thanodnl: This code could be triggered in two cases.

1. When you join a cluster that has tombstones in them
2. When you receive a gossip for a tombstone after you have locally removed the member already.

If we would not have this here there is a chance for perpetual tombstones being gossiped around the system.

swim/node.go
dansimau: I'd say the argument for keeping them is more to do with if we implement a partition-healing strategy based on faulty nodes in the future.

The reaping time defines the longest possible partition we can recover from with such a strategy.

My initial reaction is that longer is better then, for this case.

What are the chances that—depending on the infrastructure/environment—a new node might come online with the same host/port as a faulty node?

swim/member.go
dansimau: I don't think there'll be much motivation to refactor it if we don't do it now.

If you think it's too big though, then I'd say yeah, let's leave it.

thanodnl: I think it is easier to refactor this when I take a stab at refactoring the `Member` struct and the `Change` struct in preparation of labels when that will be scheduled. For this change I'd rather stay with the precedent that has been set already.

test/run-integration-tests
thanodnl: This has been fixed in the integration tests.

With regards to this line, we want to keep it so we can develop to a different branch of the integration tests when we implement new features, but we want to make sure the branch of `ringpop-common` is set to `master` on our `dev` branch (before merging).

swim/member.go
dansimau: :+1:

swim/memberlist.go
thanodnl: I see your concerns here, I will look into what it takes to centralize this.

options.go
```diff
@@ -237,6 +241,45 @@ func RingChecksumStatPeriod(period time.Duration) Option {
...
+		return nil
+	}
+}
+
+// TombstonePeriod configures the period of the last time of the lifecycle in of
+// a node in the membership list. This period should give the gossip protocol the
+// time it needs to disseminate this change. If configured to short the node in
+// question might show up again in faulty state in the distributed memberlist of
+// Ringpop.
+func TombstonePeriod(period time.Duration) Option {
```
severb: Great comments! s/to short/too short/

swim/memberlist.go
```diff
@@ -225,6 +230,32 @@ func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
...
 
+func (m *memberlist) MakeTombstone(address string, incarnation int64) []Change {
+	m.node.emit(MakeNodeStatusEvent{Tombstone})
+	return m.MakeChange(address, incarnation, Tombstone)
+}
+
+func (m *memberlist) Evict(address string) {
+	if m.local.Address == address {
+		// We should not evict our selfs from the memberlist. This should not be reached, but we will make noise in the logs
+		m.logger.Error("ringpop tried to evict the local member from the memberlist, action has been prevented")
```
severb: This is my bad, please s/our selfs/ourselves/.

swim/memberlist.go
thanodnl: This code has been moved to its own function for removing a member, also triggers a recalculation of the checksum to make sure that if you remove a member that is part of the checksum it is reflected in there.

swim/memberlist.go
```diff
@@ -225,6 +255,21 @@ func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
...
 
+func (m *memberlist) MakeTombstone(address string, incarnation int64) []Change {
+	m.node.emit(MakeNodeStatusEvent{Tombstone})
+	return m.MakeChange(address, incarnation, Tombstone)
+}
+
+func (m *memberlist) Evict(address string) {
+	if m.local.Address == address {
+		// We should not evict ourselves from the memberlist. This should not be reached, but we will make noise in the logs
+		m.logger.Error("ringpop tried to evict the local member from the memberlist, action has been prevented")
```
Motiejus: `... This  is a bug, please report to ringpop team.`

swim/state_transitions_test.go
```diff
@@ -109,12 +113,42 @@ func (s *StateTransitionsSuite) TestSuspectBecomesFaulty() {
...
-	s.NotNil(s.stateTransitions.timer(member.Address), "expected suspicion timer to be set")
+	s.NotNil(s.stateTransitions.timer(member.Address), "expected state transtition timer to be set")
 
-	time.Sleep(5 * time.Millisecond)
+	s.clock.Add(5 * time.Millisecond)
 	s.Equal(Faulty, member.Status, "expected member to be faulty")
 }
 
+func (s *StateTransitionsSuite) TestFaultyBecomesTombstone() {
+	s.stateTransitions.timeouts.Faulty = time.Millisecond
```
Motiejus: Do we need small timeouts in the unit tests, also when we mock? Can we use the defaults?

Motiejus:
Minor non-blocking comments aside, LGTM.

severb:
Awesome work, let's reap the benefits!

PR #124, opened 2016-03-09
Add event that is fired after bootstrapping.

thanodnl:
This adds an event that fires after the bootstrapping has completed.

It can be used for applications to initialize extra systems that rely on `Lookup` and `WhoAmI` to work.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -31,6 +31,9 @@ type EventListener interface {
 	HandleEvent(event Event)
 }
 
+// A Bootstrapped event is sent when ringpop has bootstrapped
+type Bootstrapped struct{}
+
 // A RingChangedEvent is sent when servers are added and/or removed from the ring
 type RingChangedEvent struct {
 	ServersAdded   []string
--- a/ringpop.go
+++ b/ringpop.go
@@ -366,6 +366,9 @@ func (rp *Ringpop) Bootstrap(userBootstrapOpts *swim.BootstrapOptions) ([]string
 	rp.setState(ready)
 
 	rp.logger.WithField("joined", joined).Info("bootstrap complete")
+
+	rp.emit(events.Bootstrapped{})
+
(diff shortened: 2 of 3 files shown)


dansimau:
Nice one.

Question: we have a `Ready()` method, which I think—by its name—signifies: "comms are ready (swim bootstrapped) and the hash ring has been calculated".

Would it be more useful to send an event that called "ready"?

Is there a distinction between "ready" and "bootstrapped"?

(This is more a question of what our contract should be, rather than the name of the event)

severb:
I like @dansimau suggestion re. calling the event `Ready`; other than that LGTM.

dansimau:
@severb: note that this is not about changing the event name. It's about changing the event's contract.

If we had a "ready" event, we would need to make sure the hashring is also ready.

thanodnl:
I like the idea to have an event when ringpop changes its state to ready (or for that matter to any other state). This is easily implemented via the `setState` function and a `StateChanged` event. Separate from that I like to have a Bootstrapped event as implemented now.

However, I don't think that it should be the contract of the event that "comms are ready". That should already be the contract of the state and a prerequisite for changing the state. An event should only be there to signal the other system that the change happened.

That said, I think this `Bootstrapped` event and a `StateChanged` event could and should live next to each other in Ringpop. Lets see this PR as the one that implements the `Bootstrapped` event, and one for a `StateChanged` event could come later. But that event has a bit more to it.

1. `state` is a non-exported struct
2. In the current package setup `StateChanged` cannot live in the `events` package because it cannot import `ringpop` to get access to the `State` struct even if we export it due to circular dependencies

So the `StateChanged` event has a bit more to it, and I don't like having an event for every state that ringpop could be in.

thanodnl:
Replaced with events for `Ready` and `Destroyed`.

PR #125, opened 2016-03-09
Implement events for Ready and Destroyed.

thanodnl:
Give users the possibility to react to Ringpop becoming ready or being destroyed.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -49,3 +49,9 @@ type LookupEvent struct {
 	Key      string
 	Duration time.Duration
 }
+
+// Ready is fired when ringpop has successfully bootstrapped and is ready to receive requests and other method calls.
+type Ready struct{}
+
+// Destroyed is fired when ringpop has been destroyed and should not be responding to requests or lookup requests.
+type Destroyed struct{}
--- a/ringpop_test.go
+++ b/ringpop_test.go
@@ -29,11 +29,26 @@ import (
 	"github.com/stretchr/testify/suite"
 	"github.com/uber/ringpop-go/discovery/statichosts"
 	"github.com/uber/ringpop-go/events"
+	eventsmocks "github.com/uber/ringpop-go/events/test/mocks"
 	"github.com/uber/ringpop-go/forward"
 	"github.com/uber/ringpop-go/swim"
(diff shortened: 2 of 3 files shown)


events/events.go
```diff
@@ -49,3 +49,9 @@ type LookupEvent struct {
 	Key      string
 	Duration time.Duration
 }
+
+// Ready is fired when the state of ringpop is changed to ready
```
dansimau: This comment is for public documentation, so can I suggest:

`Ready is fired when ringpop has successfully bootstrapped and is ready to receive requests and other method calls.`

Or something

dansimau:
LGTM, though as discussed, it might be a good idea to add a test that checks the hashring is ready when the ready event is fired.

I realise this should have already been there. I think if we can we should add it in this PR though because otherwise we probably won't add it.

events/events.go
```diff
@@ -49,3 +49,9 @@ type LookupEvent struct {
 	Key      string
 	Duration time.Duration
 }
+
+// Ready is fired when ringpop has successfully bootstrapped and is ready to receive requests and other method calls.
+type Ready struct{}
```
CorgiMan: Isn't the convention to postfix "Event" on event structs?

thanodnl: I have been thinking about this, but since they are in the events package users would type something like `events.ReadyEvent` which I found too verbose.

ringpop_test.go
```diff
@@ -29,12 +29,27 @@ import (
 	"github.com/stretchr/testify/suite"
 	"github.com/uber/ringpop-go/discovery/statichosts"
 	"github.com/uber/ringpop-go/events"
+	eventsmocks "github.com/uber/ringpop-go/events/test/mocks"
```
CorgiMan: eventmocks sounds more natural but this is reaaalllly picky

events/events.go
CorgiMan: Yeah, I guess that's true, the swim package also has events, but what you have here is good

CorgiMan:
Looks good after you have addressed the comments. If you don't want to change anything, that's alright with me as well

PR #126, opened 2016-03-11
Heal partitions trough the discover provider

CorgiMan:


swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+	joinRes, err := sendJoinRequest(node, target, time.Second/2)
+	if err != nil {
+		return err
+	}
+
+	// Create changes for partition B. Partition B is the partition the targets node is part of.
+	csForB := node.disseminator.FullSync()
+	// discard changes that are not faulty and change faulty to suspect
+	for i := 0; i < len(csForB); i++ {
+		// TODO (wieger): make filtering more precice
```
Motiejus: Maybe it would be useful if you can explain what you mean by precise?

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Create changes for partition A. Partition A is the partition this node is part of.
+	csForA := joinRes.Membership
+	// discard changes that are not faulty and change faulty to suspect
+	for i := 0; i < len(csForA); i++ {
+		if csForA[i].Status != Faulty {
+			csForA[i] = csForA[len(csForA)-1]
```
Motiejus: You are removing an element from a slice in multiple places. Maybe it makes sense to have a small method for it?

Not for DRY, but because I would find it easier to read this:

```golang
csForA = remove_item(csForA, i)
```
rather than
```golang
csForA[i] = csForA[len(csForA)-1]
csForA = csForA[:len(csForA)-1]
```

I know it's only two lines, but they would make it easier to scan the business logic.

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+			csForA = csForA[:len(csForA)-1]
+			i--
+		} else {
+			csForA[i].Status = Suspect
```
Motiejus: On the other hand, I think it would be actually more understandable if you iteratively built a new slice, and appended stuff to it.

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package swim
+
+import "time"
+
+// AttemptHeal attempts to heal a partition between the node and the target.
+// It is the responsibillity of the caller to detect that the node is indeed
+// separated from the target.
```
Motiejus: Can you explain what it means "target separated from the caller"?

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+//
+// The algorithm to heal the partition works as follows:
+// The goal is to first let every node reincarnate, we do this by declaring nodes
+// suspect. For members that are faulty in partition B, we fabricate updates with
+// a suspect in stead of faulty status and disseminate them through A. We also do
```
Motiejus: `s/in stead/instead`

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+
+import "time"
+
+// AttemptHeal attempts to heal a partition between the node and the target.
+// It is the responsibillity of the caller to detect that the node is indeed
+// separated from the target.
+//
+// The algorithm to heal the partition works as follows:
+// The goal is to first let every node reincarnate, we do this by declaring nodes
+// suspect. For members that are faulty in partition B, we fabricate updates with
```
Motiejus: Please define actors first: what is partition A, partition B, and any node names you use here.

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+// a suspect in stead of faulty status and disseminate them through A. We also do
+// the reverse. We then create a connection from partition A to B and visa versa.
+// We do this by reincarnating this node and sending the respective update to the
+// target. The target has also been reincarnated due to previous actions, this
+// update is now applied on this node's membership to create the connection from
+// partition B to partition A.
+//
+// The heal alogrithm in more detail:
+// 1. send join request to the target, this tells us whether a partition exists
```
Motiejus: Not immediately clear how this tells us the partition exists, and it would be good to explain. "... a partition exists: we can see that members which are alive in the current node are declared faulty in target" ? Is that somewhat accurate?

swim/heal_partition.go
```diff
@@ -0,0 +1,127 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package swim
+
+import "time"
+
+// AttemptHeal attempts to heal a partition between the node and the target.
```
Motiejus: I think I would skip the algorithm here and explain it in `ringpop-common` docs, with diagrams. Because I find this much harder to understand than the one in the RFC.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,130 @@
...
+	if h.previousHostListSize < h.node.CountReachableMembers() {
+		h.previousHostListSize = h.node.CountReachableMembers()
+	}
+	if h.previousHostListSize < 1 {
+		h.previousHostListSize = 1
+	}
+	return h.baseProbabillity / float64(h.previousHostListSize)
+}
+
+// Heal iterates over the hostList that the discoverProvider provider. If the
```
Motiejus: `s/provider/provides`

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,130 @@
...
+// node encounters a host that is faulty or not in the membership, we pick that
+// node as a target to perform a partition heal with.
+func (h *discoverProviderHealer) Heal() {
+	h.node.emit(&AttemptHealEvent{})
+
+	// get list from discovery provider
+	if h.node.discoverProvider == nil {
+		return
+	}
+	hostList, err := h.node.discoverProvider.Hosts()
```
Motiejus: This might indefinitely block, so might be good idea to document this call site to the documentation of `discoverProvider` interface. E.g. "used by partition healer, so set a reasonable timeout to `Hosts()` method".

swim/join_sender.go
```diff
@@ -470,6 +464,23 @@ func (j *joinSender) MakeCall(ctx json.Context, node string, res *joinResponse)
...
+		App:         node.app,
+		Source:      node.address,
+		Incarnation: node.Incarnation(),
+		Timeout:     timeout,
+	}
+	res := &joinResponse{}
+	err := json.CallPeer(ctx, peer, node.service, "/protocol/join", req, res)
+	return res, err
+}
+
```
Motiejus: `MakeCall` seems to be pretty similar. Can't we re-use it, or refactor so it's usable from here? The only thing that seems to differ is error reporting?

swim/node.go
```diff
@@ -52,6 +52,9 @@ type Options struct {
 	RollupFlushInterval time.Duration
 	RollupMaxUpdates    int
 
+	PartitionHealPeriod           time.Duration
```
Motiejus: Maybe `PartitionHealInterval`, to be consistent? See variable names above.

swim/node.go
```diff
@@ -69,6 +72,9 @@ func defaultOptions() *Options {
 		RollupFlushInterval: 5000 * time.Millisecond,
 		RollupMaxUpdates:    250,
 
+		PartitionHealPeriod:           30 * time.Second,
+		PartitionHealBaseProbabillity: 3,
```
Motiejus: Would be useful to provide some documentation on what this value means to the caller, since it's configurable. I understand that the user can go and read the source, but a couple of examples (e.g. with `1`, you get this, with `3` -- this, etcetra) would be a worthy job to document.

swim/node.go
```diff
@@ -133,13 +145,17 @@ type Node struct {
...
-	gossip       *gossip
-	rollup       *updateRollup
+	channel          shared.SubChannel
+	discoverProvider discovery.DiscoverProvider
+	memberlist       *memberlist
+	memberiter       memberIter
+	disseminator     *disseminator
+	suspicion        *suspicion
+	gossip           *gossip
+	rollup           *updateRollup
```
Motiejus: You added only one field here -- `discoverProvider`. Maybe you can add it to the end of the list, so diff is smaller?

swim/node.go
```diff
@@ -133,13 +145,17 @@ type Node struct {
...
+
+	// when we get more healer strategies we can make the abstraction of a healer interface
```
Motiejus: :+1:

swim/node.go
```diff
@@ -243,6 +265,7 @@ func (n *Node) RegisterListener(l EventListener) {
 func (n *Node) Start() {
 	n.gossip.Start()
 	n.suspicion.Reenable()
+	n.healer.Start()
```
Motiejus: Like we discovered, don't forget to actually turn it on. :-) And maybe document this method: when is it expected to be called, and what it really does. You remember my confusion this morning. :-)

swim/ping_sender.go
```diff
@@ -132,6 +132,35 @@ func (p *pingSender) MakeCall(ctx json.Context, res *ping) <-chan error {
 	return errC
 }
 
+// sendPingWithChanges sends a special ping to the target with the given changes.
+// In normal pings the disseminator is consulted to create issue the changes,
+// this is not the case in this function. Only the given changes are transmitted.
```
Motiejus: Maybe we should generalize the "normal" ping and call that with arguments to send over? In the standard case, that would be the changes from disseminator. In partition case, it would be suspects.

I`m afraid that if we introduce another "special type of ping", we will keep baking "special versions of ping", which will not look great.

swim/join_sender.go
CorgiMan: I don't like how MakeCall's function signature (returns a channel). This function is a step towards refactoring it away.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,130 @@
...
+		h.previousHostListSize = 1
+	}
+	return h.baseProbabillity / float64(h.previousHostListSize)
+}
+
+// Heal iterates over the hostList that the discoverProvider provider. If the
+// node encounters a host that is faulty or not in the membership, we pick that
+// node as a target to perform a partition heal with.
+func (h *discoverProviderHealer) Heal() {
+	h.node.emit(&AttemptHealEvent{})
```
Motiejus: Is the `h.node.emit` placed in the right place?

swim/heal_via_discover_provider.go
Motiejus: I think we should also record a metric on how long this method takes.

PR #127, opened 2016-03-14
Release v0.4.0

dansimau:
PR for the release.

This PR is slightly messed up because dev was accidentally merged on Friday.

Please review the CHANGES.md

Thanks!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,29 +1,50 @@
 ringpop-go changes
 ==================
 
-v0.DEV (to be released)
------------------------
+v0.4.0
+------
 
-* Remove `File` and `Host`-based discover providers in favor of
-  `DiscoverProvider` interface #119
-* Ringpop will always assume the current host (`ringpop.node.identity`) is part
-  of the cluster. Previously, it was true for only `Host`-based discovery.
-* Add Go 1.6 testing on CI
-* Implemented reaping of faulty nodes. By default ringpop will remove faulty nodes
-  after 24 hours of being in faulty state.
+* Feature: Faulty nodes are now automatically reaped from nodes' membership
+  lists after (24 hours by default). #123
+* New options for controlling suspect and reaping times. #123
+* Add new `Ready` and `Destroyed` events to ringpop. #125
+* Add additional logging to bring ringpop-go on par with ringpop-node log
+  messages. #116
+* Fix bug where Ringpop automatically added itself to the bootstrap hosts for
+  host-based bootstrapping, but not other bootstrapping methods. #120
+* Fix race condition where membership and hashring could be inconsistent with
+  each other. #112
+* Remove `File` and `Host` options from bootstrap options in favor of
+  `DiscoverProvider` interface. #120
+* Add Go 1.6 to testing on CI
 
 ### Release notes
 
-`BootstrapOptions.File` and `BootstrapOptions.Hosts` are replaced with
+#### Version incompatibility in protocol
+
+Since 0.4.0 introduces a new node/member state, 0.4.0 is not backwards-compatible with previous versions.
+
+Note rolling upgrades with older versions do work, but undefined behaviour will occur if two versions run in parallel for longer than the `FaultyPeriod` (default 24 hours).
+
+#### Changes to Bootstrap
+
+This release contains a breaking change to the options provided to the
+`ringpop.Bootstrap` call.
+
+`BootstrapOptions.File` and `BootstrapOptions.Hosts` have been replaced with
 `BootstrapOptions.DiscoverProvider`. `DiscoverProvider` is an interface which
 requires a single method:
 
-    type DiscoverProvider interface {
-        Hosts() ([]string, error)
-    }
+```go
+type DiscoverProvider interface {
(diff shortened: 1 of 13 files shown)


CHANGES.md
```diff
@@ -1,18 +1,29 @@
...
+v0.4.0
+------
 
-* Remove `File` and `Host`-based discover providers in favor of
-  `DiscoverProvider` interface #119
-* Ringpop will always assume the current host (`ringpop.node.identity`) is part
-  of the cluster. Previously, it was true for only `Host`-based discovery.
-* Add Go 1.6 testing on CI
+* Feature: Faulty nodes are now automatically reaped from nodes' membership
+  lists after 24 hours (by default). #123
```
Motiejus: I assume it's configurable? I would say "(by default 24 hours)", that makes it clearer that timeout is configurable.

CHANGES.md
```diff
@@ -1,18 +1,29 @@
...
+* Add new `Ready` and `Destroyed` events to ringpop. #125
+* Add additional logging to bring ringpop-go on par with ringpop-node log
+  messages. #116
+* Fix bug where Ringpop automatically added itself to the bootstrap hosts for
+  host-based bootstrapping, but not other bootstrapping methods. #120
+* Fix race condition where membership and hashring could be inconsistent with
+  each other. #112
+* Remove `File` and `Host` options from bootstrap options in favor of
+  `DiscoverProvider` interface. #120
+* Add Go 1.6 to testing on CI
```
Motiejus: In the description, it's not 100% clear what is a changing behavior, versus internal change. I think it would be nice to split them out.

As an example, I would point to documentation of [Django release notes][1]: they are notorious on their backwards-compatibility promise, and in the way how they document it from release to release. Every described backwards-incompatible change is actionable.

[1]: https://docs.djangoproject.com/en/1.9/releases/1.9/#backwards-incompatible-changes-in-1-9

dansimau: Backwards-incompatible changes (or other things users need to be aware of) are documented in the release notes.

Is that what you mean?

Motiejus: Yes, that's what I mean. I'll try to be the devil's advocate here (a.k.a. customer's). Changelog should help me to answer the question: _how should I change my code base to stay compatible?_ Let's go 1-by-1:

> Feature: Faulty nodes are now automatically reaped from nodes' membership lists after 24 hours (by default). #123

... This is changing behavior, not reflected in release notes (e.g. users might need to change their tooling).

> New options for controlling suspect and reaping times.

Nice! As long as defaults didn't change, it's great.

> Add new `Ready` and `Destroyed` events to ringpop. #125

Another sentence on what you can do with them. E.g. "these are optional events that can invoke a user callback when ringpop is ready and destroyed, respectively". Otherwise, it's a compatible change.

> Fix bug where Ringpop automatically added itself to the bootstrap hosts for

Nice, bug fix! But wait if I relied on `self` being added, what do I do now? Should I remove `self`? What if I had the DiscoverProvider in 0.3, how should I change in 0.4? I know I implemented that feature, and could have documented it the way I see it.

The remaining changelog looks perfect.

CHANGES.md
```diff
@@ -1,18 +1,29 @@
...
 
 ### Release notes
```
thanodnl: Lets add a note to the release notes to advise against running this version with an older version for more than 24 hours. After this time tombstone messages might start gossiping which is incompatible with the previous version.

Motiejus: :+1:

CHANGES.md
dansimau: I think you're right. We could go into a lot more detail about every change, if we wanted. I think this turns the release into a bigger task.

Let me try to address each of your feedback items.

> This is changing behavior, not reflected in release notes (e.g. users might need to change their tooling).

Right. If the customer has created tooling that relies on faulty members being in the membership lists longer than 24 hours, then yes. Isn't that obvious enough from the one-line description, though?

The next line then suggests that the value can be tweaked.

> Another sentence on what you can do with them. E.g. "these are optional events that can invoke a user callback when ringpop is ready and destroyed, respectively"

I think "events" implies that. And at least, if you're a Ringpop user, you know what an event is and does. So I don't think any more information is necessary here.

Perhaps we could elaborate more (and give examples, etc) in help documentation.

> Nice, bug fix! But wait if I relied on self being added, what do I do now?
> Referenced PR (#120) references 3cc9e19, which has the change commented in more detail in CHANGES.md. Do you think the information is unnecessary?

Fair criticism; apologies that I removed that detail. I was trying to make the changelog easier to parse.

I think if people are relying on the behaviour of a bug, that is unfortunate. I don't know if the details of the bug should be in the changelog/release notes, though.

Motiejus: Looks like that in the end it's a matter of taste. As a consumer of a library, I prefer authors to be more verbose than scarce. Given that, the detail needs to be organized, otherwise it's useless too.

I`m not going to block this, since, like you showed, everything can be figured out quite easily. Now you know see my preferences. :-)

LGTM.

thanodnl:
LGTM

PR #130, opened 2016-03-24
Ws refactor calls disco provider wait for convergence

CorgiMan:
Some refactoring, mostly to make it easier to make calls. This is useful for the implementation of bidirectional full syncs and partition healing which will follow after this has landed.

1. remove unnecessary concurrency in waitForConvergence
2. add stats for tree construction time
3. make discover provider part of the node and no longer part of the join sender
4. add sendPingWithChanges and sendJoinRequest method
5. refactor MakeCall to easier to understand sendJoinRequest/sendPing
 (also adds timeout via ctx.Done to sendPingWithChanges)

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -543,11 +543,7 @@ func (rp *Ringpop) handleChanges(changes []swim.Change) {
 		}
 	}
 
-	start := time.Now()
 	rp.ring.AddRemoveServers(serversToAdd, serversToRemove)
-	treeConstructionTime := time.Now().Sub(start)
-	rp.statter.RecordTimer(rp.getStatKey("tree"), nil, treeConstructionTime)
-
 }
 
 //= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
--- a/ringpop.go
+++ b/ringpop.go
@@ -543,7 +543,11 @@ func (rp *Ringpop) handleChanges(changes []swim.Change) {
 		}
 	}
 
+	start := time.Now()
 	rp.ring.AddRemoveServers(serversToAdd, serversToRemove)
+	treeConstructionTime := time.Now().Sub(start)
+	rp.statter.RecordTimer(rp.getStatKey("tree"), nil, treeConstructionTime)
+
 }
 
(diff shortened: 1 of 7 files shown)


ringpop.go
```diff
@@ -531,7 +531,11 @@ func (rp *Ringpop) handleChanges(changes []swim.Change) {
 		}
 	}
 
+	start := time.Now()
 	rp.ring.AddRemoveServers(serversToAdd, serversToRemove)
+	treeConstructionTime := time.Now().Sub(start)
+	rp.statter.RecordTimer(rp.getStatKey("tree"), nil, treeConstructionTime)
```
Motiejus: `tree` as a stat key is a bit ambiguous. Maybe `tree.construction`?

swim/join_sender.go
```diff
@@ -373,49 +375,38 @@ func (j *joinSender) JoinGroup(nodesJoined []string) ([]string, []string) {
 	j.numTries++
 	j.node.emit(JoinTriesUpdateEvent{j.numTries})
 
-	for _, node := range group {
```
Motiejus: Great change!

At this point, you know really well what `JoinGroup` is doing. Can you take some time to document the function? Especially, the input and output arguments, and how it does error reporting.

swim/join_sender.go
```diff
@@ -477,7 +438,30 @@ func sendJoinRequest(node *Node, target string, timeout time.Duration) (*joinRes
 		Timeout:     timeout,
 	}
 	res := &joinResponse{}
-	err := json.CallPeer(ctx, peer, node.service, "/protocol/join", req, res)
```
Motiejus: Likewise, can you godoc this function too?

Motiejus:
Aside minor comments in the separete commits, looks really good!

ringpop.go
thanodnl: I think this is a `hashring` stat and not a ringpop stat so also add `ring` to the path as is done for `ring.server-added` `ring.server-removed` and `ring.changed` a couple of lines above.

Motiejus:
:+1: LGTM

thanodnl:
LGTM

PR #131, opened 2016-03-24
bi-directional full sync

CorgiMan:


swim/ping_handler.go
```diff
@@ -40,8 +42,11 @@ func handlePing(node *Node, req *ping) (*ping, error) {
 	changes, fullSync :=
 		node.disseminator.IssueAsReceiver(req.Source, req.SourceIncarnation, req.Checksum)
 
+	// Bidirectional full sync.
 	if fullSync {
-		// TODO: handle full sync
+		go func() {
```
jwolski: This seems wasteful to me. If the sender applies the changes from the receiver and their checksums match, a bidirectional full-sync wouldn't be needed. Now what we've got is bi-directional full-syncs by default.

Also, the unbounded go-routine creation frightens me a bit. I know they're meant to be lightweight, but not handling this by throttling the amount of bi-directional full-syncs that we issue seems like something that would come back to haunt us, at scale.

Motiejus: I don't think it will full-sync all the time: if checksums match, it will return false: https://github.com/uber/ringpop-go/blob/6e836b4043b3171b651a803024677cca38135898/swim/disseminator.go#L165

However, I agree with the second argument: we should throttle the number of goroutines we spawn here, possibly one.

swim/ping_request_handler.go
```diff
@@ -66,8 +66,11 @@ func handlePingRequest(node *Node, req *pingRequest) (*pingResponse, error) {
 	changes, fullSync :=
 		node.disseminator.IssueAsReceiver(req.Source, req.SourceIncarnation, req.Checksum)
 
+	// Bidirectional full sync.
 	if fullSync {
-		// TODO: something...
+		go func() {
```
Motiejus: I am a bit in doubt if we want to enable full syncs in ping-reqs too. Maybe we should limit ourselves to pings only first?

swim/ping_handler.go
CorgiMan: I think what Jeff means is that if the one way full sync fixes the membership, we don't need the reverse full sync

swim/ping_request_handler.go
CorgiMan: I like that idea

swim/ping_handler.go
CorgiMan: I don't think of this as wasteful. Full syncs are necessary because less wasteful methods: the normal dissemination protocol, has failed. I'm okay if the emergency mechanism is very wasteful, yet effective. As of now, it's half as wasteful and half as effective.

In some sense the reverse full syncs are already throttled by the average of receiving 5 pings per second. If we can't trust that this is always the case, I agree that we should throttle the reverse full syncs. One at the time seems appropriate to me.

PR #132, opened 2016-03-25
Feature/reaping faulty nodes

thanodnl:
This PR reverts the revert (revert-ception) of the tombstone changes.

To make sure that newer versions of `ringpop-go` do not crash older versions we gossip the tombstone state around as a `tombstone` flag on the change. This flag is only interpreted when the state is set to faulty, in all other cases it is ignored.

Besides gossiping the `tombstones` with `tombstone` flags on the `faulty` state ringpop also accepts tombstones to be gossiped as a state. This will make it possible for a future version to drop the gossipping of the tombstone flag in favor of the proper state.

Lastly this PR removes the panic in `statePrecedence` to prevent crashing systems. Now it will return `-1` for all the unknown states which would cause the states to not take precedence.


thanodnl:
Please ignore till #133 lands

severb:
Should we also update `func (rp *Ringpop) handleChanges(changes []swim.Change) {`?

thanodnl:
This PR is a combination of the tombstone feature that is reviewed in #123 and the protocol compatibility that is reviewed in #133.

The last change on this PR is with regards to the issue that @severb found in the implementation of `handleChanges` where directly going from `alive` or `suspect` to `tombstone` would not remove the member from the hashring. Although unlikely it is a good catch. The same was happening for a `faulty` to `suspect` but then the other way around. Fixed it in `ab1cbbe`.

Please ignore the CI run on the push, this run does not incorporate the mockery fix and therefor fails. The CI run for the pr includes the fix and represents the state of the world when we land this PR.

swim/member.go
```diff
@@ -123,11 +129,27 @@ type Change struct {
 	Address           string `json:"address"`
 	Incarnation       int64  `json:"incarnationNumber"`
 	Status            string `json:"status"`
+	Tombstone         bool   `json:"tombstone,omitempty"`
```
CorgiMan: Do we have a test for the omitempty case? (Unmarshal to map[string]interface{} and checking for the field would suffice)

CorgiMan: Do we have a test for the omitempty case? (Unmarshal to map[string]interface{} and checking for the field would suffice)

swim/state_transitions_test.go
```diff
@@ -102,19 +108,44 @@ func (s *StateTransitionsSuite) TestSuspectDisabled() {
...
+	s.m.MakeSuspect(s.suspect.Address, s.suspect.Incarnation)
 	member, _ := s.m.Member(s.suspect.Address)
 	s.Require().NotNil(member, "expected cannot be nil")
 
 	s.stateTransitions.ScheduleSuspectToFaulty(*member)
-	s.NotNil(s.stateTransitions.timer(member.Address), "expected suspicion timer to be set")
+	s.NotNil(s.stateTransitions.timer(member.Address), "expected state transtition timer to be set")
 
-	time.Sleep(5 * time.Millisecond)
+	s.clock.Add(5*time.Second + 1)
```
CorgiMan: I prefer to not use the automatic conversion from 1 to Nanosecond (or whatever duration it is)

swim/memberlist.go
```diff
@@ -131,6 +136,31 @@ func (m *memberlist) Member(address string) (*Member, bool) {
 	return member, ok
 }
 
+// RemoveMember removes the member from the membership list. If the membership has
+// changed during this operation a new checksum will be computed.
+func (m *memberlist) RemoveMember(address string) bool {
```
CorgiMan: Can we have a test specifically for this function?

swim/memberlist.go
```diff
@@ -352,10 +406,16 @@ func (m *memberlist) getJoinPosition() int {
 }
 
 // applies a change directly to the member list
-func (m *memberlist) Apply(change Change) {
+func (m *memberlist) Apply(change Change) bool {
```
CorgiMan: I think a comment on what this return value means is needed

CorgiMan: Oh I was a bit confused on how this function works. I thought apply checked if the change has precedence over the current member. Maybe we can add a comment about that too though

swim/state_transitions_test.go
severb: I also find this confusing. Is it the same as `6*time.Second`?

swim/state_transitions.go
```diff
@@ -46,30 +47,40 @@ type transitionTimer struct {
 type stateTransitions struct {
 	sync.Mutex
 
-	node *Node
-
+	node     *Node
+	clock    clock.Clock
```
CorgiMan: Can we use the Node's clock instead?

swim/state_transitions_test.go
CorgiMan: Do we actually need it?

swim/state_transitions_test.go
```diff
@@ -132,6 +163,24 @@ func (s *StateTransitionsSuite) TestTimerCreated() {
 	s.NotEqual(old, s.stateTransitions.timer(member.Address), "expected timer to change")
 }
 
+// TestTimerCreated tests that starting suspicion for a node creates a
```
CorgiMan: Comment has different fn name

CorgiMan:
Still needs a lot of comments (also on the tests please). I think most unexported functions can benefit from a little comment as well.

swim/state_transitions.go
severb: I don't think that's a good idea ([LoD](https://en.wikipedia.org/wiki/Law_of_Demeter)), even if I see that pattern used virtually everywhere in the codebase. It makes it hard to mock/test things and adds unnecessary coupling.

severb:
Apart from the existing comments, LGTM. Good job!

Some food for thought, would it be easy to add a middleware stack in ringpop-go as well?

swim/state_transitions.go
CorgiMan: In my mind, time is a global. Since we don't have control over the time package we need to store this "global" somewhere, which we did in the node. We never want the clocks on both instances to have different values. I think the coupling is very much needed, not unnecessary. Therefore, for me, it makes more sense to have one clock. Choose what you like though.

severb: I've read somewhere that everything is relative 😄

dansimau: Have we considered the edge case where one node could end up in a different spacetime continuum? Then we might need multiple clocks to track the different continuums (and keep the cluster stable).

This is a very real concern, considering we are all mocking time in our Amsterdam office and the Large Hadron Collider is only roughly 400 miles away. Just saying.

swim/member.go
thanodnl: I don't think we need to cover such cases where the standard library offers us this capability. There is no functional change if it happens to fail, it is just an optimization.

CorgiMan: I trust the stdlib, but I don't trust this line to never change. I think if we change anything in the most used endpoint, we should test it.

swim/state_transitions.go
CorgiMan: Are you making fun of the last seven years of my life Dan? OUch

swim/disseminator_test.go
```diff
@@ -139,6 +139,37 @@ func (s *DisseminatorTestSuite) TestIssueChangesAsSender() {
 	s.Len(changes, 3, "expected three changes to be issued")
 }
 
+func (s *DisseminatorTestSuite) TestIssueChangesAsSenderForTombstone() {
```
CorgiMan: Can you put a 1 to 2 line comment here please

swim/disseminator_test.go
```diff
@@ -139,6 +139,37 @@ func (s *DisseminatorTestSuite) TestIssueChangesAsSender() {
...
+	s.m.MakeAlive(tombstoneAddr, s.incarnation)
+	s.m.MakeTombstone(tombstoneAddr, s.incarnation)
+
+	changes, _ := s.d.IssueAsSender()
+	s.Require().Len(changes, 1, "required to only have the tombstone change")
+	s.Assert().Equal("faulty", changes[0].Status, "expected tombstone to be gossiped as a faulty state")
+	s.Assert().Equal(true, changes[0].Tombstone, "expected tombstone to be gossiped with a tombstone flag")
+}
+
+func (s *DisseminatorTestSuite) TestIssueChangesAsReceiverForTombstone() {
```
CorgiMan: comment please

swim/handlers_test.go
```diff
@@ -243,6 +243,25 @@ func (s *HandlerTestSuite) TestPingRequestHandlerFail() {
 	s.Nil(res)
 }
 
+func (s *HandlerTestSuite) TestAdminReadHandler() {
```
CorgiMan: comment please

swim/memberlist.go
```diff
@@ -225,6 +255,21 @@ func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
 	return m.MakeChange(address, incarnation, Leave)
 }
 
+func (m *memberlist) MakeTombstone(address string, incarnation int64) []Change {
```
CorgiMan: Comment please

swim/memberlist_test.go
```diff
@@ -278,6 +313,26 @@ func (s *MemberlistTestSuite) TestCountReachableMembers() {
...
 
+func (s *MemberlistTestSuite) TestRemoveMember() {
+	// seed the membership with more members
+	s.m.Update(s.changes)
+
+	var removed bool
+
+	// remove an unknown member
+	removed = s.m.RemoveMember("192.0.2.123:1234")
+	s.Assert().False(removed, "expect to not remove an unknown member")
```
CorgiMan: It seems like you test the return value, but not that no member is removed

swim/memberlist_test.go
```diff
@@ -278,6 +313,26 @@ func (s *MemberlistTestSuite) TestCountReachableMembers() {
...
+
+	// remove a known member
+	removed = s.m.RemoveMember(s.changes[0].Address)
+	s.Assert().True(removed, "expect to remove a member that was added before")
```
CorgiMan: It seems like you test the return value, but not that the member is removed

swim/memberlist_test.go
thanodnl: `RemoveMember` returns a boolean if the member was in there before, thus if the removed value is false nothing has been removed.

swim/handlers_test.go
CorgiMan: Nvmd

swim/memberlist_test.go
thanodnl: I added tests for counting the number of members below.

swim/member.go
thanodnl: a test has been added: https://github.com/uber/ringpop-go/blob/feature/reaping-faulty-nodes/swim/member_test.go#L126

swim/memberlist.go
```diff
@@ -225,6 +255,27 @@ func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
...
+// is already higher than the incartation number provided in this function it is
+// essentially a no-op. The list of changes that is returned is the actual list of
+// changes that have been applied to the memberlist. It can be used to test if the
+// tombstone declaration has been executed atleast to the local memberlist.
+func (m *memberlist) MakeTombstone(address string, incarnation int64) []Change {
+	m.node.emit(MakeNodeStatusEvent{Tombstone})
+	return m.MakeChange(address, incarnation, Tombstone)
+}
+
+func (m *memberlist) Evict(address string) {
```
CorgiMan: I was wondering why golint didn't catch this because it doesn't have a docstring. I think because memberlist is unexported it doesn't need one. I am still for having every function in this file commented, maybe we can add this task in the backlog. (No action required, but if you feel like it, please)

CorgiMan:
Looks good!

PR #133, opened 2016-03-29
Feature/protocol compatible tombstones

thanodnl:
**IMPORTANT:** This diff is against a feature branch who's sole purpose it is to revert the revert of the reaping-faulty-nodes feature. When this lands an other PR will be opened as a formality to get all changes to `dev`.

These changes implement a new protocol for gossipping tombstones that is compatible with previous versions on top of the changes revert the revert of the reaping faulty nodes feature.

Main things to note here is that tombstones are gossiped around as `faulty` state with a `tombstone` flag on the change. This flag is ignored by older versions of ringpop and treated as the faulty state. This make sure that the `panic` condition is not triggered while gossiping tombstones during a hybrid deployment of the old and the new version.

Next there is an admin endpoint that can be used to manually initiate the reaping of faulty nodes. This will also be implemented in `ringpop-admin` for users to call after a successful upgrade to validate that the reaping works without waiting the _faulty period_.

Both changes have accompanying integration tests added to the `reaping-faulty-nodes` feature. This will be merged to the master branch of `ringpop-common` right before these changes land on `dev`.

When older versions of ringpop participate in the membership of tombstone aware nodes these changes will prevent the older versions from crashing on a panic parsing the tombstone state. Instead the tombstone flag will be ignored. This causes checksum incompatibilities during the _tombstone period_ between the older versions and the newer version. This has a side effect of elevated full syncs happening during this period which in the end causes removed members to reappear and be redeclared a tombstone after the _faulty period_. Only when all running instances run a tombstone aware version the  faulty's can be reaped effectively.

At the same time the panic statement that caused crashes is removed from the precedence calculation of states. This prevents hard crashes in the future.


thanodnl:
@Motiejus @severb Can you have a look at this PR? It is the PR discussed in the Standup that adds wire-protocol compatibility with older ringpop's for the tombstone changes.

swim/member.go
```diff
@@ -128,11 +129,27 @@ type Change struct {
 	Address           string `json:"address"`
 	Incarnation       int64  `json:"incarnationNumber"`
 	Status            string `json:"status"`
+	Tombstone         bool   `json:"tombstone,omitempty"`
 	// Use util.Timestamp for bi-direction binding to time encoded as
 	// integer Unix timestamp in JSON
 	Timestamp util.Timestamp `json:"timestamp"`
 }
 
+func (c Change) validateIncomming() Change {
```
Motiejus: s/Incomming/Incoming/

swim/memberlist.go
```diff
@@ -304,6 +304,11 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 		return nil
 	}
 
+	// validate incomming changes
```
Motiejus: s/incomming/incoming

Motiejus:
Sorry for taking that long to look at this.

Given our current situation in terms of compatibility, I really like your approach.

Motiejus:
LGTM

test/run-integration-tests
```diff
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="feature/admin-reap"
```
severb: I don't think we should push this change to ringpop-go master.

severb:
LGTM

test/run-integration-tests
thanodnl: I completely agree. That branch needs to land to master in `ringpop-common` before we merge this branch.

severb:
Awesome job! 🎩

dansimau:
winner

PR #134, opened 2016-03-29
Ws bidirectional full sync

CorgiMan:
We throttle the amount of running bidirectional syncers to a configurable amount (default 5).

### DIFF
--- a/swim/disseminator.go
+++ b/swim/disseminator.go
@@ -23,6 +23,7 @@ package swim
 import (
 	"math"
 	"sync"
+	"time"
 
 	log "github.com/uber-common/bark"
 	"github.com/uber/ringpop-go/logging"
@@ -53,16 +54,19 @@ type disseminator struct {
 	sync.RWMutex
 
 	logger log.Logger
+
+	reverseFullSyncJobs chan struct{}
 }
 
 // newDisseminator returns a new Disseminator instance.
 func newDisseminator(n *Node) *disseminator {
 	d := &disseminator{
-		node:    n,
-		changes: make(map[string]*pChange),
-		maxP:    defaultPFactor,
-		pFactor: defaultPFactor,
-		logger:  logging.Logger("disseminator").WithField("local", n.Address()),
+		node:                n,
+		changes:             make(map[string]*pChange),
+		maxP:                defaultPFactor,
+		pFactor:             defaultPFactor,
+		logger:              logging.Logger("disseminator").WithField("local", n.Address()),
+		reverseFullSyncJobs: make(chan struct{}, n.maxReverseFullSyncJobs),
 	}
 
 	return d
@@ -244,4 +248,57 @@ func (d *disseminator) ChangesCount() int {
 	c := len(d.changes)
 	d.RUnlock()
 	return c
+}
+
+// tryStartReverseFullSync fires a goroutine that performs a full sync. We omit
+// the reverse full sync if the maximum number of processes are already
+// running. This ensures no more than reverseFullSyncJobs processes are
+// running concurrently.
+func (d *disseminator) tryStartReverseFullSync(target string, timeout time.Duration) {
+	// occupy a job, return if none are available
+	select {
+	case d.reverseFullSyncJobs <- struct{}{}:
+		// continue if job is available
+	default:
+		d.logger.WithFields(log.Fields{
+			"remote": target,
+		}).Info("omit bidirectional full sync, too many already running")
+
+		d.node.emit(OmitReverseFullSyncEvent{Target: target})
+		return
+	}
+
+	// start reverse full sync
+	go func() {
+		d.reverseFullSync(target, timeout)
+
+		// create a new vacancy when the job is done
+		<-d.reverseFullSyncJobs
+	}()
+}
+
+// reverseFullSync is the second part of a bidirectional full sync. The first
+// part is performed by the IssueAsReceiver method. The reverse full sync
+// ensures that this node merges the membership of the target node's membership
+// with its own.
+func (d *disseminator) reverseFullSync(target string, timeout time.Duration) {
+	d.node.emit(StartReverseFullSyncEvent{Target: target})
+
+	res, err := sendJoinRequest(d.node, target, timeout)
+	if err != nil || res == nil {
+		d.logger.WithFields(log.Fields{
+			"remote": target,
+			"error":  err,
+		}).Warn("bidirectional full sync failed due to failed join request")
+		return
+	}
+
+	d.logger.WithFields(log.Fields{
+		"remote": target,
+	}).Info("bidirectional full sync")
+
+	cs := d.node.memberlist.Update(res.Membership)
+	if len(cs) == 0 {
+		d.node.emit(RedundantReverseFullSyncEvent{Target: target})
+	}
 }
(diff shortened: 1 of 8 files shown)


swim/disseminator.go
```diff
@@ -245,3 +248,57 @@ func (d *disseminator) ChangesCount() int {
 	d.RUnlock()
 	return c
 }
+
+// tryStartReverseFullSync fires a goroutine that performs a full sync, with a
+// maximum of five concurrent goroutines at the same time. If five are already
```
CorgiMan: don't use five because it's configurable

Motiejus: s/five/maxReverseFullSyncJobs

swim/disseminator_test.go
```diff
@@ -356,3 +360,143 @@ func contains(cs []Change, address string) bool {
...
+	bootstrapNodes(t, b, a)
+
+	// clear changes so that a full sync will be issued
+	a.node.disseminator.ClearChanges()
+	b.node.disseminator.ClearChanges()
+
+	// check that indeed b's memberlist doesn't contain a
+	addrA := a.node.Address()
+	_, has := b.node.memberlist.Member(addrA)
+	assert.False(t, has, "expected that node1 is not yet part of a's membership")
```
CorgiMan: s/node1/a/

swim/disseminator.go
```diff
@@ -245,3 +248,57 @@ func (d *disseminator) ChangesCount() int {
...
+	// start reverse full sync
+	go func() {
+		d.reverseFullSync(target, timeout)
+
+		// create a new vacancy when the job is done
+		<-d.reverseFullSyncJobs
+	}()
+}
+
+// reverseFullSync is the second part of a bidirectional full sync. It ensures
```
Motiejus: Where is the first part? If you are documenting the function, add context either to the header of the file, or somewhere close.

swim/disseminator.go
```diff
@@ -245,3 +248,57 @@ func (d *disseminator) ChangesCount() int {
...
+// that this node merges the membership of the target node with its own.
+func (d *disseminator) reverseFullSync(target string, timeout time.Duration) {
+	d.node.emit(StartReverseFullSyncEvent{Target: target})
+
+	res, err := sendJoinRequest(d.node, target, timeout)
+	if err != nil || res == nil {
+		d.logger.WithFields(log.Fields{
+			"remote": target,
+			"error":  err,
+		}).Warn("bidirectional full sync failed")
```
Motiejus: "bidirectional full sync failed" is a bit too generic, since actually just the join request is failing. You should either say "second part of", or "join part of".

swim/disseminator_test.go
```diff
@@ -356,3 +360,143 @@ func contains(cs []Change, address string) bool {
...
+	return ListenerFunc(func(e events.Event) {
+		switch e.(type) {
+		case MemberlistChangesAppliedEvent:
+			f()
+		}
+	})
+}
+
+// TestThrottleBidirectionalFullSyncs tests if we throttle the maximum number
+// of reverse full syncs running at the same time. In this test, we start six
```
Motiejus: s/six/maxReverseFullSyncJobs+1/

swim/disseminator.go
CorgiMan: The first part is what we call a full sync now. I'll change the comment to make that clear

Motiejus:
Except for nitpicks in comments, awesome work!

swim/disseminator_test.go
CorgiMan: :+1:

Motiejus:
:+1:

swim/disseminator.go
```diff
@@ -245,3 +248,60 @@ func (d *disseminator) ChangesCount() int {
...
 }
+
+// tryStartReverseFullSync fires a goroutine that performs a full sync. We omit
+// the reverse full sync if the maximum number of processes are already
+// running. This ensures no more than reverseFullSyncJobs processes are
+// running concurrently.
+func (d *disseminator) tryStartReverseFullSync(target string, timeout time.Duration) {
+	if d.reverseFullSyncJobs == nil {
+		d.reverseFullSyncJobs = make(chan struct{}, d.node.maxReverseFullSyncJobs)
+	}
```
severb: Should this be in `newDisseminator`?

swim/disseminator.go
```diff
@@ -245,3 +248,60 @@ func (d *disseminator) ChangesCount() int {
...
+		return
+	}
+
+	// start reverse full sync
+	go func() {
+		d.reverseFullSync(target, timeout)
+
+		// create a new vacancy when the job is done
+		<-d.reverseFullSyncJobs
+	}()
```
severb: _I think_ moving this in `case` would be more readable, but I'm fine either way.

swim/disseminator_test.go
```diff
@@ -356,3 +360,143 @@ func contains(cs []Change, address string) bool {
...
+	tnode.node.RegisterListener(onStartReverseFullSync(func() {
+		atomic.AddInt64(&total, 1)
+		<-block
+	}))
+
+	// Listen for a reverse full sync that is omitted because the max number of
+	// reverse full sync routines are running.
+	tnode.node.RegisterListener(onOmitReverseFullSync(func() {
+		go func() {
+			for i := 0; i < maxJobs; i++ {
```
severb: Can we use `total` insteaf of `maxJobs`? In case of regressions, it wouldn't deadlock the test.

severb:
Looks good overall. The only thing I'd say is that we are a bit light on tests. The ones that we have are complex in setup and too broad (i.e. an entire bootstrap) to test the change. They feel more like end-to-end tests, which is OK, but I'd like to see, if possible, some simpler and to the point unit-tests.

severb:
In any case, we should wait for the integration tests to pass again before landing this.

swim/disseminator.go
CorgiMan: Yeah I wasn't sure, I liked that all the code was in the same place, but I think you're right.

swim/disseminator.go
CorgiMan: In this case, since there is little code, I agree. However, I've seen the situations where the case becomes very large and the code becomes impossible to follow. The lack of indentation in the form above is also a plus.

I've found that reducing the select statement into it's smallest form, is generally the easiest way to follow concurrent code.

swim/disseminator_test.go
CorgiMan: There is a timeout on the test so it won't deadlock.   want to cleanup the blocked goroutines though

CorgiMan:
Here is a pr that let's the integration tests pass https://github.com/uber/ringpop-common/pull/39

And you're right, we could use a more fine-grained test.

swim/disseminator_test.go
severb: My concern is that if we have a off-by-one error, and we omit earlier than expected, `onOmitReverseFullSync` will block when trying to free an additional `onStartReverseFullSync`. Because `onOmitReverseFullSync` is not called in a goroutine (neither by the event system or by the test code), it can potentially run before the last `select` in the test, deadlocking the test. I think we should either use `total` or, even better, add a buffer to the `block` channel equal to `maxJobs`.

severb:
LGT:dollar:

severb:
Good stuff, LGTM.

PR #135, opened 2016-04-07
Partition healing

CorgiMan:
This PR includes an implementation of the partition healing algorithm.

When the network is partitioned and a collection of nodes A is unable to communicate with another collection of nodes B the cluster becomes in a partitioned state. A declares nodes from B faulty and B declares nodes from A faulty. If the network partition is becomes fixed, there is still no communication between A and B because faulty nodes shouldn't be pinged according to the swim protocol.

The partition healing algorithm tries to fix this state where the network is restored but the cluster is in the partitioned state described above.

A brief description of the algorithm:
- A node from A contacts a node from B and compares their memberships.
- If there are members that would cause faulty declarations when merging the two memberships according to the swim rules, the node reincarnates those members (either by declaring them suspect to its own membership, or declaring them suspect by piggybacking this over a ping to a node from B) . 
- If there are members that need to reincarnate, the node merges the partitions by applying the membership of B to it's own membership and sending its membership piggybacked over a ping to a node of B.

Please grab me to discuss, this is quite a big one.

### DIFF
--- a/swim/disseminator_test.go
+++ b/swim/disseminator_test.go
@@ -390,7 +390,7 @@ func TestBidirectionalFullSync(t *testing.T) {
 
 	// only when a reverse full sync is issued the membership of b changes
 	// we listen for a membership change and continue the main thread of the test
-	b.node.RegisterListener(on(MemberlistChangesAppliedEvent{}, func() {
+	b.node.RegisterListener(on(MemberlistChangesAppliedEvent{}, func(e Event) {
 		cont <- struct{}{}
 	}))
 
@@ -431,14 +431,14 @@ func TestThrottleBidirectionalFullSyncs(t *testing.T) {
 	total := int64(0)
 
 	// Block the reverse full sync procedure to test if we throttle correctly.
-	tnode.node.RegisterListener(on(StartReverseFullSyncEvent{}, func() {
+	tnode.node.RegisterListener(on(StartReverseFullSyncEvent{}, func(e Event) {
 		atomic.AddInt64(&total, 1)
 		<-block
 	}))
 
 	// Listen for a reverse full sync that is omitted because the max number of
 	// reverse full sync routines are running.
-	tnode.node.RegisterListener(on(OmitReverseFullSyncEvent{}, func() {
+	tnode.node.RegisterListener(on(OmitReverseFullSyncEvent{}, func(e Event) {
 		go func() {
 			for i := 0; i < maxJobs; i++ {
 				block <- struct{}{}
@@ -493,7 +493,7 @@ func TestRedundantFullSync(t *testing.T) {
 	waitForConvergence(t, time.Second, tnodes...)
 
 	quit := make(chan struct{})
-	tnodes[0].node.RegisterListener(on(RedundantReverseFullSyncEvent{}, func() {
+	tnodes[0].node.RegisterListener(on(RedundantReverseFullSyncEvent{}, func(e Event) {
 		close(quit)
 	}))
 
--- a/swim/disseminator_test.go
+++ b/swim/disseminator_test.go
@@ -373,8 +373,8 @@ func TestBidirectionalFullSync(t *testing.T) {
 	b := newChannelNode(t)
 	defer b.Destroy()
 
-	// after bootstrap a's memberlist contains a and b,
-	// but b's membership only contains b.
+	// After bootstrap, a's memberlist contains a and b, but b's membership
+	// only contains b.
 	bootstrapNodes(t, b, a)
 
 	// clear changes so that a full sync will be issued
--- a/swim/disseminator_test.go
+++ b/swim/disseminator_test.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
+	"github.com/uber/ringpop-go/events"
 	"github.com/uber/ringpop-go/util"
 )
 
@@ -112,7 +113,7 @@ func (s *DisseminatorTestSuite) TestChangesAreCleared() {
 	s.Equal(0, s.d.ChangesCount(), "expected no problems deleting non-existent changes")
 }
 
-func (s *DisseminatorTestSuite) TestFullSync() {
+func (s *DisseminatorTestSuite) TestMembershipAsChanges() {
 	addresses := fakeHostPorts(1, 1, 2, 4)
 
 	for _, address := range addresses {
@@ -390,7 +391,7 @@ func TestBidirectionalFullSync(t *testing.T) {
 
 	// only when a reverse full sync is issued the membership of b changes
 	// we listen for a membership change and continue the main thread of the test
-	b.node.RegisterListener(on(MemberlistChangesAppliedEvent{}, func(e Event) {
+	b.node.RegisterListener(on(MemberlistChangesAppliedEvent{}, func(e events.Event) {
 		cont <- struct{}{}
 	}))
 
@@ -431,14 +432,14 @@ func TestThrottleBidirectionalFullSyncs(t *testing.T) {
 	total := int64(0)
 
 	// Block the reverse full sync procedure to test if we throttle correctly.
-	tnode.node.RegisterListener(on(StartReverseFullSyncEvent{}, func(e Event) {
+	tnode.node.RegisterListener(on(StartReverseFullSyncEvent{}, func(e events.Event) {
 		atomic.AddInt64(&total, 1)
 		<-block
 	}))
 
 	// Listen for a reverse full sync that is omitted because the max number of
 	// reverse full sync routines are running.
-	tnode.node.RegisterListener(on(OmitReverseFullSyncEvent{}, func(e Event) {
+	tnode.node.RegisterListener(on(OmitReverseFullSyncEvent{}, func(e events.Event) {
 		go func() {
 			for i := 0; i < maxJobs; i++ {
 				block <- struct{}{}
@@ -493,7 +494,7 @@ func TestRedundantFullSync(t *testing.T) {
 	waitForConvergence(t, time.Second, tnodes...)
 
 	quit := make(chan struct{})
-	tnodes[0].node.RegisterListener(on(RedundantReverseFullSyncEvent{}, func(e Event) {
+	tnodes[0].node.RegisterListener(on(RedundantReverseFullSyncEvent{}, func(e events.Event) {
 		close(quit)
 	}))
 
--- a/swim/disseminator_test.go
+++ b/swim/disseminator_test.go
@@ -119,7 +119,7 @@ func (s *DisseminatorTestSuite) TestFullSync() {
 		s.m.MakeAlive(address, s.incarnation)
 	}
 
-	changes := s.d.FullSync()
+	changes := s.d.MembershipAsChanges()
 
 	s.Len(changes, 4, "expected to get change for each member")
 }
@@ -508,7 +508,7 @@ func TestRedundantFullSync(t *testing.T) {
 }
 
 // TestReverseFullSyncJoinFailure test the code path for when the join in the
-// reverseFullSync fails. If the app doesn't crash, the test has passed.
+// reverseFullSync fails.
 func TestReverseFullSyncJoinFailure(t *testing.T) {
 	tnodes := genChannelNodes(t, 5)
 	bootstrapNodes(t, tnodes...)
--- a/swim/events.go
+++ b/swim/events.go
@@ -244,33 +244,3 @@ func on(t interface{}, f func(e events.Event)) ListenerFunc {
 		}
 	})
 }
-
-// ExecuteThenWaitFor executes a function and then waits for a specific type
-// of event to occur.
-//
-// Often we want to execute some code and then wait for an event be emitted due
-// to the code being executed. However in order to not miss the event we must
-// first register an event handler before we can execute the code:
-// - register listener that signals we can continue on receiving the correct event;
-// - execute the code that will lead to an event being emitted;
-// - wait for the continue signal.
-//
-// This can be quite hard to follow. Ideally we want it to look like
-// - execute the code that will lead to an event being emitted;
-// - and then wait for a specific event.
-//
-// This function helps with making the code read like the latter.
-func ExecuteThenWaitFor(f func(), er events.EventRegistrar, t interface{}) {
-	block := make(chan struct{})
-
-	er.RegisterListener(on(t, func(e events.Event) {
-		select {
-		case <-block: // channel is closed, do nothing
-		default:
-			close(block)
-		}
-	}))
-
-	f()
-	<-block
-}
--- a/swim/events.go
+++ b/swim/events.go
@@ -234,10 +234,49 @@ type RedundantReverseFullSyncEvent struct {
 	Target string `json:"target"`
 }
 
+// AttemptHealEvent is sent when the discover provider healer attempts to heal a partition
+type DiscoHealEvent struct{}
+
+// AttemptHealEvent is sent when the healer is triggered
+type AttemptHealEvent struct{}
+
+// EventRegistrar is an object that you can register EventListeners on.
+type EventRegistrar interface {
+	RegisterListener(EventListener)
+}
+
+// on is returns an EventListener that executes a function f upon receiving an
+// event of type t.
 func on(t interface{}, f func()) ListenerFunc {
 	return ListenerFunc(func(e events.Event) {
 		if reflect.TypeOf(t) == reflect.TypeOf(e) {
 			f()
 		}
 	})
+}
+
+// ExecuteThenWaitFor executes a function and then waits for a specific type
+// of event to occur.
+//
+// Often we want to execute some code and then wait for an event be emitted due
+// to the code being executed. However in order to not miss the event we must
+// first register an event handler before we can execute the code:
+// - register listener that signals we can continue on receiving the correct event;
+// - execute the code that will lead to an event being emitted;
+// - wait for the continue signal.
+//
+// This can be quite hard to follow. Ideally we want it to look like
+// - execute the code that will lead to an event being emitted;
+// - and then wait for a specific event.
+//
+// This function helps with making the code read like the latter.
+func ExecuteThenWaitFor(f func(), er EventRegistrar, t interface{}) {
+	block := make(chan struct{})
+
+	er.RegisterListener(on(t, func() {
+		close(block)
+	}))
+
+	f()
+	<-block
 }
--- a/swim/events.go
+++ b/swim/events.go
@@ -28,6 +28,8 @@ import (
 	"github.com/uber/ringpop-go/events"
 )
 
+type Event interface{}
+
 // An EventListener handles events given to it by the SWIM node. HandleEvent should be thread safe.
 type EventListener interface {
 	HandleEvent(events.Event)
@@ -247,10 +249,10 @@ type EventRegistrar interface {
 
 // on is returns an EventListener that executes a function f upon receiving an
 // event of type t.
-func on(t interface{}, f func()) ListenerFunc {
+func on(t interface{}, f func(e Event)) ListenerFunc {
 	return ListenerFunc(func(e events.Event) {
 		if reflect.TypeOf(t) == reflect.TypeOf(e) {
-			f()
+			f(e)
 		}
 	})
 }
@@ -273,7 +275,7 @@ func on(t interface{}, f func()) ListenerFunc {
 func ExecuteThenWaitFor(f func(), er EventRegistrar, t interface{}) {
 	block := make(chan struct{})
 
-	er.RegisterListener(on(t, func() {
+	er.RegisterListener(on(t, func(e Event) {
 		close(block)
 	}))
 
--- a/swim/events.go
+++ b/swim/events.go
@@ -28,6 +28,7 @@ import (
 	"github.com/uber/ringpop-go/events"
 )
 
+// Event interface
 type Event interface{}
 
 // An EventListener handles events given to it by the SWIM node. HandleEvent should be thread safe.
--- a/swim/events.go
+++ b/swim/events.go
@@ -28,14 +28,6 @@ import (
 	"github.com/uber/ringpop-go/events"
 )
 
-// Event interface
-type Event interface{}
-
-// An EventListener handles events given to it by the SWIM node. HandleEvent should be thread safe.
-type EventListener interface {
-	HandleEvent(events.Event)
-}
-
 // The ListenerFunc type is an adapter to allow the use of ordinary functions
 // as EventListeners.
 type ListenerFunc func(events.Event)
@@ -243,14 +235,9 @@ type DiscoHealEvent struct{}
 // AttemptHealEvent is sent when the healer is triggered
 type AttemptHealEvent struct{}
 
-// EventRegistrar is an object that you can register EventListeners on.
-type EventRegistrar interface {
-	RegisterListener(EventListener)
-}
-
 // on is returns an EventListener that executes a function f upon receiving an
 // event of type t.
-func on(t interface{}, f func(e Event)) ListenerFunc {
+func on(t interface{}, f func(e events.Event)) ListenerFunc {
 	return ListenerFunc(func(e events.Event) {
 		if reflect.TypeOf(t) == reflect.TypeOf(e) {
 			f(e)
@@ -273,11 +260,15 @@ func on(t interface{}, f func(e Event)) ListenerFunc {
 // - and then wait for a specific event.
 //
 // This function helps with making the code read like the latter.
-func ExecuteThenWaitFor(f func(), er EventRegistrar, t interface{}) {
+func ExecuteThenWaitFor(f func(), er events.EventRegistrar, t interface{}) {
 	block := make(chan struct{})
 
-	er.RegisterListener(on(t, func(e Event) {
-		close(block)
+	er.RegisterListener(on(t, func(e events.Event) {
+		select {
+		case <-block: // channel is closed, do nothing
+		default:
+			close(block)
+		}
 	}))
 
 	f()
--- a/swim/events.go
+++ b/swim/events.go
@@ -234,7 +234,7 @@ type RedundantReverseFullSyncEvent struct {
 	Target string `json:"target"`
 }
 
-// AttemptHealEvent is sent when the discover provider healer attempts to heal a partition
+// DiscoHealEvent is sent when the discover provider healer attempts to heal a partition
 type DiscoHealEvent struct{}
 
 // AttemptHealEvent is sent when the healer is triggered
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -51,6 +51,7 @@ type Status struct {
 // HealResponse contains a list of nodes where healing was attempted
 type HealResponse struct {
 	Targets []string `json:"targets"`
+	Error   error    `json:"error"`
 }
 
 // notImplementedHandler is a dummy handler that returns an error explaining
@@ -122,7 +123,8 @@ func (n *Node) gossipHandlerStop(ctx json.Context, req *emptyArg) (*emptyArg, er
 }
 
 func (n *Node) discoverProviderHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
-	return &HealResponse{n.healer.Heal()}, nil
+	targets, err := n.healer.Heal()
+	return &HealResponse{Targets: targets, Error: err}, nil
 }
 
 func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -51,7 +51,7 @@ type Status struct {
 // HealResponse contains a list of nodes where healing was attempted
 type HealResponse struct {
 	Targets []string `json:"targets"`
-	Error   error    `json:"error"`
+	Error   string   `json:"error"`
 }
 
 // notImplementedHandler is a dummy handler that returns an error explaining
@@ -124,7 +124,7 @@ func (n *Node) gossipHandlerStop(ctx json.Context, req *emptyArg) (*emptyArg, er
 
 func (n *Node) discoverProviderHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
 	targets, err := n.healer.Heal()
-	return &HealResponse{Targets: targets, Error: err}, nil
+	return &HealResponse{Targets: targets, Error: err.Error()}, nil
 }
 
 func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -48,7 +48,7 @@ type Status struct {
 	Status string `json:"status"`
 }
 
-// HealStatus contains a list of nodes where healing was attempted
+// HealResponse contains a list of nodes where healing was attempted
 type HealResponse struct {
 	Targets []string `json:"targets"`
 }
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -69,7 +69,7 @@ func (n *Node) registerHandlers() error {
 		"/admin/gossip":              n.gossipHandler, // Deprecated
 		"/admin/gossip/start":        n.gossipHandlerStart,
 		"/admin/gossip/stop":         n.gossipHandlerStop,
-		"/admin/healpartition/disco": n.partitionHealerHandler,
+		"/admin/healpartition/disco": n.discoverProviderHealerHandler,
 		"/admin/tick":                n.tickHandler, // Deprecated
 		"/admin/gossip/tick":         n.tickHandler,
 		"/admin/member/leave":        n.adminLeaveHandler,
@@ -121,7 +121,7 @@ func (n *Node) gossipHandlerStop(ctx json.Context, req *emptyArg) (*emptyArg, er
 	return &emptyArg{}, nil
 }
 
-func (n *Node) partitionHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
+func (n *Node) discoverProviderHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
 	return &HealResponse{n.healer.Heal()}, nil
 }
 
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -124,7 +124,11 @@ func (n *Node) gossipHandlerStop(ctx json.Context, req *emptyArg) (*emptyArg, er
 
 func (n *Node) discoverProviderHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
 	targets, err := n.healer.Heal()
-	return &HealResponse{Targets: targets, Error: err.Error()}, nil
+	msg := ""
+	if err != nil {
+		msg = err.Error()
+	}
+	return &HealResponse{Targets: targets, Error: msg}, nil
 }
 
 func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
--- a/swim/handlers.go
+++ b/swim/handlers.go
@@ -48,6 +48,11 @@ type Status struct {
 	Status string `json:"status"`
 }
 
+// HealStatus contains a list of nodes where healing was attempted
+type HealResponse struct {
+	Targets []string `json:"targets"`
+}
+
 // notImplementedHandler is a dummy handler that returns an error explaining
 // this method is not implemented.
 func notImplementedHandler(ctx json.Context, req *emptyArg) (*emptyArg, error) {
@@ -56,18 +61,19 @@ func notImplementedHandler(ctx json.Context, req *emptyArg) (*emptyArg, error) {
 
 func (n *Node) registerHandlers() error {
 	handlers := map[string]interface{}{
-		"/protocol/join":      n.joinHandler,
-		"/protocol/ping":      n.pingHandler,
-		"/protocol/ping-req":  n.pingRequestHandler,
-		"/admin/debugSet":     notImplementedHandler,
-		"/admin/debugClear":   notImplementedHandler,
-		"/admin/gossip":       n.gossipHandler, // Deprecated
-		"/admin/gossip/start": n.gossipHandlerStart,
-		"/admin/gossip/stop":  n.gossipHandlerStop,
-		"/admin/tick":         n.tickHandler, // Deprecated
-		"/admin/gossip/tick":  n.tickHandler,
-		"/admin/member/leave": n.adminLeaveHandler,
-		"/admin/member/join":  n.adminJoinHandler,
+		"/protocol/join":             n.joinHandler,
+		"/protocol/ping":             n.pingHandler,
+		"/protocol/ping-req":         n.pingRequestHandler,
+		"/admin/debugSet":            notImplementedHandler,
+		"/admin/debugClear":          notImplementedHandler,
+		"/admin/gossip":              n.gossipHandler, // Deprecated
+		"/admin/gossip/start":        n.gossipHandlerStart,
+		"/admin/gossip/stop":         n.gossipHandlerStop,
+		"/admin/healpartition/disco": n.partitionHealerHandler,
+		"/admin/tick":                n.tickHandler, // Deprecated
+		"/admin/gossip/tick":         n.tickHandler,
+		"/admin/member/leave":        n.adminLeaveHandler,
+		"/admin/member/join":         n.adminJoinHandler,
 	}
 
 	return json.Register(n.channel, handlers, n.errorHandler)
@@ -113,6 +119,10 @@ func (n *Node) gossipHandlerStart(ctx json.Context, req *emptyArg) (*emptyArg, e
 func (n *Node) gossipHandlerStop(ctx json.Context, req *emptyArg) (*emptyArg, error) {
 	n.gossip.Stop()
 	return &emptyArg{}, nil
+}
+
+func (n *Node) partitionHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
+	return &HealResponse{n.healer.Heal()}, nil
 }
 
 func (n *Node) tickHandler(ctx json.Context, req *emptyArg) (*ping, error) {
--- a/swim/handlers_test.go
+++ b/swim/handlers_test.go
@@ -93,10 +93,20 @@ func (s *HandlerTestSuite) TestGossipStopHandler() {
 }
 
 func (s *HandlerTestSuite) TestPartitionHealerHandler() {
-	ExecuteThenWaitFor(func() {
-		_, err := s.testNode.node.partitionHealerHandler(s.ctx, &emptyArg{})
-		s.NoError(err, "calling handler should not result in error")
-	}, s.testNode.node, DiscoHealEvent{})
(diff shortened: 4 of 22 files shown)


swim/heal_partition.go
```diff
@@ -0,0 +1,111 @@
...
+	node.logger.WithField("target", target).Info("attempt heal")
+
+	// If join request succeeds a partition is detected,
+	// this node will now coordinate the healing mechanism.
+	joinRes, err := sendJoinRequest(node, target, time.Second)
+	if err != nil {
+		return err
+	}
+
+	A := node.disseminator.FullSync()
```
Motiejus: Since we are relying on the local membership information outside of full-sync, we should rename this method to something like `getLocalMembership`?

swim/heal_partition.go
```diff
@@ -0,0 +1,111 @@
...
+	B := joinRes.Membership
+	var csForA, csForB []Change
+
+	// Find changes that are alive for B and faulty for A and visa versa.
+	for _, m2 := range B {
+		m1, ok := selectMember(A, m2.Address)
+		if !ok {
+			continue
+		}
```
Motiejus: Can this be something like this (save 1 line)?

    if m1, ok := selectMember(A, m2.address); ! ok {
        continue
    }

swim/heal_partition.go
```diff
@@ -0,0 +1,111 @@
...
+	joinRes, err := sendJoinRequest(node, target, time.Second)
+	if err != nil {
+		return err
+	}
+
+	A := node.disseminator.FullSync()
+	B := joinRes.Membership
+	var csForA, csForB []Change
+
+	// Find changes that are alive for B and faulty for A and visa versa.
```
Motiejus: I think it would be easier if you can refactor this loop out to a separate function, which you could unit-test separately. Since this is side-effect free behavior, it will be easier to read the actual logic of partition healing.

swim/heal_partition.go
```diff
@@ -0,0 +1,111 @@
...
+		// needs reincarnation in partition A.
+		if m1.Status == Alive && m2.Status == Faulty && m1.Incarnation <= m2.Incarnation {
+			copy := m2
+			copy.Status = Suspect
+			csForA = append(csForA, copy)
+		}
+	}
+	// Merge partitions if no node needs to reincarnate,
+	if len(csForA) == 0 && len(csForB) == 0 {
+		// TODO: send out event/log that we are merging
```
Motiejus: Go for it. :-)

swim/heal_partition.go
```diff
@@ -0,0 +1,111 @@
...
+		_, err := sendPingWithChanges(node, target, A1, time.Second)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	// reincarnate all nodes by disseminating that they are suspect
+	// TODO: send out event/log that we are reincarnating
```
Motiejus: Yep, still a TODO.

swim/heal_partition.go
```diff
@@ -0,0 +1,111 @@
...
+
+// AttemptHeal attempts to heal a partition between the node and the target.
+//
+// Be mindfull that calling this function will not result in a heal when there
+// are nodes that need to be reincarated to take precedence over the faulty
+// declarations that occur during a network partition. A cluster may therefore
+// need multiple calls to this function with some time in between to heal.
+//
+// Check out ringpop-common/docs for a full description of the algorithm.
+func AttemptHeal(node *Node, target string) error {
```
Motiejus: I think you can split this out to 3 parts:

1. Find out whether the memberships merge cleanly or not (pure function).
2. Do the first part of partition healing (reincarnate both partitions).
3. Do the second part of partition healing.

And `AttemptHeal` would then become a fiveish-line function, which would make it really cool to read and reason about.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,343 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/benbjohnson/clock"
+	"github.com/stretchr/testify/assert"
+	"github.com/uber/ringpop-go/discovery/statichosts"
+)
+
+func TestPartitionHealWithFaulties(t *testing.T) {
```
Motiejus: Can you describe the scenario in the godoc? E.g.

// TestPartitionHealWithFaulties creates two partitions of 5 nodes. For each partition, the other side is seen as faulty. The membership lists are merge-able. We verify that when the healer kicks in, the partitions are actually healed.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+	waitForPartitionHeal(t, time.Second*3, A, B)
+
+	A.HasPartitionAs(t, A, 3, "alive")
+	A.HasPartitionAs(t, B, 5, "alive")
+	B.HasPartitionAs(t, B, 5, "alive")
+	B.HasPartitionAs(t, A, 3, "alive")
+}
+
+func TestPartitionHealWithMissing(t *testing.T) {
```
Motiejus: Also, a high-level description of the test would be nice (in the similar spirit like I wrote above).

swim/heal_partition_test.go
```diff
@@ -0,0 +1,343 @@
...
+	defer destroyNodes(A...)
+
+	B := newPartition(t, 6)
+	B1 := partition(B[0:3])
+	defer destroyNodes(B...)
+
+	A.ProgressTime(time.Millisecond * 3)
+	B.ProgressTime(time.Millisecond * 5)
+
+	// A.AddPartitionWithStatus(B1, Suspect)
```
Motiejus: Do you need this comment?

swim/heal_partition_test.go
```diff
@@ -0,0 +1,343 @@
...
+type partition []*testNode
+
+func newPartition(t *testing.T, n int) partition {
+	p := partition(genChannelNodes(t, n))
+	bootstrapNodes(t, p...)
+	waitForConvergence(t, 500*time.Millisecond, p...)
+	return p
+}
+
+func (p partition) AddPartitionWithStatus(B partition, status string) {
```
Motiejus: Please document this function. E.g. "adds B (with status `status`) to the membership of p, and clears the disseminator".

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+func (p partition) Contains(address string) bool {
+	for _, a := range p {
+		if a.node.Address() == address {
+			return true
+		}
+	}
+	return false
+}
+
+// waitForPartitionHeal is same as waitForConvergence but unlike that function,
```
Motiejus: You should describe this function without referring how it's similar to other functions. :-) If it really makes sense that way, please copy (parts of) the godoc string.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+	B := newPartition(t, 5)
+	defer destroyNodes(A...)
+	defer destroyNodes(B...)
+
+	A.AddPartitionWithStatus(B, Alive)
+
+	waitForPartitionHeal(t, time.Second*3, A, B)
+}
+
+func TestPartitionHealWithTime(t *testing.T) {
```
Motiejus: Please add a description of the test.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+	}()
+
+	waitForPartitionHeal(t, 3000*time.Millisecond, A, B)
+}
+
+// TestHealBeforeBootstrap tests that we can heal with a node that is still
+// bootstrapping. We put a node in the bootstrapping phase by giving it a
+// host-list that is impossible to join. When it is in this phase, we trigger
+// a heal attempt from another node. The heal should succeed from b's
```
Motiejus: please define b before referring to it.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+// perspective. Healing from bootstrap is very similar to joining a node that
+// is in the bootstrapping phase.
```
Motiejus: The last sentence is confusing. How is that similar to join? To what degree? If you are comparing, it has to be much more detailed (or avoid comparing altogether).

But, still, "very similar" means "a bit different". How is it different?

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+func TestHealBeforeBootstrap(t *testing.T) {
+	a := newChannelNode(t)
+	defer a.Destroy()
+
+	b := newChannelNode(t)
+	defer b.Destroy()
+
+	// block a from completing bootstrap
```
Motiejus: Uh oh how nasty. :-)

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+	discoProvider := statichosts.New(a.node.Address(), b.node.Address())
+
+	// bootstrap a and wait for it to be in bootstrap phase
+	go func() {
+		_, _ = a.node.Bootstrap(&BootstrapOptions{
+			DiscoverProvider: discoProvider,
+		})
+	}()
+
+	// bootstrap b then add a to it's bootstrap provider after
```
Motiejus: looks like sentence isn't finished.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+	// start heal
+	targets := b.node.healer.Heal()
+	assert.Len(t, targets, 1, "expected that b healed with node node a")
+	assert.Equal(t, a.node.Address(), targets[0], "expected b healed with node a")
+
+	// a is now part of b's membership
+	_, has = b.node.memberlist.Member(a.node.Address())
+	assert.True(t, has, "expected that a is now part of b's membership")
+
+	// make sure that a is still not bootstrapped, this causes a to be suspect to be
```
Motiejus: unfinished sentence.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,344 @@
...
+func (p partition) ProgressTime(T time.Duration) {
+	for _, n := range p {
+		now := n.node.clock.Now()
+		c := clock.NewMock()
+		c.Add(now.Add(T).Sub(clock.NewMock().Now()))
+		n.node.clock = c
+	}
+}
+
+// HasPartitionAs checks that every node from A, contains every node from B in its
```
Motiejus: unnecessary comma.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Heal iterates over the hostList that the discoverProvider provides. If the
+// node encounters a host that is faulty or not in the membership, we pick that
+// node as a target to perform a partition heal with.
+//
+// If heal was attempted, returns identities of the target nodes.
+func (h *discoverProviderHealer) Heal() []string {
+	h.node.emit(DiscoHealEvent{})
+	// get list from discovery provider
+	if h.node.discoverProvider == nil {
```
Motiejus: how is this possible?

swim/heal_partition.go
CorgiMan: Nope, we need m1 out of this scope too

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,137 @@
...
+	hostList, err := h.node.discoverProvider.Hosts()
+	if err != nil {
+		h.logger.Warn("unable to receive host list from discover provider")
+		return []string{}
+	}
+
+	h.previousHostListSize = len(hostList)
+
+	// filter hosts that we already know about and attempt to heal nodes that
+	// are complementary to the membership of this node.
```
Motiejus: We should add tests for it (healing more than two partitions).

test/run-integration-tests
```diff
@@ -60,12 +60,13 @@ prefix() {
...
 fetch-ringpop-common() {
     if [ ! -e "$ringpop_common_dir" ]; then
-        run git clone --depth=1 https://github.com/uber/ringpop-common.git:ws-partition-healing "$ringpop_common_dir"
+        run git clone -b ws-partition-healing --depth=1 https://github.com/uber/ringpop-common.git "$ringpop_common_dir"
     fi
 
     run cd "$ringpop_common_dir"
-    run git checkout ws-partition-healing
     run git pull
+    run git checkout ws-partition-healing
```
Motiejus: isn't this redundant if you do the `-b` above?

Motiejus:
I especially like the depth of the tests. Great work.

swim/heal_partition_test.go
CorgiMan: Nope ✋

swim/heal_via_discover_provider.go
CorgiMan: The discover provider is only set on bootstrap not in the constructor. I don't want to be in the risk that this function crashes because the node is still bootstrapping

swim/heal_partition_test.go
CorgiMan: 👍

swim/heal_partition_test.go
CorgiMan: 👍

swim/heal_via_discover_provider.go
CorgiMan: :+1:

swim/heal_partition_test.go
CorgiMan: You like it? 💅

swim/heal_partition_test.go
CorgiMan: 👍

swim/heal_partition_test.go
CorgiMan: 👍

swim/heal_partition_test.go
CorgiMan: 👍

swim/heal_partition_test.go
CorgiMan: 👍

swim/heal_partition.go
CorgiMan: Will do

swim/heal_partition.go
CorgiMan: 👍

swim/heal_partition.go
CorgiMan: Will do

swim/heal_partition_test.go
```diff
@@ -306,8 +303,10 @@ func (p partition) Contains(address string) bool {
 	return false
 }
 
-// waitForPartitionHeal is same as waitForConvergence but unlike that function,
-// waitForPatitionHeal doesn't stop when no dissemination is detected.
+// waitForPartitionHeal let's the nodes of all partitions gossip and returns
```
Motiejus: s/let's/lets

CorgiMan: 👍

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+// Start the partition healing loop
+func (h *discoverProviderHealer) Start() {
+	go func() {
+		for {
+			// attempt heal with the pro
+			p := rand.Float64()
+			P := h.Probability()
```
severb: Rename please

swim/disseminator_test.go
```diff
@@ -119,7 +119,7 @@ func (s *DisseminatorTestSuite) TestFullSync() {
 		s.m.MakeAlive(address, s.incarnation)
 	}
 
-	changes := s.d.FullSync()
+	changes := s.d.MembershipAsChanges()
```
thanodnl: Lets change the name of this test according to the rename of the `FullSync` function

swim/events.go
```diff
@@ -28,6 +28,9 @@ import (
 	"github.com/uber/ringpop-go/events"
 )
 
+// Event interface
+type Event interface{}
```
thanodnl: can't we reuse `events.Event` and keep the helpers surrounding it a bit centralized? A lot in this file seems to be duplicate from the `events` pacakge. eg. The `EventListener` interface.

Besides that the `on` function described below could be useful on other places with regards to events.

swim/events.go
```diff
@@ -234,10 +237,49 @@ type RedundantReverseFullSyncEvent struct {
...
+// - register listener that signals we can continue on receiving the correct event;
+// - execute the code that will lead to an event being emitted;
+// - wait for the continue signal.
+//
+// This can be quite hard to follow. Ideally we want it to look like
+// - execute the code that will lead to an event being emitted;
+// - and then wait for a specific event.
+//
+// This function helps with making the code read like the latter.
+func ExecuteThenWaitFor(f func(), er EventRegistrar, t interface{}) {
```
thanodnl: This pattern is very dangerous to use!

You register a listener which in the body closes a channel without un-registering the listener. The second time this event fires it will cause a panic!

swim/handlers.go
```diff
@@ -56,18 +61,19 @@ func notImplementedHandler(ctx json.Context, req *emptyArg) (*emptyArg, error) {
...
-		"/admin/member/join":  n.adminJoinHandler,
+		"/protocol/join":             n.joinHandler,
+		"/protocol/ping":             n.pingHandler,
+		"/protocol/ping-req":         n.pingRequestHandler,
+		"/admin/debugSet":            notImplementedHandler,
+		"/admin/debugClear":          notImplementedHandler,
+		"/admin/gossip":              n.gossipHandler, // Deprecated
+		"/admin/gossip/start":        n.gossipHandlerStart,
+		"/admin/gossip/stop":         n.gossipHandlerStop,
+		"/admin/healpartition/disco": n.partitionHealerHandler,
```
thanodnl: does the caller care that the partition is healed via the discovery provider? `/disco` seems unnecessary and maybe even un wanted as a name in the endpoint.

swim/handlers_test.go
```diff
@@ -92,6 +92,13 @@ func (s *HandlerTestSuite) TestGossipStopHandler() {
 	s.True(s.testNode.node.gossip.Stopped())
 }
 
+func (s *HandlerTestSuite) TestPartitionHealerHandler() {
+	ExecuteThenWaitFor(func() {
+		_, err := s.testNode.node.partitionHealerHandler(s.ctx, &emptyArg{})
+		s.NoError(err, "calling handler should not result in error")
+	}, s.testNode.node, DiscoHealEvent{})
```
thanodnl: Such use of this function might result in never ending tests in the case that this event is not emitted. Please add a timeout to this function.

swim/heal_partition.go
```diff
@@ -0,0 +1,147 @@
...
+	joinRes, err := sendJoinRequest(node, target, time.Second)
+	if err != nil {
+		return nil, err
+	}
+
+	A := node.disseminator.MembershipAsChanges()
+	B := joinRes.Membership
+
+	// Get the nodes that aren't mergeable and need to reincarnate
+	csForA, csForB := nodesThatNeedToReincarnate(A, B)
```
thanodnl: cs?

swim/heal_partition.go
```diff
@@ -0,0 +1,147 @@
...
+	// Find changes that are alive for B and faulty for A and visa versa.
+	for _, b := range B {
+		a, ok := selectMember(A, b.Address)
+		if !ok {
+			continue
+		}
+
+		// faulty for partition A, alive for partition B
+		// needs reincarnation in partition B.
+		if a.Status == Faulty && b.Status == Alive && a.Incarnation >= b.Incarnation {
```
thanodnl: This breaks in the case where it is a suspect or a tombstone that could still contaminate the other partition. Can we use state precedence here in combination with the incarnation number to figure out what to do?

swim/heal_partition.go
```diff
@@ -0,0 +1,147 @@
...
+func mergePartitions(node *Node, target string, B []Change) error {
+	node.healer.logger.WithField("target", target).Info("merge two partitions")
+
+	// Add membership of B to this node, so that the membership
+	// information of B will be disseminated through A.
+	node.memberlist.Update(B)
+
+	// Send membership of A to the target node, so that the membership
+	// information of partition A will be disseminated through B.
+	A1 := node.disseminator.MembershipAsChanges()
```
thanodnl: Is there a specific reason to call this `A1` over `A`?

swim/heal_partition_test.go
```diff
@@ -0,0 +1,475 @@
...
+	block := make(chan struct{})
+	a.node.RegisterListener(on(JoinTriesUpdateEvent{}, func(e Event) {
+		<-block
+	}))
+
+	discoProvider := statichosts.New(a.node.Address(), b.node.Address())
+
+	// bootstrap a and wait for it to be in bootstrap phase
+	go func() {
+		_, _ = a.node.Bootstrap(&BootstrapOptions{
```
thanodnl: if you don't care about any of the variables you can omit the assignment.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,475 @@
...
+	_, has := b.node.memberlist.Member(a.node.Address())
+	assert.False(t, has, "expected that a is not yet part of b's membership")
+
+	// start heal
+	targets := b.node.healer.Heal()
+	assert.Len(t, targets, 0, "expected that b cannot completely heal with a")
+
+	// a is now part of b's membership
+	_, has = b.node.memberlist.Member(a.node.Address())
+	assert.True(t, has, "expected that a is now part of b's membership")
```
thanodnl: This complete test worries me.

In the past we have seen cases where nodes fail to bootstrap but became known to the rest of the ring because the `/join` endpoint used to `makeAlive` joining nodes and starts gossiping its existence.

If I read this line correctly node a (the one that you blocked during bootstrapping) can now become part of the member while the node its self is not in a state where it sees its self as part of the membership.

swim/heal_partition_test.go
```diff
@@ -0,0 +1,475 @@
...
+	}, a.node, RequestBeforeReadyEvent{})
+
+	// allows a to bootstrap
+	ExecuteThenWaitFor(func() {
+		close(block)
+	}, a.node, JoinCompleteEvent{})
+
+	// progress timers so that incarnation numbers can bump
+	c := clock.NewMock()
+	c.Add(time.Millisecond)
```
thanodnl: what is the reason for progressing the newly created clock mock a millisecond before it is used?

thanodnl: what is the reason for progressing the newly created clock mock a millisecond before it is used?

swim/heal_partition_test.go
```diff
@@ -0,0 +1,475 @@
...
+		for _, node := range nodes {
+			if node.HasChanges() {
+				continue Tick
+			}
+		}
+
+		// continue until all members are reachable
+		for _, node := range nodes {
+			if node.memberlist.CountReachableMembers() != len(nodes) {
+				continue
```
thanodnl: I think you missed the `Tick` here, otherwise this statement does nothing.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,166 @@
...
+
+// discoverProviderHealer attempts to heal a ringpop partition by consulting
+// the discoverProvider for nodes that are faulty or not available in its
+// membership. It attempts the heal probabalisticly so that the discovery provider
+// doesn't get overloaded with request -- with default settings 6
+// times per minutes on avarage for the entire cluster.
+type discoverProviderHealer struct {
+	node *Node
+
+	discoverProvider *discovery.DiscoverProvider
```
thanodnl: this variable seems to be unused as the `discoverProvider` is read from `node` throughout this file.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,166 @@
...
+// If heal was attempted, returns identities of the target nodes.
+func (h *discoverProviderHealer) Heal() []string {
+	h.node.emit(DiscoHealEvent{})
+	// get list from discovery provider
+	if h.node.discoverProvider == nil {
+		return []string{}
+	}
+	hostList, err := h.node.discoverProvider.Hosts()
+	if err != nil {
+		h.logger.Warn("unable to receive host list from discover provider")
```
thanodnl: can we leave a hint in the logline that it was during a heal attempt?

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,166 @@
...
+		if !ok || m.Status == Faulty {
+			targets = append(targets, address)
+		}
+	}
+
+	// filter hosts that we already know about and attempt to heal nodes that
+	// are complementary to the membership of this node.
+	var ret []string
+
+	for len(targets) != 0 {
```
thanodnl: from the documents I thought we only try to heal 1 partition at a time, but reading this for loop it seems to me that it now reaches out to all hosts that is does not know about (skipping the ones that are found in a partition that has a heal attempt executed).

swim/node.go
```diff
@@ -96,6 +110,12 @@ func mergeDefaultOptions(opts *Options) *Options {
 	opts.RollupMaxUpdates = util.SelectInt(opts.RollupMaxUpdates, def.RollupMaxUpdates)
 	opts.RollupFlushInterval = util.SelectDuration(opts.RollupFlushInterval, def.RollupFlushInterval)
 
+	opts.PartitionHealPeriod = util.SelectDuration(opts.PartitionHealPeriod,
+		def.PartitionHealPeriod)
```
thanodnl: Can we put this on 1 line?

swim/node.go
```diff
@@ -278,6 +309,10 @@ func (n *Node) Stopped() bool {
 
 // Destroy stops the SWIM protocol and all sub-protocols.
 func (n *Node) Destroy() {
+	if n.state.destroyed {
```
thanodnl: Good catch!

However, there is a race here where `Destroy` might be called in quick succession. What is the effect of having `Stop` and `rollup.Destroy` called twice?

swim/node_bootstrap_test.go
```diff
@@ -68,24 +70,14 @@ func (s *BootstrapTestSuite) TestBootstrapJoinsTimeOut() {
...
-			JoinTimeout:      time.Millisecond,
-		})
-		errChan <- err
-	}()
-
-	// Block until the error is received from the bootstrap
-	// goroutine above.
-	chanErr := <-errChan
-	s.EqualError(chanErr, "node destroyed while attempting to join cluster")
+	// node := newChannelNode(s.T()).node
```
thanodnl: dead code

test/run-integration-tests
```diff
@@ -60,12 +60,13 @@ prefix() {
 #
 fetch-ringpop-common() {
     if [ ! -e "$ringpop_common_dir" ]; then
-        run git clone --depth=1 https://github.com/uber/ringpop-common.git "$ringpop_common_dir"
+        run git clone -b ws-partition-healing --depth=1 https://github.com/uber/ringpop-common.git "$ringpop_common_dir"
```
thanodnl: make sure the changes regarding the branch selection are removed before merging.

swim/heal_via_discover_provider.go
CorgiMan: If there are a lot of single node partitions then this would take a long time to heal them. This counteracts that but I'm not sure if this is the best solution. I believe Ben suggested this on the RFC.

swim/heal_via_discover_provider.go
CorgiMan: Doesn't the module logger log this? I'm not sure actually

CorgiMan: Nope it does not. Will change.

swim/heal_partition.go
CorgiMan: Not anymore

swim/heal_partition_test.go
CorgiMan: Isn't what you describe normal join behaviour?

swim/heal_partition_test.go
CorgiMan: `a.node.clock` is at 0 milliseconds. Here a clock is created that is at 1 millisecond.

swim/heal_partition_test.go
CorgiMan: Good catch

swim/node.go
CorgiMan: I believe it hangs

swim/disseminator_test.go
CorgiMan: Done

swim/handlers.go
CorgiMan: We might implement different ways of healing. With faulties or by specifying a host we want to heal with. I don't think it's a big deal that the user has to call it like this.

swim/handlers_test.go
CorgiMan: I think this is okay and not worth the effort. If the test hangs you will notice it then.

swim/handlers.go
thanodnl: In that case we should reflect it as well in the name of the handler.

test/run-integration-tests
CorgiMan: Will do, that PR still needs to land, can you take a look if you have any more comments on that?

swim/handlers.go
CorgiMan: Agreed

swim/heal_partition_test.go
thanodnl: No, not really. It has to do with `T212923` that is resolved in both ringpop-go and ringpop-node

- ringpop-go: https://github.com/uber/ringpop-go/pull/65
- ringpop-node: https://github.com/uber/ringpop-node/pull/217

The problem here is that the node is being gossiped around as alive while the node its self has not started pinging because it is not able to complete the bootstrapping. This caused non-recovering partitions.

Even though the partition healing would eventually heal the partition I believe it is undesirable to have a node in the ring that is not pinging on its own. So we should prevent nodes that have not bootstrapped completely (eg. blocked in the bootstrapping phase) to join a ring due to partition healing.

swim/heal_partition_test.go
thanodnl: I understand that the mock clock is now at millisecond 1, but why do we need to progress the time of the mock clock?

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,164 @@
...
+		h.logger.Warn("healer unable to receive host list from discover provider")
+		return []string{}
+	}
+
+	h.previousHostListSize = len(hostList)
+
+	var targets []string
+	for _, address := range hostList {
+		m, ok := h.node.memberlist.Member(address)
+		if !ok || m.Status == Faulty {
```
severb: I think we should use `statePrecedence` here, to make sure this will work with `tombstone` in the future.

swim/handlers_test.go
dansimau: Disagree with the previous comment. If the test hangs, that is really bad and annoying. Imagine on Travis: the CI result would take 10 minutes to come in and blocks everything.

Considering how easy it is to add a timeout (2-3 lines of code?) I think we should add a timeout that fails the test when the timeout hits.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,164 @@
...
+
+	logger log.Logger
+}
+
+func newDiscoverProviderHealer(n *Node, baseProbability float64, period time.Duration) *discoverProviderHealer {
+	return &discoverProviderHealer{
+		node:             n,
+		baseProbabillity: baseProbability,
+		period:           period,
+		quit:             make(chan struct{}),
```
thanodnl: With the `quit` channel initialized here only once there is no way to `Start` the healer again after it `Stop` has been called on it. This would be the first submodule of `Node` that has this behavior, both the `gossip` and `stateTransitition` modules can be stopped and started multiple times through their life time.

swim/events.go
```diff
@@ -234,10 +229,48 @@ type RedundantReverseFullSyncEvent struct {
...
+// - register listener that signals we can continue on receiving the correct event;
+// - execute the code that will lead to an event being emitted;
+// - wait for the continue signal.
+//
+// This can be quite hard to follow. Ideally we want it to look like
+// - execute the code that will lead to an event being emitted;
+// - and then wait for a specific event.
+//
+// This function helps with making the code read like the latter.
+func ExecuteThenWaitFor(f func(), er events.EventRegistrar, t interface{}) {
```
thanodnl: I am still a bit scared by this code, it does not cause panic's anymore which is already better. However it causes a lot of garbage to accumulate in the event listener if this would be used outside of the tests. Since it is only used as a helper function during testing can we move it to a file ending in `_test.go` so it is available to tests, but can never be used by the library?

severb: Is there a way to unsubscribe from events? We can have a `once` utility similar to `on`.

swim/heal_partition_test.go
CorgiMan: If I understand correctly, that pr prevents the join handler from adding the node to the membership.
So if node a sends a join to node b, node b no longer adds a to its membership.

It is essential for nodes to be able to join nodes that are still bootstrapping, otherwise we could never form a cluster.

swim/heal_partition_test.go
CorgiMan: So that the incarnation number can bump

swim/heal_partition.go
```diff
@@ -0,0 +1,154 @@
...
+// selectMember selects the member with the specified address from the partition.
+func selectMember(partition []Change, address string) (Change, bool) {
+	for _, m := range partition {
+		if m.Address == address {
+			return m, true
+		}
+	}
+
+	return Change{}, false
+}
```
severb: I really like the procedural approach for this algorithm (even if imperative). I think this makes the algorithm easy to follow and removes the philosophical questions of what object each method belongs to and how to share state.

swim/handlers_test.go
CorgiMan: I disagree seeing the likelihood that this would happen is small. But since you both feel this way I'll do so

swim/events.go
CorgiMan: I added it to test_utils.go but then moved it back. I'll move it there again. I think it's a shortcoming of the event system that you can't unsubscribe though.

swim/heal_partition.go
CorgiMan: Thanks Sever

swim/heal_partition.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, err
+	}
+
+	A := node.disseminator.MembershipAsChanges()
+	B := joinRes.Membership
+
+	// Get the nodes that aren't mergeable and need to reincarnate
+	changesForA, changesForB := nodesThatNeedToReincarnate(A, B)
+
+	// Reincarnate the nodes that need to reincarnate
```
Motiejus: s/reincarnate/be reincarnated/

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,168 @@
...
+// If heal was attempted, returns identities of the target nodes.
+func (h *discoverProviderHealer) Heal() []string {
+	h.node.emit(DiscoHealEvent{})
+	// get list from discovery provider
+	if h.node.discoverProvider == nil {
+		return []string{}
+	}
+	hostList, err := h.node.discoverProvider.Hosts()
+	if err != nil {
+		h.logger.Warn("healer unable to receive host list from discover provider")
```
Motiejus: The admin endpoint exposes a list of hosts to be returned, and an error. Can you propagate the error from here, so the query from the admin endpoint can see it too?

Reason: operator triggers a /heal, which returns a success and empty list of hosts, but actually fails.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,168 @@
...
+	var ret []string
+
+	for len(targets) != 0 {
+		target := targets[0]
+		targets = del(targets, target)
+
+		// try to heal partition
+		hostsOnOtherSide, err := AttemptHeal(h.node, target)
+		if err != nil {
+			h.logger.WithField("error", err.Error()).Warn("heal failed")
```
Motiejus: Same here, maybe error should be returned and propagated.

swim/heal_via_discover_provider.go
Motiejus: I am still not convinced this is the better way to go.

In case of single-node clusters that haven't seen the rest of the cluster before (the only realistic way that many single-node clusters can happen IMO), _they_ will start the healing processes separately, thus quickly joining to the "big" cluster. Unless there are more arguments, I think it's a bit premature to develop for that case.

@benfleis what's your take on it?

swim/handlers.go
```diff
@@ -115,6 +121,10 @@ func (n *Node) gossipHandlerStop(ctx json.Context, req *emptyArg) (*emptyArg, er
 	return &emptyArg{}, nil
 }
 
+func (n *Node) discoverProviderHealerHandler(ctx json.Context, req *emptyArg) (*HealResponse, error) {
+	return &HealResponse{n.healer.Heal()}, nil
```
Motiejus: Like mentioned below, it might be useful to return an error here.

swim/test_utils.go
```diff
@@ -156,15 +157,21 @@ func bootstrapNodes(t *testing.T, testNodes ...*testNode) []string {
 	return hostports
 }
 
+// waitForConvergence let's the nodes gossip and returns when the nodes are converged.
```
Motiejus: minor, but since there is feedback anyway: `s/let's/lets`

Motiejus:
I'm happy with the changes, except:

1. `/admin/healpartition/disco` should return errors if the heal fails.
2. I am not convinced that reaching out to all the nodes which are applicable for healing is the best approach (see inline comments).

Since I'm OOO tomorrow and you might want to merge it before Thursday, please have a chat with the rest of the team about (2). If there is consensus that we indeed want to reach out to all the nodes, then LGTM from me.

swim/heal_via_discover_provider.go
benfleis: I haven't reviewed this diff, so my response is generic.

IMHO, there ought to be an independent process which considers the list of expected, but unavailable peers (EBUPs), and pingjoins a small constant number of them per period (1-3). That period ought to be reasonable but short (1-5s?); in the normal/stable case, it will trigger nothing.

Possible tweaks: When a pingjoin initiator adds (a) new node(s), it could also immediately enqueue another pingjoin op, with the ideal that healing begets healing. It could equally retard that timer when all EBUPs have repeatedly timed out, although this seems less useful if the main timer is well set. IIC, this lets many small partitions seek each other out less frequently, while allowing the first healers to accelerate upon contact.

@Motiejus I don't wholly buy the notion that "buildup" syncs are super expensive, or that bandwidth caps are a threat. Do we have measurements of a full-sync when a singleton joins a 1k cluster? I would be curious to know the numbers.

Motiejus: Membership information for a single node is around 100 byte range (`source`, `address`, `status`, `incarnationNumber`). On 1000 nodes, that's 100 KB for a single message `/join` message.

For every `failed` node (that's 1000 in this case, when a single node is joining an existing healthy cluster), it will send another `/join`, they will all return their memberships. That's 100KB * 1000 = 100MB, just in this loop. Let's ignore what happens later.

Let's say we start 100 nodes and are initially started as single-node clusters (e.g. problem with the discovery provider). Within ~10 seconds, partition healing will kick in in every one of them; assuming the discovery provider returns the full cluster, the new nodes will cause 10GB of data to be transferred in 10 seconds.

For me, a single node being able to cause 100MB of metadata to be sent seems large already, even in non-flapping case.

Motiejus: I had another chat with Wieger and take back my concerns. I'm fine with looping here without stopping.

The key that I missed: `size(membership) != size(membership-from-discoverProvider)`.

Motiejus:
LGTM after the `/admin/healpartition/disco` starts propagating errors.

swim/heal_via_discover_provider.go
CorgiMan: I think we need a balance here. If we continue to heal while there is an even split network partition of 1000 nodes, we don't want to have try every one of them. We also don't want to stop at the first faulty we see. There might be a legitimate faulty node. I think if we set this balance to an somewhat arbitrary 10 nodes, we at least have an upper bound of how long the partition heal loop will last, and also enough (from what we know from production clusters) resilience against true faulty nodes.

util/util.go
```diff
@@ -184,6 +184,14 @@ func ShuffleStrings(strings []string) []string {
 	return newStrings
 }
 
+// ShuffleStrings uses the Fisher–Yates shuffle to randomize the strings in place.
```
severb: ❤️ This is not trivial to get right, glad to see it done right.

severb:
What a remarkable PR. Now that the suspect transfer issue @thanodnl raised is solved, I think this is ready to land.

Kudos to @Motiejus too. LGT💵

swim/handlers.go
```diff
@@ -48,6 +48,12 @@ type Status struct {
 	Status string `json:"status"`
 }
 
+// HealResponse contains a list of nodes where healing was attempted
+type HealResponse struct {
+	Targets []string `json:"targets"`
+	Error   error    `json:"error"`
```
thanodnl: the `error` interface does not nicely serialize to json. I think it will be better to call [`Error()`](https://golang.org/pkg/builtin/#error) on the interface and put the string here.

quick test showed that an error gets serialized like this in json:
```golang
```

swim/heal_partition.go
```diff
@@ -0,0 +1,145 @@
...
+	// Find changes that are alive for B and faulty for A and visa versa.
+	for _, b := range B {
+		a, ok := selectMember(A, b.Address)
+		if !ok {
+			continue
+		}
+
+		// If a node would be overwritten and would stop being pingable, it is
+		// not suited for merging.
+		if b.isPingable() && a.overrides(b) && !a.isPingable() {
```
thanodnl: really love how understandable this line has become! good job.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,190 @@
...
+	select {
+	case h.started <- struct{}{}:
+	default:
+		return
+	}
+
+	go func() {
+		for {
+			// attempt heal with the pro
+			if rand.Float64() < h.Probability() {
```
thanodnl: how sure are we that `rand` has been seeded? By default rand does not seed causing all nodes to hit the same numbers from `Float64()`.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,190 @@
...
+		targets = del(targets, target)
+
+		// try to heal partition
+		hostsOnOtherSide, err := AttemptHeal(h.node, target)
+
+		if err != nil {
+			h.logger.WithFields(log.Fields{
+				"error":   err.Error(),
+				"failure": failures,
+			}).Warn("heal attempt failed (10 in total)")
```
thanodnl: I don't think we need the `(10 in total)` part of this log line. We log the `healer reached max failures` if we exceed the allowed number of failures.

swim/heal_via_discover_provider.go
```diff
@@ -0,0 +1,190 @@
...
+	case h.started <- struct{}{}:
+	default:
+		return
+	}
+
+	go func() {
+		for {
+			// attempt heal with the pro
+			if rand.Float64() < h.Probability() {
+				h.Heal()
```
thanodnl: Would it be possible to stat a counter when we initiate a heal via this probability so we can validate in production that the probability mechanism works as expected?

swim/handlers.go
CorgiMan: Great feedback

swim/heal_via_discover_provider.go
CorgiMan: Good idea, I added a stat that listens for DiscoHealEvent

PR #136, opened 2016-04-12
Fix mocks not being generated on Travis

dansimau:
`mockery` behaviour has changed upstream, breaking our mock generation.

See this issue:
https://github.com/vektra/mockery/issues/73

To fix this in the meantime, we install a specific version of `mockery` on Travis.


dansimau:
It turns out there's yet another change in mockery that makes it break even at `b0804d5`.

`130a05e` seems to work. Will use that for now and possibly look at filing another issue upstream.

CorgiMan:
Looks good!

mennopruijssers:
👍

PR #137, opened 2016-04-13
Check for race conditions on Travis

dansimau:
This PR adds a `make test-race` that runs the Go race detector and sets up Travis to run this automatically.


thanodnl:
LGTM

Lets try and squash the races that are in there at some point so we can make it a mandatory test.

CorgiMan:
Looks good!

PR #138, opened 2016-04-28
Bump severity level for membership log

severb:
Some folks requested this message to be logged with a higher severity level.

### DIFF
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -300,7 +300,7 @@ func (m *memberlist) MakeChange(address string, incarnation int64, status string
 	if len(changes) > 0 {
 		m.logger.WithFields(bark.Fields{
 			"update": changes[0],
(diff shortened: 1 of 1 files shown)


swim/memberlist.go
```diff
@@ -300,7 +300,7 @@ func (m *memberlist) MakeChange(address string, incarnation int64, status string
 	if len(changes) > 0 {
 		m.logger.WithFields(bark.Fields{
 			"update": changes[0],
-		}).Debugf("ringpop member declares other member %s", changes[0].Status)
+		}).Infof("ringpop member declares other member %s", changes[0].Status)
```
jwolski: This type of change is going to impact all of Ringpop customers. Some customers are a bit more sensitive to log spew. Are you sure you want to do this? :)

severb: Right. We added the module logger for this type of situations.

PR #141, opened 2016-05-04
Support headers forwarding

severb:
Support optional headers when forwarding.

The implementation is similar to @prashantv's #24 PR; the difference being that the headers are optional. This keeps the API backward compatible.


forward/forwarder.go
```diff
@@ -50,6 +50,7 @@ type Options struct {
 	RerouteRetries bool
 	RetrySchedule  []time.Duration
 	Timeout        time.Duration
+	Headers        []byte
```
Motiejus: Headers in tchannel are represented as a map of strings:
in `tchannel-go/context_header.go`:
```
30     Headers() map[string]string
```

Maybe we should expose a data structure similar to tchannel's?

severb: *I think* this is a generic approach that works when forwarding raw requests, where the headers can be anything.

Maybe @prashantv can shed some light on why he chose to use `[]byte` in his original PR.

prashantv: Sure. So TChannel itself only has a notion of "bag of bytes" for arg2. The encoding of headers from `map[string]string` to the `[]byte` is encoding dependent. The specific encoding details can be found at:

severb: @prashantv thanks for the response. Indeed, for HTTP, `arg2` has other uses too; and I think the same can be said for JSON since there is no restriction on having only string values in the object.

I feel good about the Ringpop forward API exposing headers as `[]bytes` but I understand how this can be confusing. @thanodnl suggested that I should add a few examples in Ringpop on how the headers forwarding works in practice. This will demo the idiomatic way of using this feature.

Alternatively, we also discussed adding an utility function for the headers encoding. But since each encoding is different (HTTP needs the method and the URL and JSON is not limited to `map[string]string`) I think this won't bring too much value to our users; educating them on how to forward `arg2` as bytes and what serialization is available already in tchannel is probably better.

weikai77:
Ideally I'd like ringpop to hide all the serialization/deserialization, but given that has not been the design decision from the get-go, I'm ok with passing the headers as bytes.

sunzl1989:
LGTM. And +1 to adding examples to demo how to use this idiomatically.

Motiejus:
LGTM

thanodnl:
LGTM

severb:
Thanks! Let's land this!

PR #142, opened 2016-05-10
Change stat name emitted when healing

mennopruijssers:
Change stat name emitted from `heal.triggered` to `heal.triggered.discover_provider` when healing through discover provider.
Because the partition healing feature is built with multiple strategies in mind, it makes sense to continue on this path to make sure to support multiple strategies in the stat-name as well.


Motiejus:
👍 

We missed one when we did golang, maybe you can take a minute and add this name to https://github.com/uber/ringpop-common/blob/master/docs/running_ringpop.md#monitoring ?

mennopruijssers:
Done: uber/ringpop-common#57

thanodnl:
LGTM

mennopruijssers:
We decided to change the one in node instead.

PR #144, opened 2016-05-13
Update testpop to emit stats to network or file

severb:
In preparation for automated perf. tests we want to be able to emit stats to a file/network from testpop. 

This PR adds a new flag to testpop to control where the stats should go; if the flag is omitted, no stats are emitted.


mennopruijssers:
It looks like builds are failing to due a missing dependency:
```
scripts/testpop/statter.go:70: undefined: statsd.NewClientWithSender
```

severb:
Yes, indeed. The build needs the latest (master tip) version of the statsd client library. That's because they added a new constructor that can be used to pass in your own custom `Sender` implementation.

scripts/testpop/testpop.go
```diff
@@ -57,14 +60,33 @@ func main() {
...
-	)
+		ringpop.SuspectPeriod(5 * time.Second),
+		ringpop.FaultyPeriod(5 * time.Second),
+		ringpop.TombstonePeriod(5 * time.Second)}
+
+	if *stats != "" {
+		var statsdClient statsd.Statter
+		var err error
+		if statsPattern.MatchString(*stats) {
+			statsdClient, err = statsd.New(*stats, "")
```
Motiejus: I think the current approach is quite brittle: what if the name is `./oneoff-test:1234`? Could we match a [URI scheme][1] instead? E.g. prefixes like `file://`, `udp://`?

[1]: http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

Motiejus:
I don't see tests that different `stats` parameters show different behaviors. Are you going to add integration tests for it?

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/cactus/go-statsd-client/statsd"
```
Motiejus: I don't see this in `Godeps`. Are we missing some files?

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+	return len(data), nil
+}
+
+// Close implements the stats.Sender interface
+func (fs *FormattedSender) Close() error {
+	return fs.s.Close()
+}
+
+// NewFileStatsd returns a statsd.Statter that writes to file. Each entry is
+// prefixed by a date/time with nanosecond resolution on it's own line
```
Motiejus: Looks like Node version [gives only milliseconds][1]. Should we keep them compatible?

[1]: https://github.com/uber/ringpop-node/pull/271/files#diff-7a9076d6d94e62c13d641aa71f19ae8eR142

scripts/testpop/testpop.go
mennopruijssers: The problem with a file-scheme is that it officially doesn't support relative paths. There's a trade-off between the two and we thought supporting relative files is something that would improve developer experience.

scripts/testpop/statter.go
mennopruijssers: @severb We agreed on RFC3339 or RFC3339Nano for go right?
Both of them should be compatible with what node is sending:
```
new Date().toISOString()
'2016-05-23T16:50:21.852Z'
```

scripts/testpop/testpop.go
Motiejus: Should it then be `--stats-file` and `--stats-udp` to disambiguate between the two?

scripts/testpop/statter.go
mennopruijssers: It was already in godeps. I needed to update it though.

Makefile
```diff
@@ -71,4 +71,4 @@ test-race:
 	test/go-test-prettify -race ./...
 
 testpop:	clean
-	go build scripts/testpop/testpop.go
+	go build ./scripts/testpop/
```
mennopruijssers: Since testpop now needs the other file in the testpop-directory (`statter.go`) as well, changed this to a relative path to the directory.

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+func NewWriteCloserToSender(wc io.WriteCloser) statsd.Sender {
+	return &WriteCloserToSender{
+		bufWriter: bufio.NewWriter(wc),
+		closer:    wc}
+}
+
+// FormattedSender wraps a statsd.Sender and adds a timestamp and new lines
+// after each Send call.
+type FormattedSender struct {
+	s statsd.Sender
```
CorgiMan: if you embed this field (ie omit the `"s "`), the close method will be automatically embedded as well.

It's a neat trick which you might not know yet. I'm cool with letting it be extra explicit though.

I think you can also make it a typedef `type FormattedSender statsd.Sender`, but you might run into a recursion problem

CorgiMan:
Looks good!

scripts/testpop/testpop.go
```diff
@@ -57,14 +60,40 @@ func main() {
...
+		}
+
+		if *statsFile != "" {
+			statsdClient, err = NewFileStatsd(*statsFile)
+			if err != nil {
+				log.Fatalf("colud not open stats file: %v", err)
+			}
+		}
+
+		stattter := bark.NewStatsReporterFromCactus(statsdClient)
```
btromanova: Change this to "statter", pls.

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/cactus/go-statsd-client/statsd"
+)
+
+// WriteCloserToSender is an adapter from io.WriteCloser to statsd.Sender
+type WriteCloserToSender struct {
```
btromanova: I suggest we change the name to smth like "FileSender", because I read current version as "write closer to (= not far from) sender".

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// FormattedSender wraps a statsd.Sender and adds a timestamp and new lines
+// after each Send call.
+type FormattedSender struct {
+	s statsd.Sender
+}
+
+// Send implements the statsd.Sender interface
+func (fs *FormattedSender) Send(data []byte) (int, error) {
```
btromanova: Why do we need extra wrapper here? Can we have a single FileSender, that writes formatted data? Less code, easier to read.

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+	line := fmt.Sprintf("%s: %s\n", time.Now().UTC().Format(time.RFC3339Nano), data)
+	_, err := fs.s.Send([]byte(line))
+	// Because we're changing the undelying bytes sent, make sure:
```
btromanova: Is this comment still relevant?

CorgiMan:
I still have some comments

scripts/testpop/statter.go
CorgiMan: I agree with Tania. This would be easier to read if we had one type that implements a statsd.Sender given a file name.

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+
+// WriteCloserToSender is an adapter from io.WriteCloser to statsd.Sender
+type WriteCloserToSender struct {
+	bufWriter *bufio.Writer
+	closer    io.Closer
+}
+
+// Send implements the statsd.Sender interface
+func (w *WriteCloserToSender) Send(data []byte) (int, error) {
+	return w.bufWriter.Write(data)
```
CorgiMan: Don't we need to flush the writer every time we want to send, else I don't really get why we'd use a buffered writer

scripts/testpop/statter.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Close implements the statsd.Sender interface
+func (w *WriteCloserToSender) Close() error {
+	err := w.bufWriter.Flush()
+	w.closer.Close()
+	return err
+}
+
+// NewWriteCloserToSender returns a new WriteCloserToSender adapter
+func NewWriteCloserToSender(wc io.WriteCloser) statsd.Sender {
```
CorgiMan: It think it's more useful to return a WriterCloserToSender and comment that it implements a statsd.Sender.
Same that os.Open returns a *File which implements the reader interface

scripts/testpop/statter.go
mennopruijssers: We flush before closing, until then we're buffering the default buffer size (4K) in memory so we don't have IO on each stat.

scripts/testpop/statter.go
mennopruijssers: > I think you can also make it a typedef type FormattedSender statsd.Sender, but you might run into a recursion problem
Yeah, that won't work but embedding the sender is nice :)

scripts/testpop/statter.go
mennopruijssers: I agree.
@severb could you chip in here and explain as the original author your preference?

scripts/testpop/statter.go
mennopruijssers: Yes

scripts/testpop/statter.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Close implements the statsd.Sender interface
+func (fs *fileSender) Close() error {
+	err := fs.bufWriter.Flush()
+	fs.closer.Close()
+	return err
+}
+
+// NewWriteCloserToSender returns a new WriteCloserToSender adapter
```
btromanova: Change comment.

btromanova:
lgtm

CorgiMan:
Excellent! Looks good.

PR #145, opened 2016-05-13
Fix: Race condition on hostlist reuse

thanodnl:
There could be a case when the implementation of the `DiscoveryProvider` reuses the `hostlist` it returns that the in place shuffle would be triggered concurrently on that list.

Since the shuffled `hostlist` is only used to create a `targets` list we do not have to rely on the `hostlist` being shuffled if we shuffle the `targets` list. Since we 'own' the `targets` list we know it will not be used concurrently by an other routine.


mennopruijssers:
LGTM!

mennopruijssers:
(This is also improving performance since the list to shuffle is shorter ;))

dansimau:
LGTM 👍

PR #147, opened 2016-05-26
Make hashring consistent on hash colissions

CorgiMan:
When a server is added to the red-black tree with the same hash as an server that already exists in the hashring, the server with the smallest lexicografical ordering wins.

### DIFF
--- a/hashring/hashring_test.go
+++ b/hashring/hashring_test.go
@@ -368,3 +368,23 @@ func TestConsistentOnCollision(t *testing.T) {
 
 	assert.Equal(t, m1, m2, "hashring isn't consistent due to hash collisions")
 }
+
+func TestConsistentOnCollision2(t *testing.T) {
+	// make 11 replica points so that 127.0.0.1:3010 collides
+	r1 := New(farm.Fingerprint32, 11)
+	r2 := New(farm.Fingerprint32, 11)
+
+	r1.AddServer("127.0.0.1:30")
+	r1.AddServer("127.0.0.1:301")
+	r1.AddServer("127.0.0.1:4006")
+	r1.RemoveServer("127.0.0.1:30")
+
+	r2.AddServer("127.0.0.1:301")
+	r2.AddServer("127.0.0.1:4000")
+	r2.AddServer("127.0.0.1:4006")
+
+	m1, _ := r1.Lookup("127.0.0.1:3010")
+	m2, _ := r2.Lookup("127.0.0.1:3010")
+
+	assert.Equal(t, m1, m2, "hashring isn't consistent due to hash collisions")
+}
--- a/hashring/hashring_test.go
+++ b/hashring/hashring_test.go
@@ -352,3 +352,19 @@ func BenchmarkHashRingLookupN(b *testing.B) {
 		}
 	}
 }
+
+func TestConsistentOnCollision(t *testing.T) {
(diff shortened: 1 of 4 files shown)


Motiejus:
The sentence can be a bit clarified to:

When a server is added to the red-black tree with the same hash as an server that already exists in the hashring, the server with the smallest lexicografical ordering wins.

thanodnl:
LGTM

hashring/hashring_test.go
```diff
@@ -352,3 +352,19 @@ func BenchmarkHashRingLookupN(b *testing.B) {
...
+
+	r1.AddServer("127.0.0.1:30")
+	r1.AddServer("127.0.0.1:301")
+	r2.AddServer("127.0.0.1:301")
+	r2.AddServer("127.0.0.1:30")
+
+	m1, _ := r1.Lookup("127.0.0.1:3010")
+	m2, _ := r2.Lookup("127.0.0.1:3010")
+
+	assert.Equal(t, m1, m2, "hashring isn't consistent due to hash collisions")
```
mennopruijssers: The message of an assert should describe the expected behavior.

mennopruijssers: I would prefer to actually compare the full hash rings for completeness if possible. (also my previous comment on the assert still holds up).

hashring/rbtree.go
```diff
@@ -236,7 +238,11 @@ func (t *redBlackTree) Delete(val int) bool {
 
 func (n *redBlackNode) search(val int) (string, bool) {
 	if n.val == val {
-		return n.str, true
+		m := n
```
btromanova: Maybe it's better to use recursion instead of loop, like that:
```
if  n.left == nil {
  return n.str, true
}
return n.left.search(val)
```

hashring/rbtree.go
```diff
@@ -154,7 +155,7 @@ func (t *redBlackTree) Insert(val int, str string) (ret bool) {
 
 // Delete removes a value from the redBlackTree
```
btromanova: Can you please update comments on all the methods that changed?

hashring/rbtree_test.go
```diff
@@ -567,13 +567,62 @@ func TestDelete(t *testing.T) {
...
 	assert.Equal(t, 1, height, "expected tree to have black height of 1")
 }
 
+func TestCollision(t *testing.T) {
+	tree := makeTree()
+
+	var str string
+	var ok bool
+
+	// "seven"<"seven-two" so "seven" remains in tree
```
btromanova: seven and seven-two are both in the tree, right?
seven-two is not searchable, but I think this comment in the test is confusing.

CorgiMan: You are right, that comment is stale

hashring/rbtree.go
CorgiMan: Recursion might make things clearer and I think you're right about lookupN

hashring/hashring_test.go
CorgiMan: We need to have more tests indeed, I don't feel great about playing around in this code so maybe let's make sure that at least the tests are robust.

malandrew:
@CorgiMan This pull request seems related to my changes in #211 that just landed. Can we get this pull request updated and landed?

thanodnl:
This has been fixed in #196 
Closing

PR #148, opened 2016-05-30
add suspect, faulty and tombstone period flags

CorgiMan:
Add suspect, faulty and tombstone period flags so that integration tests can set a small faulty->tombstone period of 5 seconds while having a big faulty->tombstone period when running from tickcluster.

### DIFF
--- a/scripts/testpop/testpop.go
+++ b/scripts/testpop/testpop.go
@@ -38,9 +38,9 @@ var (
 	hostport = flag.String("listen", "127.0.0.1:3000", "hostport to start ringpop on")
 	hostfile = flag.String("hosts", "./hosts.json", "path to hosts file")
 
-	suspectPeriod   = flag.Int("suspect-period", 5000, "suspect period in ms")
-	faultyPeriod    = flag.Int("faulty-period", 24*60*60*1000, "faulty period in ms")
-	tombstonePeriod = flag.Int("tombstone-period", 5000, "tombstone period in ms")
+	suspectPeriod   = flag.Int("suspect-period", 5000, "The lifetime of a suspect member in ms. After that the member becomes faulty.")
+	faultyPeriod    = flag.Int("faulty-period", 24*60*60*1000, "The lifetime of a faulty member in ms. After that the member becomes a tombstone.")
+	tombstonePeriod = flag.Int("tombstone-period", 5000, "The lifetime of a tombstone member in ms. After that the member is removed from the membership.")
 
 	hostportPattern = regexp.MustCompile(`^(\d+.\d+.\d+.\d+):\d+$`)
 )
--- a/scripts/testpop/testpop.go
+++ b/scripts/testpop/testpop.go
@@ -35,8 +35,13 @@ import (
 )
 
 var (
-	hostport        = flag.String("listen", "127.0.0.1:3000", "hostport to start ringpop on")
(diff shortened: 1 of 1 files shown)


scripts/testpop/testpop.go
```diff
@@ -35,8 +35,13 @@ import (
 )
 
 var (
-	hostport        = flag.String("listen", "127.0.0.1:3000", "hostport to start ringpop on")
-	hostfile        = flag.String("hosts", "./hosts.json", "path to hosts file")
+	hostport = flag.String("listen", "127.0.0.1:3000", "hostport to start ringpop on")
+	hostfile = flag.String("hosts", "./hosts.json", "path to hosts file")
+
+	suspectPeriod   = flag.Int("suspect", 5, "suspect period in seconds")
```
dansimau: * Why not "--suspect-period="? They are called "long options", after all 😄  I think if it's just called "suspect" one might have to consult the --help to clarify what it means. (Consider "suspect=5" without context. What does it mean? "5 suspects"?)

* Why in seconds and not milliseconds? Can the value have a decimal place (e.g. 1.5)? Why not just use Go's built in notation (e.g user can specify "5s", or  "800ms")? I understand we'd probably want parity with main.js in Node too; so that's a consideration.

CorgiMan: I'll add the "-period" prefix, I can make it ms. Didn't use go's duration flag because of node..

scripts/testpop/testpop.go
```diff
@@ -35,8 +35,13 @@ import (
...
+	faultyPeriod    = flag.Int("faulty", 24*60*60, "faulty period in seconds")
```
mennopruijssers: swapped the default values for the faulty and tombstone periods?

scripts/testpop/testpop.go
```diff
@@ -61,9 +66,10 @@ func main() {
 		ringpop.Channel(ch),
 		ringpop.Identity(*hostport),
 		ringpop.Logger(bark.NewLoggerFromLogrus(logger)),
-		ringpop.SuspectPeriod(5*time.Second),
-		ringpop.FaultyPeriod(5*time.Second),
-		ringpop.TombstonePeriod(5*time.Second),
+
+		ringpop.SuspectPeriod(time.Duration(*suspectPeriod)*time.Second),
```
mennopruijssers: Why not just `*..period * time.Second`?

scripts/testpop/testpop.go
CorgiMan: No, the time from faulty to tombstone is 24 * 60 * 60

scripts/testpop/testpop.go
CorgiMan: you can't multiply an int by a duration.
when you hav 5*time.Second, the literal 5 is automatically converted by the compiler

mennopruijssers:
Maybe make the descriptions of the flags a little bit more descriptive (e.g. "The period in ms a member stays a suspect period." for `--suspect-period`)

but besides that: LGTM.

mennopruijssers:
Owh, there is a linter error though: 
```
main.js: line 76, col 29, Missing semicolon. (W033)
```

CorgiMan:
Anything else?

Motiejus:
LGTM

PR #150, opened 2016-06-07
When creating the change for reincarnation use your self as the source.

thanodnl:
When a gossip was received that would overwrite the state of the local node we overwrite this change with a reincarnated version of our self. However we used the source of the original change as the source for this change making it harder for the reincarnated change to reach the node that declared it a suspect in the first place.

This change will use the source and new reincarnation number of the local node for this new change. This prevents the reincarnation change to be filtered out in pings to the original source causing convergence to speedup and potentially prevent fullsyncs from happening.

### DIFF
--- a/swim/memberlist.go
+++ b/swim/memberlist.go
@@ -336,11 +336,12 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 		// if change is local override, reassert member is alive
 		if member.localOverride(m.node.Address(), change) {
 			m.node.emit(RefuteUpdateEvent{})
+			newIncNo := nowInMillis(m.node.clock)
 			overrideChange := Change{
-				Source:            change.Source,
-				SourceIncarnation: change.SourceIncarnation,
+				Source:            m.node.Address(),
+				SourceIncarnation: newIncNo,
 				Address:           change.Address,
-				Incarnation:       nowInMillis(m.node.clock),
+				Incarnation:       newIncNo,
 				Status:            Alive,
 				Timestamp:         util.Timestamp(time.Now()),
 			}
--- a/swim/memberlist_test.go
+++ b/swim/memberlist_test.go
@@ -201,9 +201,12 @@ func (s *MemberlistTestSuite) TestMultipleUpdates() {
 }
 
 func (s *MemberlistTestSuite) TestUpdateTriggersReincarnation() {
+	source := "192.0.2.1:1234"
+	s.NotEqual(source, s.m.local.Address, "this test relies on the source and the target of the change to be different")
+
 	applied := s.m.Update([]Change{
 		Change{
-			Source:            "192.0.2.1:1234",
+			Source:            source,
 			SourceIncarnation: 1337,
 
 			Address:     s.m.local.Address,
--- a/swim/memberlist_test.go
+++ b/swim/memberlist_test.go
@@ -200,6 +200,28 @@ func (s *MemberlistTestSuite) TestMultipleUpdates() {
 	s.Equal(Leave, member.Status, "expected member to be leave")
 }
 
+func (s *MemberlistTestSuite) TestUpdateTriggersReincarnation() {
+	applied := s.m.Update([]Change{
+		Change{
+			Source:            "192.0.2.1:1234",
+			SourceIncarnation: 1337,
+
+			Address:     s.m.local.Address,
+			Incarnation: s.m.local.Incarnation,
+			Status:      Suspect,
+		},
+	})
+
+	s.Len(applied, 1, "expected change to be applied")
+
+	change := applied[0]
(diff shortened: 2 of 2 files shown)


swim/memberlist.go
```diff
@@ -336,11 +336,12 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 		// if change is local override, reassert member is alive
 		if member.localOverride(m.node.Address(), change) {
 			m.node.emit(RefuteUpdateEvent{})
+			newIncNo := nowInMillis(m.node.clock)
 			overrideChange := Change{
```
Motiejus: I think I prefer the [old node version][1] where the change was taken as-is and only the interesting fields got overwritten. If you agree with that, can we do the same here?

I just checked Go documentation, and couldn't find a way to have mixed typespecs, like in protobuffs*. Maybe there is a way that I don't know?

[1]: https://github.com/uber/ringpop-node/pull/282#discussion_r66072017
[2]: https://developers.google.com/protocol-buffers/docs/proto
[*]: From [their docs][2]: _However, the unknown fields are not discarded, and if the message is later serialized, the unknown fields are serialized along with it – so if the message is passed on to new code, the new fields are still available._

swim/memberlist_test.go
```diff
@@ -200,6 +200,28 @@ func (s *MemberlistTestSuite) TestMultipleUpdates() {
 	s.Equal(Leave, member.Status, "expected member to be leave")
 }
 
+func (s *MemberlistTestSuite) TestUpdateTriggersReincarnation() {
+	applied := s.m.Update([]Change{
+		Change{
+			Source:            "192.0.2.1:1234",
```
Motiejus: Like in the PR for Go, can we make sure this value does not equal `s.m.local.Address` to make sure the test is correct?

swim/memberlist.go
CorgiMan: If the change is a value and not a pointer we can:
```
override := change
overrideChange.Source = m.node.Address()
overrideChange.SourceIncNo = newIncNo
overrideChange.Status = Alive
overrideChange.Incarnation = newIncNo
```

CorgiMan:
Looks good after you have addressed Motiejus' comments. Do we test this in the integration tests?

swim/memberlist.go
thanodnl: @Motiejus The concern I have with copying the old change over constructing a new one is that this issue was actually cased by that exact case. In the nodejs version it used to extend the change it received, causing the source to not be overwritten and thus causing the change to take over the source address and incarnation number of the original change instead of putting its own information on there. I'd rather create a new change here as currently is done and run into more noticeable issues when we miss information from a change in here then have information fall through that should not have been copied.

If we want to go with the route of basing the change on the original change that came in we can do this with value types as well. If you assign a value type to a local variable it will copy all of its contents to the new struct and then allows us to update the fields we care about.

thanodnl: To add to it; this code would benefit a lot from a refactor where it does a reincarnation by setting the new incarnation value to a struct describing the state of the node somewhere in memory and from there construct a change that automatically adds the source information. I would even go as far as removing the source information completely from changes, but that will be a different discussion.

swim/memberlist_test.go
thanodnl: Good catch, I will make sure that `whoami` is a different value.

thanodnl:
@CorgiMan the integration test to test this is here: https://github.com/uber/ringpop-common/pull/69

swim/memberlist_test.go
```diff
@@ -201,9 +201,12 @@ func (s *MemberlistTestSuite) TestMultipleUpdates() {
 }
 
 func (s *MemberlistTestSuite) TestUpdateTriggersReincarnation() {
+	source := "192.0.2.1:1234"
+	s.NotEqual(source, s.m.local.Address, "this test relies on the source and the target of the change to be different")
```
mennopruijssers: Isn't is easier to use an "impossible" source to prevent flaky tests? (like: `355.1.1.1:1234` or `192.0.2.300:1234`)

mennopruijssers:
besides my small question: LGTM

swim/memberlist_test.go
thanodnl: addresses are not validated, so we can even use simple words. From what I understand happening here is that it does only create the swin state machine (`memberlist`) in memory and executes updates against it. But other tests also use ip addresses here. Although `127.0.0.1:xxxx` in this case, the reason to use `192.0.2.x:xxxx` here is that we settled on using TEST-NET-1 as specified in [rfc5735](https://tools.ietf.org/html/rfc5735) for our tests going forward.

Motiejus:
LGTM

PR #152, opened 2016-06-08
Emit lookup stat correctly

mennopruijssers:
Previously the lookup stat was emitted from the event-handler inside of `ringpop.go` when a `LookupEvent` was emitted. However, the lookup-event is emitted on ringpop itself so it's internal event-handler isn't triggered.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -58,7 +58,7 @@ type LookupEvent struct {
 // A LookupNEvent is sent when a lookupN is performed on the Ringpop's ring
 type LookupNEvent struct {
 	Key      string
-	N 		 int
+	N        int
 	Duration time.Duration
 }
 
--- a/events/events.go
+++ b/events/events.go
@@ -55,6 +55,13 @@ type LookupEvent struct {
 	Duration time.Duration
 }
 
+// A LookupNEvent is sent when a lookupN is performed on the Ringpop's ring
+type LookupNEvent struct {
+	Key      string
+	N 		 int
+	Duration time.Duration
+}
+
 // Ready is fired when ringpop has successfully bootstrapped and is ready to receive requests and other method calls.
 type Ready struct{}
 
--- a/ringpop.go
+++ b/ringpop.go
@@ -621,7 +621,7 @@ func (rp *Ringpop) LookupN(key string, n int) ([]string, error) {
 
 	rp.emit(events.LookupNEvent{
 		Key:      key,
-		N: n,
+		N:        n,
 		Duration: duration,
 	})
 
--- a/ringpop.go
+++ b/ringpop.go
@@ -484,9 +484,6 @@ func (rp *Ringpop) HandleEvent(event events.Event) {
 	case swim.JoinTriesUpdateEvent:
 		rp.statter.UpdateGauge(rp.getStatKey("join.retries"), nil, int64(event.Retries))
 
-	case events.LookupEvent:
-		rp.statter.RecordTimer(rp.getStatKey("lookup"), nil, event.Duration)
-
 	case swim.MakeNodeStatusEvent:
 		rp.statter.IncCounter(rp.getStatKey("make-"+event.Status), nil, 1)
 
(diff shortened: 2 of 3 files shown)


ringpop.go
```diff
@@ -612,7 +612,20 @@ func (rp *Ringpop) LookupN(key string, n int) ([]string, error) {
...
+	duration := time.Now().Sub(startTime)
+	rp.statter.RecordTimer(rp.getStatKey(fmt.Sprintf("lookupn.%d", n)), nil, duration)
+
+	if len(destinations) == 0 {
+		err := errors.New("could not find destinations for key")
+		rp.logger.WithField("key", key).Warn(err)
+		return destinations, err
+	}
+
+	return destinations, nil
```
Motiejus: Nice. Thank you for adding this.

ringpop.go
```diff
@@ -612,7 +612,20 @@ func (rp *Ringpop) LookupN(key string, n int) ([]string, error) {
...
+
+	destinations := rp.ring.LookupN(key, n)
+
+	duration := time.Now().Sub(startTime)
+	rp.statter.RecordTimer(rp.getStatKey(fmt.Sprintf("lookupn.%d", n)), nil, duration)
+
+	if len(destinations) == 0 {
+		err := errors.New("could not find destinations for key")
+		rp.logger.WithField("key", key).Warn(err)
+		return destinations, err
```
Motiejus: Could we somehow unit-test this branch too?

ringpop.go
```diff
@@ -612,7 +612,20 @@ func (rp *Ringpop) LookupN(key string, n int) ([]string, error) {
 	if !rp.Ready() {
 		return nil, ErrNotBootstrapped
 	}
-	return rp.ring.LookupN(key, n), nil
+	startTime := time.Now()
```
btromanova: Does it make sense to emit event from lookupN as well? We do it in Lookup.

btromanova:
Just one question about event from LookupN. Otherwise, lgtm. I've double checked that there are no other events that are processed in Ringpop HandleEvent and emitted form Ringpop itself.

ringpop.go
mennopruijssers: good catch!

ringpop.go
mennopruijssers: Both method (lookup/lookupN) are poorly tested. I added tests for both of them. All lines/branches are covered now :)

ringpop_test.go
```diff
@@ -577,6 +588,33 @@ func (s *RingpopTestSuite) TestLookupNNotReady() {
...
+	s.Error(err)
+}
+
+func (s *RingpopTestSuite) TestLookupN() {
+	createSingleNodeCluster(s.ringpop)
+
+	result, err := s.ringpop.LookupN("foo", 5)
+
+	s.Nil(err)
+	s.Equal(1, len(result), "LookupN returns not more results then number of nodes")
```
Motiejus: s/then/than

ringpop.go
```diff
@@ -619,6 +619,12 @@ func (rp *Ringpop) LookupN(key string, n int) ([]string, error) {
 	duration := time.Now().Sub(startTime)
 	rp.statter.RecordTimer(rp.getStatKey(fmt.Sprintf("lookupn.%d", n)), nil, duration)
 
+	rp.emit(events.LookupNEvent{
+		Key:      key,
+		N: n,
```
Motiejus: Indentation seems to be off, did you run gofmt?

Motiejus:
It's better to rebase on top of `origin/dev` rather than merge it to your branch. Could you do that?

events/events.go
```diff
@@ -55,6 +55,13 @@ type LookupEvent struct {
 	Duration time.Duration
 }
 
+// A LookupNEvent is sent when a lookupN is performed on the Ringpop's ring
+type LookupNEvent struct {
+	Key      string
+	N 		 int
```
Motiejus: Indentation seems to be off.. `gofmt`?

Motiejus:
LGTM

btromanova:
lgtm

PR #153, opened 2016-06-08
Fix double reincarnation

thanodnl:
This change will add a check for the incarnation number before reincarnating a node by incrementing its incarnation number. This reduces the number of gossips being sent in the case of reincarnation. Which will result in faster memberlist convergence.

### DIFF
--- a/swim/member.go
+++ b/swim/member.go
@@ -92,7 +92,9 @@ func (m *Member) nonLocalOverride(change Change) bool {
 	return statePrecedence(change.Status) > statePrecedence(m.Status)
 }
 
-// localOverride returns whether a change should be applied to to the member
+// localOverride returns whether the change will override the state of the local
+// member. When it will override the state the member should reincarnate itself
+// to make sure that other members see this node in a correct state.
 func (m *Member) localOverride(local string, change Change) bool {
 	if m.Address != local {
 		return false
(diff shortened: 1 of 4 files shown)


swim/member.go
```diff
@@ -92,11 +92,20 @@ func (m *Member) nonLocalOverride(change Change) bool {
...
+// member. When it will override the state the member should reincarnate itself
+// to make sure that other members see this node in a correct state.
 func (m *Member) localOverride(local string, change Change) bool {
 	if m.Address != local {
 		return false
 	}
+
+	// if the incarnation number of the change is smaller than the current
+	// incarnation number it is not overriding the change
+	if change.Incarnation < m.Incarnation {
```
mennopruijssers: if change.Incarnation is greater than m.Incarnation, it always overrides right? not only when change.status is one of faulty, suspect or tombstone. (should not happen but might be nice to add as an extra check)

test/run-integration-tests
```diff
@@ -6,7 +6,7 @@ set -eo pipefail
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="fix/double-reincarnation"
```
mennopruijssers: Change back before merging (and after the related PR in common is merged) ;)

mennopruijssers:
lgtm

Motiejus:
LGTM

CorgiMan:
This change is great for the failure testing!

PR #157, opened 2016-07-01
Test examples/ping-*/README.md

Motiejus:
Using [cram][1] to execute and verify instructions in the following files:

* `examples/ping-json/README.md`
* `examples/ping-thrift-gen/README.md`
* `examples/ping-thrift/README.md`

Currently, all of the tests are failing due to incomplete instructions in the README, but opening the pull  request anyway to set the path forward to fixing them.

[1]: https://pypi.python.org/pypi/cram

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@ go:
   - 1.6
   - 1.5
 
+sudo: false
+addons:
+  apt:
+    packages:
+    - python-virtualenv
+
 install:
   - go get -u github.com/Masterminds/glide
   - go get github.com/axw/gocov/gocov
@@ -11,6 +17,8 @@ install:
   - go get github.com/golang/lint/golint
   - ./scripts/travis/get-thrift.sh
   - ./scripts/travis/get-thrift-gen.sh
+  - ./scripts/travis/get-cram.sh
+  - npm install -g tcurl
   - make setup
 
 env:
@@ -18,6 +26,7 @@ env:
   - RUN="make test-unit"
   - RUN="make test-integration"
   - RUN="make test-race"
+  - RUN="make test-examples"
 
 matrix:
   allow_failures:
@@ -33,3 +42,4 @@ after_success:
 cache:
   directories:
     - $HOME/.glide/cache
+    - _venv
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ go:
   - 1.6
   - 1.5
 
-sudo: false
 addons:
   apt:
     packages:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -3,6 +3,12 @@ go:
   - 1.6
   - 1.5
 
+sudo: false
```
dansimau: Why is `sudo: false` required? IIRC, this is the default

Motiejus: Leftovers from development process. Good catch, removed.

dansimau:
LGTM, happy to merge, though I do have this philosophical thought:

Is storing the tests in the README actually a good idea? I know it seems cool because it's less files or whatever, but it is sane/obvious/intuitive?

Note we have sacrificed formatting in the README to do this:
https://github.com/uber/ringpop-go/blob/32aed6541ce6ae53100d81f82a47ce7fe346e90a/examples/ping-json/README.md#running-the-example

I don't feel strongly enough about this to block the merge of this PR. Practically, I don't think this will cause us a huge issue/effort in the future. But I do think tests in the README is a bit strange and may increase the average WTFs/min in the codebase.

Motiejus:
@dansimau thanks for the comment, and this is indeed a good philosophical discussion.

If we look at what we want to achieve:

* Have clear instructions on how to build and use the project.
* Make sure the instructions are sane (= automatically tested).

I see two approaches:

1. *More common:* separate the documentation generation and testing. Usually deliverable is a docbook- or a sphinx-like documentation artifact, which, in the back, nobody cares how it's generated.
2. *Less common:* test the actual final documentation.
3. Move the steps from README to a script, and test the script.

(1) is a standard practice. However, since github does not really allow it easily (I want the README to be readable as-is).
(2) gives more wtfs/minute, but it will work, as long as people keep adding the correct examples.
(3) seems to be least wtfs/minute, but leads to a least readable README.

Motiejus:
cc @severb can you check out the examples?

test/run-example-tests
```diff
@@ -0,0 +1,105 @@
...
+#
+# Runs the ringpop-admin tests.
+#
+# This script launches a ringpop cluster using tick-cluster and then runs
+# a series of tests against the ringpop-admin command output.
+#
+set -euo pipefail
+
+
+declare cram_opts="-v --shell=/bin/bash"
```
thanodnl: I don't like the formatting of the README.md files now. Due to only two spaces they are hard to read on github. if we add `--indent=4` to the cram options we can indent the code blocks by 4 spaces which will put a box around them in markdown making it a bit easier to read IMHO.

thanodnl:
LGTM

Motiejus:
@dansimau we actually fixed the formatting in README. Can you re-check?

dansimau:
LGTM

PR #158, opened 2016-07-01
Improve cache by not storing ORIG_HEAD

Motiejus:
`glide install` step touches `ORIG_HEAD` of some dependencies, causing it to update the whole cache on every run. If we remove it, the cache will not need to be updated after every execution, speeding all tests.

Tests here:
- https://travis-ci.org/uber/ringpop-go/jobs/141601533 (initial run, filling cache)
- https://travis-ci.org/uber/ringpop-go/jobs/141608840 (subsequent run)

At the end of the second run, observe "store build cache.. nothing changed, not updating cache".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,6 @@ after_success:
 cache:
   directories:
     - $HOME/.glide/cache
+
+before_cache:
+  - find $HOME/.glide/cache -name ORIG_HEAD -exec rm {} \;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,3 +33,6 @@ after_success:
 cache:
   directories:
     - $HOME/.glide/cache
+
+before_cache:
+  - find $HOME/.glide/cache -name ORIG_HEAD -exec rm {} \;
```
dansimau: Awesome optimisation! Might be a good idea to leave a comment explaining what it's doing

dansimau:
LGTM

thanodnl:
LGTM

PR #159, opened 2016-07-25
Status refactor 1: Refactor membership to use localMember as the source of truth

thanodnl:
With this refactor we keep an explicit copy of what the local state of a member is and reincarnate from there if we receive a gossip that would overwrite the state of our node in the network.

It is analogous to https://github.com/uber/ringpop-node/pull/291

### DIFF
--- a/swim/heal_partition.go
+++ b/swim/heal_partition.go
@@ -72,19 +72,35 @@ func nodesThatNeedToReincarnate(MA, MB []Change) (changesForA, changesForB []Cha
 		// If a node would be overwritten and would stop being pingable, it is
 		// not suited for merging.
 		if b.isPingable() && a.overrides(b) && !a.isPingable() {
-			changesForB = append(changesForB, Change{
-				Address:     a.Address,
-				Incarnation: a.Incarnation,
-				Status:      Suspect,
-			})
+			// take a look at the state of the member
+			member := Member{}
+			member.populateFromChange(&a)
+
+			// create the member into a change
+			change := Change{}
+			change.populateSubject(&member)
+
+			// gossip the suspect status
+			change.Status = Suspect
+
+			// record the change to be sent to B
+			changesForB = append(changesForB, change)
 		}
 
 		if a.isPingable() && b.overrides(a) && !b.isPingable() {
-			changesForA = append(changesForA, Change{
-				Address:     b.Address,
-				Incarnation: b.Incarnation,
-				Status:      Suspect,
-			})
+			// take a look at the state of the member
+			member := Member{}
+			member.populateFromChange(&b)
+
+			// create the member into a change
+			change := Change{}
+			change.populateSubject(&member)
+
+			// gossip the suspect status
+			change.Status = Suspect
+
+			// record the change to be sent to A
+			changesForA = append(changesForA, change)
 		}
 	}
 
--- a/swim/heal_partition.go
+++ b/swim/heal_partition.go
@@ -69,8 +69,6 @@ func nodesThatNeedToReincarnate(MA, MB []Change) (changesForA, changesForB []Cha
 			continue
 		}
 
-		// TODO: DRY
-
 		// If a node would be overwritten and would stop being pingable, it is
 		// not suited for merging.
 		if b.isPingable() && a.overrides(b) && !a.isPingable() {
--- a/swim/heal_partition.go
+++ b/swim/heal_partition.go
@@ -69,16 +69,21 @@ func nodesThatNeedToReincarnate(MA, MB []Change) (changesForA, changesForB []Cha
 			continue
 		}
 
+		// TODO: DRY
+
 		// If a node would be overwritten and would stop being pingable, it is
 		// not suited for merging.
 		if b.isPingable() && a.overrides(b) && !a.isPingable() {
-			// take a look at the state of the member
-			member := Member{}
-			member.populateFromChange(&a)
+			// take the change in a local variable, this protects to inadvertly
+			// changing the type of `a` from a value type to a pointer type.
+			var change Change
+			change = a
 
-			// create the member into a change
-			change := Change{}
-			change.populateSubject(&member)
+			// Remove the source information from the change. If the source
+			// information is present and it is gossiped to the other partition
+			// it might cause the partitions to heal before they are in a safe
+			// state.
+			change.scrubSource()
 
 			// gossip the suspect status
 			change.Status = Suspect
@@ -88,13 +93,16 @@ func nodesThatNeedToReincarnate(MA, MB []Change) (changesForA, changesForB []Cha
 		}
 
 		if a.isPingable() && b.overrides(a) && !b.isPingable() {
-			// take a look at the state of the member
-			member := Member{}
-			member.populateFromChange(&b)
+			// take the change in a local variable, this protects to inadvertly
+			// changing the type of `a` from a value type to a pointer type.
+			var change Change
+			change = b
 
-			// create the member into a change
-			change := Change{}
-			change.populateSubject(&member)
+			// Remove the source information from the change. If the source
+			// information is present and it is gossiped to the other partition
+			// it might cause the partitions to heal before they are in a safe
+			// state.
+			change.scrubSource()
 
 			// gossip the suspect status
 			change.Status = Suspect
--- a/swim/member.go
+++ b/swim/member.go
@@ -22,8 +22,8 @@ package swim
 
 import (
 	"bytes"
-	"fmt"
 	"math/rand"
+	"strconv"
 
 	"github.com/uber/ringpop-go/util"
 )
@@ -70,7 +70,9 @@ func (m *Member) populateFromChange(c *Change) {
 // checksumString fills a buffer that is passed with the contents that this node
 // needs to add to the checksum string.
 func (m Member) checksumString(b *bytes.Buffer) {
-	fmt.Fprintf(b, "%s%s%v", m.Address, m.Status, m.Incarnation)
+	b.WriteString(m.Address)
+	b.WriteString(m.Status)
+	b.WriteString(strconv.FormatInt(m.Incarnation, 10))
 }
 
 // shuffles slice of members pseudo-randomly, returns new slice
--- a/swim/member.go
+++ b/swim/member.go
@@ -87,9 +87,10 @@ func shuffle(members []*Member) []*Member {
 	return newMembers
 }
 
-// shouldProcessGossip evaluates the rules of swim and returns wether the gossip
-// should be processed. eg. Copy the memberstate of the gossip to the known
-// memberstate in the memberlist (creating the member when is does not exist).
+// shouldProcessGossip evaluates the rules of swim and returns whether the
+// gossip should be processed. eg. Copy the memberstate of the gossip to the
+// known memberstate in the memberlist (creating the member when is does not
+// exist).
 func shouldProcessGossip(old *Member, gossip *Member) bool {
 	// tombstones will not be accepted if we have no knowledge about the member
 	if gossip.Status == Tombstone && old == nil {
--- a/swim/member.go
+++ b/swim/member.go
@@ -206,6 +206,11 @@ func (c *Change) populateSource(m *Member) {
 	c.SourceIncarnation = m.Incarnation
 }
 
+func (c *Change) scrubSource() {
+	c.Source = ""
+	c.SourceIncarnation = 0
+}
+
 // suspect interface
 func (c Change) address() string {
 	return c.Address
--- a/swim/member.go
+++ b/swim/member.go
@@ -21,8 +21,9 @@
 package swim
 
 import (
+	"bytes"
+	"fmt"
 	"math/rand"
-	"sync"
 
 	"github.com/uber/ringpop-go/util"
 )
@@ -46,7 +47,6 @@ const (
 
 // A Member is a member in the member list
 type Member struct {
-	sync.RWMutex
 	Address     string `json:"address"`
 	Status      string `json:"status"`
 	Incarnation int64  `json:"incarnationNumber"`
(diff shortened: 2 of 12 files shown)


swim/heal_partition.go
```diff
@@ -72,19 +72,35 @@ func nodesThatNeedToReincarnate(MA, MB []Change) (changesForA, changesForB []Cha
...
-				Incarnation: a.Incarnation,
-				Status:      Suspect,
-			})
+			// take a look at the state of the member
+			member := Member{}
+			member.populateFromChange(&a)
+
+			// create the member into a change
+			change := Change{}
+			change.populateSubject(&member)
```
mennopruijssers: It might be nicer/more readable to just copy the existing change (`a`) and change the status to suspect or have a `change.populateFromChange`-method.

swim/member.go
```diff
@@ -61,6 +61,18 @@ func (m Member) incarnation() int64 {
...
+func (m *Member) populateFromChange(c *Change) {
+	m.Address = c.Address
+	m.Incarnation = c.Incarnation
+	m.Status = c.Status
+}
+
+// checksumString fills a buffer that is passed with the contents that this node
+// needs to add to the checksum string.
+func (m Member) checksumString(b *bytes.Buffer) {
+	fmt.Fprintf(b, "%s%s%v", m.Address, m.Status, m.Incarnation)
```
mennopruijssers: - since the formatting is just an append of the three fields, why not write directly to the buffer?
- `b` could be an `io.Writer`

swim/member.go
```diff
@@ -73,40 +85,49 @@ func shuffle(members []*Member) []*Member {
...
 
-// nonLocalOverride returns wether a change should be applied to the member.
-// This function assumes that the address of the member and the change are
-// equal.
-func (m *Member) nonLocalOverride(change Change) bool {
-	// change is younger than current member
-	if change.Incarnation > m.Incarnation {
+// acceptGossip evaluates the rules of swim to accept the gossip as the new state
+// of the member.
+func acceptGossip(old *Member, gossip *Member) bool {
```
mennopruijssers: I'm not sure if this name covers the intention: the gossip won't be accepted "as is" if it's overwriting the local member. This method just checks if the incoming gossip should be processed or could be ignored right?

swim/member.go
```diff
@@ -166,6 +187,23 @@ func (c Change) validateOutgoing() Change {
 	return c
 }
 
+func (c *Change) populateSubject(m *Member) {
+	if m == nil {
+		return
```
mennopruijssers: or reset to default values?

swim/member.go
```diff
@@ -166,6 +187,23 @@ func (c Change) validateOutgoing() Change {
 	return c
 }
 
+func (c *Change) populateSubject(m *Member) {
```
mennopruijssers: The `populateSubject` and `populateSource` methods feel a little bit like a constructor. Is it possible/feasible to add a constructor method (e.g. `newChange(subject, source *Member)`)?

swim/member.go
```diff
@@ -166,6 +187,23 @@ func (c Change) validateOutgoing() Change {
...
+
+func (c *Change) populateSource(m *Member) {
+	if m == nil {
+		return
+	}
+	c.Source = m.Address
+	c.SourceIncarnation = m.Incarnation
+}
+
 // suspect interface
```
mennopruijssers: (not related to your diff but this comment doesn't make sense?)

swim/member_test.go
```diff
@@ -74,52 +75,6 @@ func newChange(addr string, s state) Change {
 	}
 }
 
-func (s *MemberTestSuite) TestNonLocalOverride() {
```
mennopruijssers: Maybe replace with a test for `acceptGossip`?

swim/member_test.go
```diff
@@ -139,3 +94,16 @@ func TestChangeOmitTombstone(t *testing.T) {
...
+	member := Member{
+		Address:     "192.168.2.1:1234",
+		Status:      Alive,
+		Incarnation: 42,
+	}
+
+	var b bytes.Buffer
+	member.checksumString(&b)
+
+	assert.Equal(t, "192.168.2.1:1234alive42", b.String(), "member checksum serialization failed")
```
mennopruijssers: Normally the message of an assert describes the expect behaviour.

swim/memberlist.go
```diff
@@ -250,11 +277,47 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
...
 }
 
+// TODO: MakeLeave is only during testing. Need to figure out if we want to keep
+// it around or rewrite the tests.
 func (m *memberlist) MakeLeave(address string, incarnation int64) []Change {
+	// for backwards compatibility in tests we allow to update the local member
+	// in place when the gossip is about the local node and an incarnation number
+	// that is expected to have accept the update.
+	if address == m.local.Address && m.local.Incarnation <= incarnation {
+		m.local.Status = Leave
```
mennopruijssers: Maybe change to `m.SetLocalStatus` so the checksum is also correctly updated?

swim/memberlist.go
```diff
@@ -320,50 +387,52 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
...
+				// if the member was not already present in the list we will it
+				// needs to be added and randomly inserted in the list to ensure
+				// guarantees for pinging
+				if !has {
+					m.members.byAddress[gossip.Address] = &gossip
+					i := m.getJoinPosition()
+					m.members.list = append(m.members.list[:i], append([]*Member{&gossip}, m.members.list[i:]...)...)
+				} else {
+					// copy the state of the gossip to the member
+					*member = gossip
```
mennopruijssers: It might improve understandability if you change this to `m.members.byAddress[gossip.Address] = &gossip` or make it more clear with the comment that that's what you're basically doing.

swim/memberlist.go
```diff
@@ -105,20 +113,24 @@ func (m *memberlist) ComputeChecksum() {
 // generates string to use when computing checksum
 func (m *memberlist) GenChecksumString() string {
 	var strings sort.StringSlice
+	var buffer bytes.Buffer
 
 	for _, member := range m.members.list {
```
mennopruijssers: Missing readlock

swim/memberlist.go
```diff
@@ -414,40 +483,6 @@ func (m *memberlist) getJoinPosition() int {
 	return rand.Intn(l)
```
mennopruijssers: Document that `getJoinPosition` needs the mutex to be (read-)locked

mennopruijssers:
Other "issues" found when checking the locks:
[ ] `memberlist.go#200 NumPingableMembers()`: can use readlock instead of a full lock.
[ ] `memberlist.go#234 GetMembers()`: `members` could be initialized with the right size to prevent array resizing
[ ] `memberlist.go#504 GetReachableMembers()`: `active` could be initialized with a default size of the members (most of them will be reachable) to prevent array resizing.

swim/heal_partition.go
thanodnl: You can't copy the change here because it is essential for the working of partition healing that the source information of the change is not set on a change sent to the other partition. If you add the source information on there you have the chance of a reverse full sync that will merge the partitions before they are in a merge-able state. The same goes for populating a change from a change, it is debatable if it should copy the source information of a change.

We can do two things:

1. document better why we go change->member->change
2. have a `change.scrubSourceInformation()` to explicitly remove the source information from the change here.

swim/member.go
thanodnl: I like the idea. Both `Address` and `Status` could written directly to the `Buffer` as it has a `WriteString` method (`io.Writer` only accepts `byte[]`). But `Incarnation` is an `int64` which can not be written directly. But we could use `strconv.Itoa()` to translate the number into a string.

swim/member.go
thanodnl: I'd rather have a separate function that can set them back to the default values as described above. Adding it here feels a bit hacky and might be hard to understand when reading the code from the caller site.

swim/member.go
thanodnl: I have been thinking about doing it in a constructor but this gives a bit more control when a change needs to be constructed which information is being copied to a change. I will revisit it now that the all the call sites are known.

swim/member.go
thanodnl: the comment is hard to understand, but the function here is needed by the state transitions component which defined/defines a suspect interface from the time it was called suspect timer. Don't think I want to change it as part of this PR but we can open a new PR to update the comments on these functions to be more explaining.

swim/member_test.go
thanodnl: YES! I totally thought that I had a test for that function, but it is such a crucial part of the protocol all paths should be tested for `acceptGossip`. Thanks!

swim/member_test.go
thanodnl: True, but in retrospect I will remove this test as it is the exact same test as the doc_test above. Doesn't feel right to test the same thing twice.

swim/memberlist.go
thanodnl:
```bash
$ git grep "GenChecksumString"
swim/memberlist.go:     checksum := farm.Fingerprint32([]byte(m.GenChecksumString()))
swim/memberlist.go:func (m *memberlist) GenChecksumString() string {
```

Since it is only called from `ComputeChecksum` I will un-export this function.

swim/memberlist.go
thanodnl: Good note about the checksum being out of date after calling this. However calling `SetLocalStatus` also reincarnates the node by design which had a side effect on the tests where this function is being used. To be more clear, this function is _ONLY_ being used in tests at the moment.

I'm not happy that we still have this function in the membership anyway. I will have a second look in what it takes to remove this function.

swim/memberlist.go
thanodnl: It does the inverse of the code snippet you posted. It copies the values of the gossip to the structure that is already stored in the list and the map. I tried to make that clear in the comment above but I will add `by value instead of by reference` to the comment.

swim/memberlist.go
```diff
@@ -232,7 +244,22 @@ func (m *memberlist) GetMembers() (members []Member) {
 // Reincarnate sets the status of the node to Alive and updates the incarnation
 // number. It adds the change to the disseminator as well.
 func (m *memberlist) Reincarnate() []Change {
-	return m.MakeAlive(m.node.address, nowInMillis(m.node.clock))
+	return m.MakeAlive(m.local.Address, nowInMillis(m.node.clock))
```
mennopruijssers: `SetLocalState('alive')`?

swim/memberlist.go
```diff
@@ -232,7 +244,22 @@ func (m *memberlist) GetMembers() (members []Member) {
...
+}
+
+// creates a change that reincarnates the local node. It does not change the
+// state of the node but rather reasserts its state by bumping its incarnation
+// number.
+func (m *memberlist) reincarnationChange() Change {
```
mennopruijssers: This method does more than just generating a change: it also bumps the incarnation number.
Maybe rename it to `bumpIncarnation` and call it from
- reincarnate
- PostLocalUpdate?

swim/memberlist.go
thanodnl: Since this function is only being used in tests I will un-export the function and document that it is only used in tests. I'm even inclined to move it to the test file but I'm not completely sure about that.

Then in a subsequent PR I will try to remove all obsolete `Make` functions (eg. `MakeLeave` and `MakeAlive`) since only `MakeSuspect`, `MakeFaulty` and `MakeTombstone` are allowed to be used by any given node. All other states can only be declared by the node its self and therefore should be routed via `SetLocalStatus`.

swim/memberlist.go
```diff
@@ -233,6 +233,7 @@ func (m *memberlist) RandomPingableMembers(n int, excluding map[string]bool) []*
 // returns an immutable slice of members representing the current state of the membership
 func (m *memberlist) GetMembers() (members []Member) {
 	m.members.RLock()
+	members = make([]Member, 0, len(m.members.list))
 	for _, member := range m.members.list {
```
mennopruijssers: or `copy(members, m.members.list)`?

swim/member.go
```diff
@@ -204,6 +206,11 @@ func (c *Change) populateSource(m *Member) {
 	c.SourceIncarnation = m.Incarnation
 }
 
+func (c *Change) scrubSource() {
```
mennopruijssers: What if we populate more "source" fields in the future? Currently there's no "safeguard" against forgetting to clear it here. Maybe at least add a comment to `populateSource`?
Bonus points for a test that:
- creates a change -> `popuplateSource` -> `scrubSource` -> compare with an new/empty change.

swim/memberlist.go
```diff
@@ -111,7 +111,7 @@ func (m *memberlist) ComputeChecksum() {
 }
 
 // generates string to use when computing checksum
-func (m *memberlist) GenChecksumString() string {
+func (m *memberlist) genChecksumString() string {
```
mennopruijssers: nice!

swim/memberlist.go
thanodnl: There is two reasons:

1. The two slices are of different types, the original slice is `[]*Member` which is a pointer to a struct, the returning slice is `[]Member` which puts the struct in place in the slice. This makes sure that the receiver of the slice is not able to change the members kept in memory of `memberlist`.
2. In the future a predicate will be passed into this function to return a filtered list of members eg. based on labels.

swim/memberlist.go
thanodnl: There is a PR ( #161 ) to remove this code.

mennopruijssers:
LGTM!

mennopruijssers:
Tests are failing though..

swim/heal_partition.go
```diff
@@ -69,22 +69,46 @@ func nodesThatNeedToReincarnate(MA, MB []Change) (changesForA, changesForB []Cha
 			continue
 		}
 
+		// TODO: DRY
```
Motiejus: I think if we want to refactor the two blocks below, the best time to do it is now. Or agree we will not, and remove this comment.

I think it's good as-is.

swim/member.go
```diff
@@ -73,40 +87,50 @@ func shuffle(members []*Member) []*Member {
 	return newMembers
 }
 
-// nonLocalOverride returns wether a change should be applied to the member.
-// This function assumes that the address of the member and the change are
-// equal.
-func (m *Member) nonLocalOverride(change Change) bool {
-	// change is younger than current member
-	if change.Incarnation > m.Incarnation {
+// shouldProcessGossip evaluates the rules of swim and returns wether the gossip
```
Motiejus: s/wether/whether

Motiejus:
LGTM

PR #161, opened 2016-08-01
Remove MakeLeave as a function from the membership

thanodnl:
Remove `MakeLeave` function as it is only used during testing to get the membership into a certain state, instead we will use Update with crafted changes for this as that is the most public interface to interact with the membership.

_NOTE: This PR is on top of #159, its only purpose is to remove `MakeLeave` from the memberlist_

/cc @mennopruijssers


mennopruijssers:
Awesome!

Motiejus:
LGTM

PR #162, opened 2016-08-01
Support imported files in thrift-gen

thanodnl:
This PR adds support for imported thrift files during the code generation phase.

`GoUnusedProtection__` is the same trick as used in [`thrift-gen`](https://github.com/uber/tchannel-go/blob/v1.0.9/thrift/thrift-gen/tchannel-template.go#L22)

### DIFF
--- a/test/remoteservice/remoteservice.thrift
+++ b/test/remoteservice/remoteservice.thrift
@@ -7,5 +7,5 @@ include "./shared.thrift"
 include "./unused.thrift"
 
 service RemoteService {
-  void RemoteCall(1: shared.UUID id)
+  void RemoteCall(1: shared.Name name)
 }
--- a/test/remoteservice/remoteservice.thrift
+++ b/test/remoteservice/remoteservice.thrift
@@ -1,3 +1,5 @@
+include "./shared.thrift"
+
 service RemoteService {
-  void RemoteCall(1: string name)
+  void RemoteCall(1: shared.UUID id)
 }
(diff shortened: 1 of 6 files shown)


test/remoteservice/remoteservice.thrift
```diff
@@ -1,3 +1,11 @@
...
+
+// the unused file contains types that are not used in service endpoints. These
+// types might actually be used in embedded structs here, but since the
+// generated code could potentially contain the import of an unused package in
+// the RingpopAdapter. To prevent this the generator uses `GoUnusedProtection__`
+include "./unused.thrift"
+
 service RemoteService {
-  void RemoteCall(1: string name)
+  void RemoteCall(1: shared.UUID id)
```
mennopruijssers: Why change it from `name` to `id`? To have less changes, you could keep it name and define a `Name` type in shared.thrift (`typedef string Name`)?

mennopruijssers:
I'm fine with changing the name to an uuid although it implies more changes.
Feel free to change it or keep it like this (LGTM)

mennopruijssers:
🎉

thanodnl:
The failure is a flap:

```
--- FAIL: TestProtocolStats (5.09s)
	Error Trace:	stats_test.go:89
	Error:		Should not be zero, but was 0
		
	Error Trace:	stats_test.go:90
	Error:		Should not be zero, but was 0
		
--- FAIL: TestStatsTestSuite (5.21s)
FAIL
FAIL	github.com/uber/ringpop-go/swim	9.984s
```

Nothing changed that would impact a failure in `swim` with this change. Re-running test.

test/remoteservice/unused.thrift
```diff
@@ -0,0 +1 @@
+typedef string UnusedType
```
CorgiMan: This isn't used anywhere, what is it for?

CorgiMan:
Looks good

PR #165, opened 2016-08-09
Use narrower duck-typing instead of embedding whole interface

yurishkuro:
The `tchannel.Registrar` interface has been extended with an extra method, and there are compile errors in our internal go-common library against ringpop-go/0.6.0

ringpop does not need the whole `tchannel.Registrar` interface in its duck-typed TChannel  interface, just the Register() method.

### DIFF
--- a/shared/interfaces.go
+++ b/shared/interfaces.go
@@ -4,7 +4,7 @@ import "github.com/uber/tchannel-go"
 
 // The TChannel interface defines the dependencies for TChannel in Ringpop.
 type TChannel interface {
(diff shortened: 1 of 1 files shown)


dansimau:
LGTM 👍

yurishkuro:
Should this merge to master or to dev?

mennopruijssers:
> Should this merge to master or to dev?

`Dev`. However, the default branch on github is `master` because `go get`, `glide` and other tooling is using the default branch.
Sorry for the inconvenience.

PR #166, opened 2016-08-09
Use narrower duck-typing instead of embedding whole interface

yurishkuro:
Recreating #165 to merge to `dev` instead of `master`

---
The `tchannel.Registrar` interface has been extended with an extra method, and there are compile errors in our internal go-common library against ringpop-go/0.6.0

ringpop does not need the whole `tchannel.Registrar` interface in its duck-typed TChannel  interface, just the Register() method.


dansimau:
LGTM 👍

PR #167, opened 2016-08-09
Feature: Labels

thanodnl:
This PR will add a possibility to add labels to a node in ringpop. Labels set on a node will be gossiped around the network in the same fashion as updates to the status of a node do and therefore will converge so all members of the membership will have the same labels for all nodes during normal operation where the membership is converged.

Labels can be used to annotate nodes with extra information. In the examples an application is provided where a role for a node is set on its labels. Other members of the membership can then query ringpop for all nodes that fulfill a specific role.

- [x] Add test to make sure setting private labels from public interface returns error (from: [#172](https://github.com/uber/ringpop-go/pull/172#issuecomment-239798337))
- [x] Revisit locking to prevent races


swim/member.go
```diff
@@ -73,6 +80,55 @@ func (m Member) checksumString(b *bytes.Buffer) {
...
+// labels are set, but 0 does not indicate that no labels are set. It could be
+// possible that 0 is computed as the checksum.
+func (l LabelMap) checksum() int32 {
+	var lb bytes.Buffer
+	var checksum uint32
+	for key, value := range l {
+		// decide if we want to escape both fields to enforce uniqueness of the
+		// key-value checksum.
+		lb.WriteString("/")
+		lb.WriteString(key)
```
thanodnl: TODO need to change this function to safely escape these values to prevent collisions which will be negated due to the xor operation later.

swim/member_predicate.go
```diff
@@ -0,0 +1,43 @@
...
+func TestMember(member Member, predicates ...MemberPredicate) bool {
+	for _, p := range predicates {
+		if !p(member) {
+			return false
+		}
+	}
+	return true
+}
+
+// ReachableMember tests if a member is deemed to be reachable. This filters out
```
thanodnl: unexported function, need to update the doc.

swim/memberlist.go
```diff
@@ -487,25 +566,14 @@ func (m *memberlist) Iter() *memberlistIter {
...
-		}
-	}
-	m.members.RUnlock()
-
-	return active
-}
-
-func (m *memberlist) CountReachableMembers() int {
+// CountMembers returns the number of members maintained by the swim membership
+// protocol
```
thanodnl: Add: for all members that match the predicates

test/lib.sh
```diff
@@ -2,7 +2,7 @@
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="feature/labels"
```
thanodnl: land this branch before merging this.

examples/role-labels/gen-go/role/ringpop-role.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if c.SetRole != nil {
+		if c.SetRole.Key == nil {
+			return errors.New("configuration for endpoint SetRole is missing a Key function")
+		}
+	}
+	return nil
+}
+
+// NewRingpopRoleServiceAdapter creates an implementation of the TChanRoleService interface. This specific implementation will use to configuration provided during construction to deterministically route calls to nodes from a ringpop cluster. The channel should be the channel on which the service exposes its endpoints. Forwarded calls, calls to unconfigured endpoints and calls that already were executed on the right machine will be passed on the the implementation passed in during construction.
```
mennopruijssers: Maybe add some line breaks to make this a little bit easier to read?

examples/role-labels/main.go
```diff
@@ -0,0 +1,126 @@
...
+	if _, err := worker.ringpop.Bootstrap(opts); err != nil {
+		log.Fatalf("ringpop bootstrap failed: %v", err)
+	}
+
+	log.Println("Bootstrapped")
+
+	if labels, err := worker.ringpop.Labels(); err != nil {
+		log.Fatalf("unable to get access to ringpop labels: %v", err)
+	} else {
+		labels.Set(roleKey, *rolename)
```
mennopruijssers: Is it possible to `Set` labels before bootstrapping?

ringpop.go
```diff
@@ -646,21 +648,28 @@ func (rp *Ringpop) ringEvent(e interface{}) {
...
+// active membership list.
+func (rp *Ringpop) GetReachableMembers(predicates ...swim.MemberPredicate) ([]string, error) {
 	if !rp.Ready() {
 		return nil, ErrNotBootstrapped
 	}
-	return rp.node.GetReachableMembers(), nil
+
+	members := rp.node.GetReachableMembers(predicates...)
+
+	addresses := make([]string, 0, len(members))
```
mennopruijssers: 👍

ringpop.go
```diff
@@ -646,21 +648,28 @@ func (rp *Ringpop) ringEvent(e interface{}) {
...
+	for _, member := range members {
+		addresses = append(addresses, member.Address)
+	}
+	return addresses, nil
 }
 
 // CountReachableMembers returns the number of members currently in this
-// instance's membership list that aren't faulty.
```
mennopruijssers: maybe change this to the more abstract `.. list that have a status that makes them reachable (e.g. not faulty)`

ringpop.go
```diff
@@ -646,21 +648,28 @@ func (rp *Ringpop) ringEvent(e interface{}) {
...
-func (rp *Ringpop) CountReachableMembers() (int, error) {
+// instance's active membership list.
```
mennopruijssers: Maybe add "... matching the provided predicates"?

ringpop.go
```diff
@@ -646,21 +648,28 @@ func (rp *Ringpop) ringEvent(e interface{}) {
 }
 
 // GetReachableMembers returns a slice of members currently in this instance's
-// membership list that aren't faulty.
-func (rp *Ringpop) GetReachableMembers() ([]string, error) {
+// active membership list.
```
mennopruijssers: + predicates? (See comment on `CountReachableMembers`)

swim/handlers.go
```diff
@@ -147,6 +154,11 @@ func (n *Node) adminLeaveHandler(ctx json.Context, req *emptyArg) (*Status, erro
 	return &Status{Status: "ok"}, nil
 }
 
+func (n *Node) adminSetLabelHandler(ctx json.Context, req *SetLabelRequest) (*Status, error) {
+	n.Labels().Set(req.Key, req.Value)
```
mennopruijssers: No error handling? What if the key or value is invalid?

swim/node.go
```diff
@@ -517,14 +519,23 @@ func (n *Node) pingNextMember() {
...
 }
 
-// GetReachableMembers returns a slice of members currently in this node's
-// membership list that aren't faulty.
-func (n *Node) GetReachableMembers() []string {
-	return n.memberlist.GetReachableMembers()
+// GetReachableMembers returns a slice of members containing only the reachable
+// members that satisfies the predicates passed in.
+func (n *Node) GetReachableMembers(predicates ...MemberPredicate) []Member {
+	predicates = append(predicates, reachableMember)
```
mennopruijssers: Might be preferred to "prepend" the predicates with `reachableMember`. We know that `reachableMember` is only doing a simple status-check; who knows what others are passing in as a predicate? ¯\_(ツ)_/¯

swim/member_predicate.go
```diff
@@ -0,0 +1,43 @@
...
+// all members that are known to be unresponsive. Most operations will only ever
+// be concerned with Members that are in a Reachable state. In SWIM terms a
+// member is considered reachable when it is either in Alive status or in
+// Suspect status. All other Members are considered to not be reachable.
+func reachableMember(member Member) bool {
```
mennopruijssers: Naming: I'll prefer `memberIsReachable`; what do you think?

swim/member_predicate.go
```diff
@@ -0,0 +1,43 @@
...
+	return member.isReachable()
+}
+
+// LabeledMember returns a predicate able to test if the value of a label on a
+// member is equal to the provided value.
+func LabeledMember(key, value string) MemberPredicate {
```
mennopruijssers: Same here: `MemberWithLabel`?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
 	m.postLocalUpdate()
 }
 
+func (m *memberlist) SetLocalLabel(key, value string) error {
+	// TODO implement a sane limit for the size of the labels to prevent users
```
mennopruijssers: TODO? :)

swim/member.go
```diff
@@ -73,6 +80,55 @@ func (m Member) checksumString(b *bytes.Buffer) {
...
+
+		checksum = checksum ^ farm.Fingerprint32(lb.Bytes())
+
+		lb.Reset()
+	}
+
+	var signedChecksum int32
+	// This line converts an unsigned integer to a signed integer (32 bits).
+	// It is needed to be able to calculate the same value in nodejs for
+	// ringpop-node and the integration tests in ringpop-common.
```
mennopruijssers: is it? even when converting it to a string on line 104?

swim/member.go
```diff
@@ -73,6 +80,55 @@ func (m Member) checksumString(b *bytes.Buffer) {
...
+	// checksum string for the node.
+	b.WriteString("#labels")
+	b.WriteString(strconv.Itoa(int(checksum)))
+}
+
+// checksum computes a checksum for the labels. It will return 0 (zero) when no
+// labels are set, but 0 does not indicate that no labels are set. It could be
+// possible that 0 is computed as the checksum.
+func (l LabelMap) checksum() int32 {
+	var lb bytes.Buffer
```
mennopruijssers: Might be good to pool this buffer: this method will be called a lot

swim/member.go
```diff
@@ -73,6 +80,55 @@ func (m Member) checksumString(b *bytes.Buffer) {
...
+	var checksum uint32
+	for key, value := range l {
+		// decide if we want to escape both fields to enforce uniqueness of the
+		// key-value checksum.
+		lb.WriteString("/")
+		lb.WriteString(key)
+		lb.WriteString("/")
+		lb.WriteString(value)
+
+		checksum = checksum ^ farm.Fingerprint32(lb.Bytes())
```
mennopruijssers: whats the reasoning behind this xor? why not just fingerprint the whole labels as a string?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+	// from impacting the performance of the gossip protocol.
+
+	// ensure that there is a labels map
+	if m.local.Labels == nil {
```
mennopruijssers: Do we need locking around this?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+		m.local.Labels = make(map[string]string)
+	}
+
+	// set the label
+	m.local.Labels[key] = value
```
mennopruijssers: (And here as well?)

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+
+// LocalLabelsAsMap copies the labels set on the local node into a map for the
+// callee to use. Changes to this map will not be reflected in the labels kept
+// by this node.
+func (m *memberlist) LocalLabelsAsMap() map[string]string {
+	if len(m.local.Labels) == 0 {
+		return nil
+	}
+
+	cpy := make(map[string]string)
```
mennopruijssers: `cpy := make(map[string]string, len(m.local.Labels))`?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+// Remove a label from the local map of labels. This will trigger a reincarnation
+// of the member to gossip its labels around. It returns true if all labels have
+// been removed.
+func (m *memberlist) RemoveLocalLabel(keys ...string) bool {
+	if len(m.local.Labels) == 0 || len(keys) == 0 {
+		// nothing to delete
+		return false
+	}
+
+	removed := true
```
mennopruijssers: I would prefer a less optimistic approach here:
`removed := false
...
removed = removed || has
`

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+	for _, key := range keys {
+		_, has := m.local.Labels[key]
+		delete(m.local.Labels, key)
+		removed = removed && has
+	}
+	m.postLocalUpdate()
```
mennopruijssers: Maybe only `postLocalUpdate` if there are changes?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+		m.local.Labels = make(map[string]string)
+	}
+
+	// copy the key-value pairs to our internal labels. By not setting the map
+	// of labels to the Labels value of the local member we prevent removing labels
+	// that the user did not specify in the new map.
+	for key, value := range labels {
+		m.local.Labels[key] = value
+	}
+	m.postLocalUpdate()
```
mennopruijssers: Maybe only `postLocalUpdate` if there are changes?

swim/member_predicate.go
```diff
@@ -0,0 +1,43 @@
+package swim
+
+// MemberPredicate is a function that tests if a Member satisfies a condition.
+// It is advised to use exported functions on Member instead of its exported
+// fields in case we want to extract the functionality of Member to an Interface
+// in the future. This is likely to happen if we pursue plugable membership.
+type MemberPredicate func(member Member) bool
+
+// TestMember can take multiple predicates and test them against a member
+// returning if the member satisfies all the predicates. This means that if one
```
mennopruijssers: Maybe be explicit in the case of `0` predicates (current code returns true)?

swim/member_predicate.go
```diff
@@ -0,0 +1,43 @@
...
+// test fails it will stop executing and return with false.
+func TestMember(member Member, predicates ...MemberPredicate) bool {
```
mennopruijssers: Not a big fan of the `Test` prefix: in general it's used for tests...

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+	// from impacting the performance of the gossip protocol.
+
+	// ensure that there is a labels map
+	if m.local.Labels == nil {
+		m.local.Labels = make(map[string]string)
+	}
+
+	// set the label
+	m.local.Labels[key] = value
+	m.postLocalUpdate()
```
mennopruijssers: Maybe only when `key` is new or `value` is different to prevent unnecessary reincarnations?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+}
+
+// SetLocalLabels updates multiple labels at once. It will take all the labels
+// that are set in the map passed to this function and overwrite the value with
+// the value in the map. Keys that are not present in the provided map will
+// remain in the labels of this node.
+func (m *memberlist) SetLocalLabels(labels map[string]string) {
+	// ensure that there is a labels map
+	if m.local.Labels == nil {
+		m.local.Labels = make(map[string]string)
```
mennopruijssers: use capacity?

swim/memberlist.go
```diff
@@ -279,6 +281,83 @@ func (m *memberlist) SetLocalStatus(status string) {
...
+	}
+
+	// copy the key-value pairs to our internal labels. By not setting the map
+	// of labels to the Labels value of the local member we prevent removing labels
+	// that the user did not specify in the new map.
+	for key, value := range labels {
+		m.local.Labels[key] = value
```
mennopruijssers: sanity check input?

swim/memberlist.go
mennopruijssers: nvm; I thought it was returning true when at least one label was removed but documentation states "all"...

swim/member.go
```diff
@@ -128,6 +184,14 @@ func shouldProcessGossip(old *Member, gossip *Member) bool {
 		return false
 	}
 
+	// Only pick the gossip if the labels have a lower checksum then the checksum
+	// of the currently known labels
+	gossipLabelsChecksum := gossip.Labels.checksum()
+	oldLabelsChecksum := old.Labels.checksum()
+	if gossipLabelsChecksum < oldLabelsChecksum {
```
mennopruijssers: Nit: why not higher? Higher state precedence return true as well.

swim/memberlist.go
```diff
@@ -231,11 +231,13 @@ func (m *memberlist) RandomPingableMembers(n int, excluding map[string]bool) []*
 }
 
 // returns an immutable slice of members representing the current state of the membership
```
mennopruijssers: maybe update the documentation to mention the predicats?

swim/node.go
```diff
@@ -136,14 +136,16 @@ func mergeDefaultOptions(opts *Options) *Options {
...
+	CountReachableMembers(predicates ...MemberPredicate) int
 	Destroy()
 	GetChecksum() uint32
-	GetReachableMembers() []string
+	GetReachableMembers(predicates ...MemberPredicate) []Member
 	MemberStats() MemberStats
 	ProtocolStats() ProtocolStats
 	Ready() bool
 	RegisterListener(l events.EventListener)
+
```
mennopruijssers: nit: why the extra linebreak?

mennopruijssers:
I left some comments; besides that, it would be great if you could update the `changes.md`. Maybe add an explicit line about the changes in interface?

examples/role-labels/gen-go/role/ringpop-role.go
thanodnl: This is part of generated code, I'd rather not change that at this point. Will make a branch to see if it helps to add a comment to the top of the generated code indicating it is generated so github doesn't show the contents

examples/role-labels/main.go
thanodnl: Not in the current implementation. Which is a downside. I've been thinking about ways to solve this but couldn't come up with one easy.

Would be great if we can set labels before bootstrapping so they can be configured from the outside via the Options pattern.

swim/member.go
thanodnl: By using the XOR of the individual checksums the algorithm is independent of the order in which the labels are iterated through. This is beneficial for two reasons:

1. We save on list creation/manipulation of the keys to sort the keys in deterministic order for a consistent hash independent of label order
2. As @Motiejus pointed out the sorting of strings might be local dependent which can cause problems if systems run on different locales.

swim/member.go
thanodnl: The converting only works on signed integers, if we do not convert it we would run into a possibility where an overflow error (panic) might occur when transforming it into an int.

Besides, nodejs gives us potentially signed numbers when multiple farmhashes are xorred, therefore I have chosen to transform the number to its signed equivalent given the bits.

An other option I have thought about is transforming the hash into its HEX representation, but this was non trivial in both golang and nodejs.

swim/member.go
thanodnl: Because the choice for the 'winning' view of a member then starts relying on the order in which gossips are received. eg. the last update to come in that is different will always win. As long as the member owning the keys is in the network it will be solved in the end because it will keep reasserting as long as 2 competing states are gossipped around, however if that member is not in the network anymore the network might have a hard time to converge as long as both gossips are maintained.

With this check it is order independent which state will be chosen by an actor in the network and therefore converge (I do not have formal proof for this, but in a [prior system](https://github.com/thanodnl/Evade) where I implemented an eventually consistent data structure I used the order independence to converge which worked well).

swim/member_predicate.go
thanodnl: I'm open for a suggestion. I prefixed it with `Test` here because it tests if the member matches the predicates. However I share your sentiment about the `Test` prefix being very specific for tests in golang.

swim/member_predicate.go
thanodnl: `MemberWithLabel` makes sense.

In the role example one could imagine that a node can fulfill multiple roles at once by using the role as the key of the label and omit/remove the label when the node does not fulfill the role in question. However it could be made more explicit by using the value `true` or `1`. But I like to very explicit with this predicate for now so I would opt for the `MemberWithLabelAndValue`.

swim/memberlist.go
thanodnl: I think locking makes sense. The SetLocalState should share the same lock, and potentially the bumIncarnation as well.

swim/memberlist.go
thanodnl: Don't think we want sanity checking here. These functions are internals. If we want a private namespace for system settings it should be applied at a higher level where a user is able to input labels.

As long as the labels are string:string it works on the gossip layer.

swim/member.go
mennopruijssers: Makes sense. Maybe one of us should write a javascript implementation of your tests for this function so we're sure that we can do the same in js-world before landing this.

swim/member_predicate.go
mennopruijssers: `MemberMatchesPredicates`?

swim/memberlist.go
mennopruijssers: In that case you should move the todo from `SetLocalLabel` to a higher level as well ;)

swim/node.go
thanodnl: That is true, but I don't think there is a prepend function to an array so I went for this one.

An other thing that I have been thinking is making a combiner predicate that takes a predicate and a variadic predicates and tests them all. This would also 'prevent' list expansion/memory allocation. But for both I think we need some more measurements and testing.

Since in a stable ring there should be only nodes in alive and suspect state I think the optimization of moving it to the front is needless at this point till we run into a case where it will be actually make a difference, but that needs a good benchmarking suite.

swim/member.go
thanodnl: Fair point on your update. I will change it around so we pick the highest value in all checks.

Besides I will add a second if statement to always return false on a lower number to harden for future additions in the same way that all other checks are in pairs.

swim/member.go
thanodnl: I wrote this because the other was not doable in nodejs. I needed to write the nodejs version because of the integration tests: https://github.com/uber/ringpop-common/blob/feature/labels/test/membership-checksum.js#L35-L42

mennopruijssers: Awesome!

swim/member.go
mennopruijssers: We could off course just has `key=value` and disallow `=` to be part of the key (input validation for key and value is still a todo right?)

swim/handlers.go
```diff
@@ -155,7 +155,10 @@ func (n *Node) adminLeaveHandler(ctx json.Context, req *emptyArg) (*Status, erro
 }
 
 func (n *Node) adminSetLabelHandler(ctx json.Context, req *SetLabelRequest) (*Status, error) {
-	n.Labels().Set(req.Key, req.Value)
+	err := n.Labels().Set(req.Key, req.Value)
+	if err != nil {
+		return nil, err
+	}
 	return &Status{Status: "ok"}, nil
```
mennopruijssers: `Set` returns false when nothing changed right? Maybe nice to put that in the response as well?

swim/member.go
```diff
@@ -184,14 +197,24 @@ func shouldProcessGossip(old *Member, gossip *Member) bool {
...
-	// of the currently known labels
+	// keep the checksum values in local variables. The checksums are not cached
+	// and require some compute to get them, better to do once than twice.
 	gossipLabelsChecksum := gossip.Labels.checksum()
 	oldLabelsChecksum := old.Labels.checksum()
-	if gossipLabelsChecksum < oldLabelsChecksum {
+
+	// Gossips with a higher checksum should be processed to let the cluster
+	// converge to the labels that cause the highest checksum.
+	if gossipLabelsChecksum > oldLabelsChecksum {
```
mennopruijssers: awesome!

Do you think we should add warn-logging here? different label(-checksums) with the same incarnation number is a sign of an implementation error.

swim/member_predicate.go
```diff
@@ -18,18 +18,18 @@ func TestMember(member Member, predicates ...MemberPredicate) bool {
...
+func memberIsReachable(member Member) bool {
 	return member.isReachable()
 }
 
-// LabeledMember returns a predicate able to test if the value of a label on a
-// member is equal to the provided value.
-func LabeledMember(key, value string) MemberPredicate {
+// MemberWithLabelAndValue returns a predicate able to test if the value of a
+// label on a member is equal to the provided value.
+func MemberWithLabelAndValue(key, value string) MemberPredicate {
```
mennopruijssers: I really love the new naming ❤️ !

swim/memberlist.go
```diff
@@ -290,9 +291,16 @@ func (m *memberlist) SetLocalLabel(key, value string) error {
 		m.local.Labels = make(map[string]string)
 	}
 
+	old, had := m.local.Labels[key]
+
 	// set the label
 	m.local.Labels[key] = value
-	m.postLocalUpdate()
+
+	if !had || old != value {
```
mennopruijssers: Do you think we should return a `bool` here to indicate a change (like on removal)?

swim/memberlist.go
```diff
@@ -327,16 +335,27 @@ func (m *memberlist) LocalLabelsAsMap() map[string]string {
 func (m *memberlist) SetLocalLabels(labels map[string]string) {
 	// ensure that there is a labels map
 	if m.local.Labels == nil {
-		m.local.Labels = make(map[string]string)
+		m.local.Labels = make(map[string]string, len(labels))
 	}
 
+	// keep track if we made changes to the labels
+	changes := false
```
mennopruijssers: Same here?

swim/member.go
Motiejus: s/local/locale/

Can you include this comment to the code? Will be super useful for future readers/maintainers.

swim/member.go
```diff
@@ -124,11 +124,19 @@ func (l LabelMap) checksum() int32 {
...
 
+		// The checksum is calculated by xorring the checksums of all individual
+		// labels. This makes the checksum of the labels order independant. This
+		// is easier compared to sorting the labels by their key because of two
+		// reasons.
+		// 1. It saves memory allocations to have the keys in a slice.
+		// 2. This method is guaranteed to be locale independant where sorting
+		//    of strings might be different on different locale settings. This
+		//    would cause indefinite fullsync storms because two ringpops would
+		//    never agree on membership checksums.
```
Motiejus: like

swim/memberlist.go
thanodnl: Has been implemented in #171

mennopruijssers:
Looks good!

swim/member.go
```diff
@@ -122,10 +123,14 @@ func (l LabelMap) checksum() int32 {
 	// make sure the buffer is empty after getting a buffer from our pool
 	lb.Reset()
 	for key, value := range l {
-		// decide if we want to escape both fields to enforce uniqueness of the
-		// key-value checksum.
+		// The bytes used for checksumming are the following:
+		// <4 bytes: length of key>
+		binary.Write(lb, binary.BigEndian, int32(len(key)))
```
mennopruijssers: Note: `len(string)` returns the length in _bytes_. This means that in node we should use [`Buffer.byteLength`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_bytelength_string_encoding) instead of `string.length`

thanodnl: That was the exact behavior I was aiming for. This is implemented in nodejs via: https://github.com/uber/ringpop-common/commit/7e45a443939f651fffff0b7a1ec1a8e7bf8cb816 exactly as you describe by taking the length of the buffer.

mennopruijssers:
LGTM! (+ less races 🎉 !)

swim/member.go
```diff
@@ -73,6 +90,71 @@ func (m Member) checksumString(b *bytes.Buffer) {
...
 	b.WriteString(m.Status)
 	b.WriteString(strconv.FormatInt(m.Incarnation, 10))
+	m.Labels.checksumString(b)
+}
+
+// checksumString adds the label portion of the checksum to the buffer that is
+// passed in. The string will not be appended in the case where labels are not
+// set on this member. This is for backwards compatibility reasons with older
+// versions.
+func (l LabelMap) checksumString(b *bytes.Buffer) {
```
Motiejus: How about moving the label checksumming part to a different file? I think it's pure and quite independent?

swim/memberlist.go
```diff
@@ -275,16 +278,138 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
 }
 
 func (m *memberlist) SetLocalStatus(status string) {
+	m.members.Lock()
 	m.local.Status = status
+	m.members.Unlock()
 	m.postLocalUpdate()
 }
 
+func (m *memberlist) SetLocalLabel(key, value string) error {
```
Motiejus: Why not call `SetLocalLabels` with a map of 1 element?

swim/memberlist.go
```diff
@@ -275,16 +278,138 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
...
+	}
+
+	return nil
+}
+
+// Remove a label from the local map of labels. This will trigger a reincarnation
+// of the member to gossip its labels around. It returns true if all labels have
+// been removed.
+func (m *memberlist) RemoveLocalLabel(keys ...string) bool {
+	m.members.Lock()
```
Motiejus: Maybe we can acquire the lock below?

swim/memberlist.go
```diff
@@ -275,16 +278,138 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
...
+	m.members.Unlock()
+
+	if changes {
+		m.postLocalUpdate()
+	}
+
+	return nil
+}
+
+// Remove a label from the local map of labels. This will trigger a reincarnation
```
Motiejus: s/a label/a list of labels/

Also, mention that it is not an error to call this function with a non-existing key.

swim/memberlist.go
```diff
@@ -275,16 +278,138 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
...
+// of the member to gossip its labels around. It returns true if all labels have
+// been removed.
+func (m *memberlist) RemoveLocalLabel(keys ...string) bool {
```
Motiejus: RemoveLocalLabel**s** ?

swim/memberlist_test.go
```diff
@@ -338,6 +342,114 @@ func (s *MemberlistTestSuite) TestApplyUnknownTombstone() {
...
+	s.Assert().Equal("world", value, "expected a value of world for the local label")
+	s.Assert().Equal(1, s.node.disseminator.ChangesCount(), "expected to have 1 change recorded in the disseminator")
+
+	// testing to overwrite the value
+	s.m.SetLocalLabel("hello", "baz")
+	value, has = s.m.GetLocalLabel("hello")
+
+	s.Assert().True(has, "expected to have the local label hello after is has been overwritten")
+	s.Assert().Equal("baz", value, "expected a value of baz for the local label")
+	s.Assert().Equal(1, s.node.disseminator.ChangesCount(), "expected to have 1 change recorded in the disseminator after the labels are overwritten")
```
Motiejus: I think I see a race here: from the time label is set until the `disseminator.ChangesCount()` is called. What happens if during the time the dissemination period kicks in and the change is disseminated? Or we are disabling pings somewhere, and I can't see it (likely)?

Motiejus:
Reviewed. Please ping me when there is an update. :)

Motiejus:
Oh, could you also add a test which sets, gets and removes an empty label?

swim/memberlist.go
thanodnl: I had thought about it, but was worried about garbage creation which is not really relevant since it is not expected to be called so often that it would matter.

I can refactor this code to use `SetLocalLabels` with a 1-element-map

swim/memberlist_test.go
thanodnl: that is a good question.

I don't know for sure if automated pinging is disabled or not during this test. But I think we are fine here. The gossip is only removed from the disseminator if it has been gossiped more than the piggyback count which has a lower bound of 15 I believe.

swim/member.go
thanodnl: I think it is quite closely related to the member at this moment, however with the thoughts @mennopruijssers had earlier this week (to modularize checksums and potentially keep multiple) it would make sense to put the logic of the actual membership checksum, which includes the label checksumming, to its own file.

With regards to this function, now that the checksum calculation (the actual number) of the labels is extracted and reused below the body of this function might as well be collapsed into the `Member#checksumString` function since it is very much related with how a member is checksummed.

swim/memberlist_test.go
thanodnl: Just to make sure this is not a race I had a look at the code, and in the setup of the test suite the `Start` method is never called, meaning there is no routine consuming gossips from the disseminator.

Motiejus:
LGTM

Issue #160, opened 2016-07-28
`go tool vet` is not running on all examples


thanodnl:
When running `make lint` the vetting tool is not ran on all our examples by default. This let a vetting error slip through the cracks in the `ping-thrift-gen` example.

For some reason `go tool vet` only runs on this example after installing the example:

```bash
$ go install github.com/uber/ringpop-go/examples/ping-thrift-gen/
```

When you run `make lint` after an installation you will see the following error

```bash
$ make lint
ERROR: ./examples/ping-thrift-gen/main.go:95: github.com/uber/ringpop-go/examples/ping-thrift-gen/gen-go/ping.Pong composite literal uses unkeyed fields
make: *** [lint] Error 1
```

This will prevent the precommit hook that is setup to finish which makes it harder to commit. The easiest way to remove this error is by removing the `ping-thrift-gen` from `pkg` via

```bash
$ rm -rf $GOPATH/pkg/darwin_amd64/github.com/uber/ringpop-go/examples/ping-thrift-gen # OSX specific
```

We should find the root cause for the vetter to not be running correctly and fix the vetting issue that it finds when it does run.

prashantv:
The error in question is a valid error, it's complaining about:
https://github.com/uber/ringpop-go/blob/master/examples/ping-thrift-gen/main.go#L95
```go
	return &gen.Pong{"Hello, world!", identity, &pHeader}, nil
```

The vet tool raises an error when you don't specify keys for literals for types that are defined outside of your package. You can fix this by specifying keys (e.g., `&gen.Pong{Message: "Hello, world!"}`).

The reason this is an error is to make it easier for package owners to add fields to structs without breaking all code that depends on this package. If some package uses `B.Pong{"Hello world"}`, now that call will fail when the owner of B adds a new field to `Pong`, since unkeyed literals must specify all fields.

The vet tool uses the package cache to determine the types. E.g., if the `Pong` type was actually a slice, then you wouldn't need keys.

thanodnl:
Thanks for the explanation why it needs the files in `pkg`. Now we need to find a way to get these files available on travis-ci so the `make lint` test starts failing. Because now we ran into this 'by accident' just wondering if there are more (important) files failing on `vet` as well.

prashantv:
To get packages into the `pkg` directory, you need to do `go install`, or `go test -i` or `go build -i`.

[...]

PR #170, opened 2016-08-10
Fix: Examples Tests

thanodnl:
During the development of labels I needed to add an example and documentation for the new feature. To make sure that the example works before merging I would like to let travis tell me whether or not `make test-examples` works. This fixes all existing examples and as a bonus solves a couple of the linting issues `make lint` complains about after the examples have been built locally which has been discussed in #160.

#160 cannot be closed yet because we still didn't solve the problem where the linter is not ran on the examples on travis.

### DIFF
--- a/examples/ping-json/README.md
+++ b/examples/ping-json/README.md
@@ -22,7 +22,7 @@ Lookup the node `my_key` key belongs to using [tcurl][2]:
 Call the `/ping` endpoint (multiple times) and see the request being forwarded. Each request is sent to a random node in the cluster because of the `-P hosts.json` argument--but is always handled by the node owning the key. This can be seen in the `from` field of the response:
 
     $ tcurl pingchannel -P hosts.json /ping '{"key": "my_key"}'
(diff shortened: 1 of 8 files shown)


Motiejus:
LGTM. cc @severb

examples/ping-json/README.md
```diff
@@ -22,7 +22,7 @@ Lookup the node `my_key` key belongs to using [tcurl][2]:
 Call the `/ping` endpoint (multiple times) and see the request being forwarded. Each request is sent to a random node in the cluster because of the `-P hosts.json` argument--but is always handled by the node owning the key. This can be seen in the `from` field of the response:
 
     $ tcurl pingchannel -P hosts.json /ping '{"key": "my_key"}'
```
mennopruijssers: maybe add the `pheader` here as well?

mennopruijssers:
one small, non-blocking question but: LGTM!

examples/ping-json/README.md
thanodnl: The pheader is already in the second test below this one. This example/test is without the header, but it was failing because the repsonse always contains the `pheader` even when it is empty.

PR #171, opened 2016-08-11
[Labels] Implement sane limits on Labels.

thanodnl:
Since labels are gossiped around for all members it can cause gossip bloat when used irresponsibly. This will add enforceable limits on labels to prevent the such bloat from happening. Labels will be ignored when the application tries to set a label or multiple labels that would violate the allowed budget for the labels.

### DIFF
--- a/swim/labels.go
+++ b/swim/labels.go
@@ -18,7 +18,7 @@ var (
 	//    memberlist being sent over the wire 5 times a second.
 	// When all contiditions are met the Labels would add the following load
 	// (non-compressed) to the network:
-	//    (32+128)*5*1000*5*8 = ~32mbit
+	//    (32+128)*5*1000*5*8 = ~32mbit/s
 	DefaultLabelOptions = LabelOptions{
 		KeySize:   32,
 		ValueSize: 128,
--- a/swim/labels.go
+++ b/swim/labels.go
@@ -3,15 +3,141 @@ package swim
 import (
 	"errors"
 	"strings"
+
+	"github.com/uber/ringpop-go/util"
 )
 
 var (
+	// DefaultLabelOptions contain the default values to be used to limit the
+	// amount of data being gossipped over the network. The defaults have been
+	// chosen with the following assumptions.
+	// 1. 1000 node cluster
+	// 2. Every byte available is used (yes, characters are not bytes but these
+	//    are ballpark figures to protect developers)
+	// 3. Worst case would have continues fullsync's, meaning the complete
+	//    memberlist being sent over the wire 5 times a second.
+	// When all contiditions are met the Labels would add the following load
+	// (non-compressed) to the network:
+	//    (32+128)*5*1000*5*8 = ~32mbit
+	DefaultLabelOptions = LabelOptions{
+		KeySize:   32,
+		ValueSize: 128,
+		Count:     5,
+	}
+
+	// ErrLabelSizeExceeded indicates that an operation on labels would exceed
+	// the configured size limits on labels. This is to prevent applications
+	// bloating the gossip protocol with too much data. Ringpop can be
+	// configured with the settings to control the amount of data that can be
+	// put in a nodes labels.
+	ErrLabelSizeExceeded = errors.New("label operation exceeds configured label limits")
+
 	// ErrLabelInternalKey is an error that is returned when an application
 	// tries to set a label in the internal namespace that is used by ringpop
 	ErrLabelInternalKey = errors.New("label can't be altered by application because it is in the internal ringpop namespace")
 
 	labelsInternalNamespacePrefix = "__"
 )
+
+// LabelOptions controlls the limits on labels. Since labels are gossiped on
+// every ping/ping-req/fullsync we need to limit the amount of data an
+// application stores in their labels. When needed the defaults can be
+// overwritten during the construction of ringpop. This should be done with care
+// to not overwhelm the network with data.
+type LabelOptions struct {
+	// KeySize is the length a key may use at most
+	KeySize int
+
+	// ValueSize is the length a value may use at most
+	ValueSize int
+
+	// Count is the number of maximum allowed (public) labels on a node
+	Count int
+}
+
+func mergeLabelOptions(opts LabelOptions, def LabelOptions) LabelOptions {
+	return LabelOptions{
+		KeySize:   util.SelectInt(opts.KeySize, def.KeySize),
+		ValueSize: util.SelectInt(opts.ValueSize, def.ValueSize),
+		Count:     util.SelectInt(opts.Count, def.Count),
+	}
+}
+
+func (lo LabelOptions) validateLabel(current map[string]string, key, value string) error {
+	if isInternalLabel(key) {
+		// we ignore internal labels in the limits
+		return nil
+	}
+
+	if len(key) > lo.KeySize || len(value) > lo.ValueSize {
+		// if the either the key or the value of the label is bigger then
+		// the max allowed size an error is returned
+		return ErrLabelSizeExceeded
+	}
+
+	// keep track of all labels that are new in this operation to calculate if
+	// we are exceeding the maximum allowed number of labels
+	additionalLabelCount := 0
+
+	_, has := current[key]
+	if !has {
+		// only add a count to the countAfter if the key we are looking
+		// at is a new key
+		additionalLabelCount++
+	}
+
+	// get the count of the current labels, internal labels will not be counted
+	currentCount := countNonInternalLabels(current)
+
+	if additionalLabelCount > 0 && (currentCount+additionalLabelCount) > lo.Count {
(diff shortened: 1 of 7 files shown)


swim/labels.go
```diff
@@ -1,5 +1,154 @@
...
+
+	labelsPrivateNamespacePrefix = "__"
+)
+
+// LabelOptions controlls the limits on labels. Since labels are gossiped on
+// every ping/ping-req/fullsync we need to limit the amount of data an
+// application stores in their labels. When needed the defaults can be
+// overwritten during the construction of ringpop. This should be done with care
+// to not overwhelm the network with data.
+type LabelOptions struct {
```
mennopruijssers: Could be unsigned int's to prevent negative values

swim/labels.go
```diff
@@ -1,5 +1,154 @@
...
+	additionalLabelCount := 0
+
+	_, has := current[key]
+	if !has {
+		// only add a count to the countAfter if the key we are looking
+		// at is a new key
+		additionalLabelCount++
+	}
+
+	if len(key) > lo.KeySize || len(value) > lo.ValueSize {
```
mennopruijssers: Thous could be moved to line 66 to prevent a unnecessary loop?

swim/labels.go
```diff
@@ -1,5 +1,154 @@
...
+		return ErrLabelSizeExceeded
+	}
+
+	// all is ok
+	return nil
+}
+
+func (lo LabelOptions) validateLabels(current map[string]string, additional map[string]string) error {
+	currentCount := 0
+	for key := range current {
```
mennopruijssers: If you move this to line 136, you prevent an unnecessary loop in the case of invalid key or value sizes.

swim/labels_test.go
```diff
@@ -0,0 +1,298 @@
...
+	assert.Equal(t, DefaultLabelOptions.Count, merged.Count, "expected the Count to be taken from the default options")
+
+	onlyCount := LabelOptions{Count: 1}
+	merged = mergeLabelOptions(onlyCount, DefaultLabelOptions)
+	assert.Equal(t, 1, merged.Count, "expected the Count to be set to the value set in the optional LabelOptions")
+	assert.Equal(t, DefaultLabelOptions.ValueSize, merged.ValueSize, "expected the ValueSize to be taken from the default options")
+	assert.Equal(t, DefaultLabelOptions.KeySize, merged.KeySize, "expected the KeySize to be taken from the default options")
+}
+
+// This test relies on the default settings set to:
```
mennopruijssers: Might be "safer" to explicitly set them so we don't need to change tests when the defaults are changed.

swim/labels_test.go
```diff
@@ -0,0 +1,298 @@
...
+	assert.Error(t, err, "expected error when the value length exceeds the configured length")
+
+	err = DefaultLabelOptions.validateLabel(fixtureLabelsRoom, fixtureExceedKeySizePrivate, "some value")
+	assert.NoError(t, err, "expected no error when the key length exceeds the configured length for a private key")
+
+	err = DefaultLabelOptions.validateLabel(fixtureLabelsRoom, fixtureExceedKeySizePrivate, fixtureExceedValueSize)
+	assert.NoError(t, err, "expected no error when the value length exceeds the configured length for a private key")
+}
+
+func TestLabelOptions_validateLabels(t *testing.T) {
```
mennopruijssers: non-blocking:  This are a lot of tests; what i'm missing is adding multiple non-private or private labels (you only test one label or a combination of private and non-private) :)

swim/memberlist.go
```diff
@@ -332,7 +334,12 @@ func (m *memberlist) LocalLabelsAsMap() map[string]string {
 // that are set in the map passed to this function and overwrite the value with
 // the value in the map. Keys that are not present in the provided map will
 // remain in the labels of this node.
```
mennopruijssers: Update docs?

swim/memberlist.go
```diff
@@ -283,8 +283,10 @@ func (m *memberlist) SetLocalStatus(status string) {
 }
 
 func (m *memberlist) SetLocalLabel(key, value string) error {
```
mennopruijssers: Add docs?

mennopruijssers:
Overall, looks really good and well tested! What I'm missing though is a test that makes sure the validator is called when settings labels.
Besides that I added some small non-blocking comments.

swim/labels.go
```diff
@@ -145,6 +131,24 @@ func (lo LabelOptions) validateLabels(current map[string]string, additional map[
 	return nil
 }
 
+func countLabels(labels map[string]string, countPrivate bool) int {
```
mennopruijssers: Is there any reason why we support counting private and non-private labels? It looks like we only need a way to count non private labels. maybe "simplify" this method and only support counting non-privates for know? (And maybe rename it to `countNonPrivateLabels` to be extra specific?)

swim/labels.go
```diff
@@ -145,6 +131,24 @@ func (lo LabelOptions) validateLabels(current map[string]string, additional map[
...
+	if countPrivate {
+		// if we also count private labels we can simply use len()
+		return len(labels)
+	}
+
+	count := 0
+	for key := range labels {
+		if isPrivateLabel(key) {
+			// we ignore private labels in the limits
```
mennopruijssers: nit: this line shuoldn't care about limits; it should only care about the `countPrivate` param.

thanodnl: You're right, this was just copied from when it was calculated in both validation functions. I'll remove it here.

swim/labels.go
thanodnl: Currently there is no reason to count private labels and one could use `len(labels)` in the code when there is a need, it was more due to naming but I will rename the function to `countNonPrivateLabels`.

swim/labels.go
thanodnl: The loop has been extracted into its own function that is called after this check.

swim/labels.go
thanodnl: By changing these value's to `uint` it is non-trivial to compare the numbers to the return value of `len()` as you can see with this test in `gore`.

```
$ gore
gore version 0.2.5  :help for help
gore> var ui uint
gore> var m map[string]string
gore> ui > len(m)
# command-line-arguments
/var/folders/9q/ntq49_lj1_39ry71kwl5x8480000gn/T/347859520/gore_session.go:10: invalid operation: ui > len(m) (mismatched types uint and int)
error: exit status 2
exit status 2
gore>
```

swim/memberlist.go
```diff
@@ -328,11 +334,18 @@ func (m *memberlist) LocalLabelsAsMap() map[string]string {
...
-// that are set in the map passed to this function and overwrite the value with
-// the value in the map. Keys that are not present in the provided map will
-// remain in the labels of this node.
-func (m *memberlist) SetLocalLabels(labels map[string]string) {
+// SetLocalLabels updates multiple labels at once. This operation is validated
+// against the configured limits for labels and will return an
+// ErrLabelSizeExceeded in the case this operation would alter the labels of the
+// node in such a way that the configured limits are exceeded.
+// When an error is returned non of the labels will be updated. When there is no
+// error all labels are updated with their new value.
```
mennopruijssers: `s/value/values`

mennopruijssers:
LGTM!

mennopruijssers:
LGTM! 👍

swim/labels_test.go
```diff
@@ -1,12 +1,98 @@
...
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+var (
```
Motiejus: Can you add a test which fully initializes ringpop with much smaller configured limits, and verify that they are honored?

swim/labels.go
```diff
@@ -3,16 +3,142 @@ package swim
...
+	// amount of data being gossipped over the network. The defaults have been
+	// chosen with the following assumptions.
+	// 1. 1000 node cluster
+	// 2. Every byte available is used (yes, characters are not bytes but these
+	//    are ballpark figures to protect developers)
+	// 3. Worst case would have continues fullsync's, meaning the complete
+	//    memberlist being sent over the wire 5 times a second.
+	// When all contiditions are met the Labels would add the following load
+	// (non-compressed) to the network:
+	//    (32+128)*5*1000*5*8 = ~32mbit
```
Motiejus: add :`/s`

Motiejus:
Besides 1 suggested test, LGTM!

swim/labels_test.go
thanodnl: Thanks, this uncovered that I missed the configuration propagation from ringpop to the memberlist. The tests can be found here: https://github.com/uber/ringpop-go/pull/171/files#diff-2800ccf278cea772a67ec71a391a7761R866

Motiejus:
LGTM

PR #174, opened 2016-08-15
[Labels] Prevent checksum collisions for labels

thanodnl:
This PR hardens the checksumming algorithm behind labels by preventing the easy forging of checksum collisions. The labels checksum is calculated by `hash(key + "=" + value)`. By disallowing applications to use the `=` in their keys it is non-trivial to create a hash collision.

Since the hashes are xorred for order independence having the same hash twice effectively negates a certain value in the checksum.

### DIFF
--- a/swim/labels.go
+++ b/swim/labels.go
@@ -1,5 +1,48 @@
 package swim
 
+import (
+	"errors"
+	"strings"
+)
+
+var (
+	// labelsBannedKeyCharacters keeps a list of all banned occurances for label keys
+	labelsBannedKeyCharacters = []string{
+		// the equal sign is used in label checksum calculation. To prevent
+		// unintended or forced collisions on the checksums of labels we do not
+		// allow the equal sign to be used in the key portion of a label
+		"=",
+	}
+
+	// ErrLabelInvalidKey is an error that indicates that a key contains a
+	// sequence of characters that are not allowed.
+	ErrLabelInvalidKey = errors.New("the label key contains an unallowed sequence of characters")
+)
+
+// validateLabel validates if the values of both the key and the value are
+// allowed to be gossiped around. Preventing the use of characters that have
+// special meaning in swim.
+func validateLabel(key, value string) error {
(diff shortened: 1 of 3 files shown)


swim/labels.go
```diff
@@ -1,5 +1,48 @@
...
+	"strings"
+)
+
+var (
+	// labelsBannedKeyCharacters keeps a list of all banned occurances for label keys
+	labelsBannedKeyCharacters = []string{
+		// the equal sign is used in label checksum calculation. To prevent
+		// unintended or forced collisions on the checksums of labels we do not
+		// allow the equal sign to be used in the key portion of a label
+		"=",
```
Motiejus: `=` seems like too useful character to blacklist. Maybe we can replace it with `\0` instead?

Or rethink the checksum calculation so that we do not need to blacklist any of the characters? E.g. use `concat(hash(hash(key),hash(value)))`, and xor these instead of concatenating `key` and `value`?

mennopruijssers: I don't agree. `=` is only blacklisted for keys and it doesn't seem useful for a key.

mennopruijssers:
Looks good! The only thing missing is a test to make sure they're validated when set through `SetLocalLabel` / `SetLocalLabels`.

swim/labels.go
Motiejus: Well, we don't know how customers are going to use our labels. So I would rather be very prescriptive (e.g. allow only `[a-zA-Z0-9_-]`), or be not prescriptive at all - allow everything.

It feels backward to disable only one symbol just because we don't feel customers will be using them. What if they want their keys for some reason to be base64-encoded movie titles?

dansimau: I strongly agree with @Motiejus here.

thanodnl:
I have change checksum calculation to write the size (in bytes) of the key and value prior to their respective bytes in the buffer that is fingerprinted. This prevents hash collisions that would have been caused by the same input. This diff is now not needed anymore.

PR #177, opened 2016-08-23
Feature: Self-Eviction

thanodnl:
This feature adds an eviction sequence to ringpop to evict a member from the membership prior to shutting down. This can be used to prevent the suspicion period to kick in during rolling restarts, upgrades and downscaling of the number of running applications. Ringpop modules can hook into this system by implementing the `SelfEvictHooks` interface in `swim` and register the hook in `Ringpop`.


ringpop.go
```diff
@@ -60,6 +60,11 @@ type Interface interface {
 
 	HandleOrForward(key string, request []byte, response *[]byte, service, endpoint string, format tchannel.Format, opts *forward.Options) (bool, error)
 	Forward(dest string, keys []string, request []byte, service, endpoint string, format tchannel.Format, opts *forward.Options) ([]byte, error)
+
+	// extend the functionality of self eviction to ringpop
+	// Mockery didn't allow me to embed the interface so we make things WET here
```
mennopruijssers: WET?

scripts/testpop/testpop.go
```diff
@@ -114,6 +117,40 @@ func main() {
...
+		s = syscall.SIGINT
+	} else {
+		s = syscall.SIGTERM
+	}
+
+	signal.Notify(sigchan, s)
+
+	// wait on a signal
+	<-sigchan
+	log.Println("received signal, initiating self eviction")
```
mennopruijssers: maybe add the signal name to the `Println` for completeness?

scripts/testpop/testpop.go
```diff
@@ -114,6 +117,40 @@ func main() {
...
+		log.Error("triggered graceful shutdown. Press Ctrl+C again to force exit.")
+		go func() {
+			// exit on second signal
+			<-sigchan
+			log.Error("Force exiting...")
+			os.Exit(1)
+		}()
+	}
+
+	rp.SelfEvict()
```
mennopruijssers: Should you handle the error here?
And on error: should you do a `os.Exit(0)`?

swim/node.go
```diff
@@ -94,6 +96,11 @@ func defaultOptions() *Options {
 		Clock: clock.New(),
 
 		MaxReverseFullSyncJobs: 5,
+
+		SelfEvict: SelfEvictOptions{
+			PingDisable: false,
```
mennopruijssers: Not sure on the Go convention here but I would expect `DisablePing` or `PingDisabled`

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+	PostEvict()
+}
+
+type hookFn func(SelfEvictHook)
+
+// SelfEvictOptions configures how self eviction should behave. Applications can
+// configure if ringpop should proactively ping members of the network on self
+// eviction and what percentage/ratio of the memberlist should be pinged at most
+type SelfEvictOptions struct {
+	PingDisable bool
```
mennopruijssers: Not sure on the Go convention here but I would expect `DisablePing` or `PingDisabled`

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+type SelfEvictHook interface {
+	// Name returns the name under which the eviction should be regitered
+	Name() string
+
+	// PreEvict is the hook that will be called before ringpop evicts itself
+	// from the membership
+	PreEvict()
+
+	// PostEvict is the hook that will be called after ringpop has evicted
+	// itsels from them memership
```
mennopruijssers: s/itsels/itself

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// ErrEvictionInProgress is returned when ringpop is already in the process
+	// of evicting
+	ErrEvictionInProgress = errors.New("ringpop is already executing an eviction")
+)
+
+// SelfEvict defines the functions that interact with the self eviction of nodes
+// from the membership prior to shutting down
+type SelfEvict interface {
+	RegisterSelfEvictHook(hooks SelfEvictHook) error
```
mennopruijssers: Shouldn't we document these two funcs as well?

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+func (s *selfEvict) SelfEvict() error {
+	if s.currentPhase() != nil {
+		return ErrEvictionInProgress
+	}
+
+	s.logger.Info("ringpop is initiating self eviction sequence")
+
+	// this starts the shutdown sequence, when it returns the complete sequence
+	// has been executed
+	s.preEvict()
```
mennopruijssers: In "Go World" I might prefer each phase not calling the next one move all the phase invocations here. What do you think?

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+	s.runHooks(SelfEvictHook.PostEvict)
+
+	// next phase
+	s.done()
+}
+
+func (s *selfEvict) done() {
+	s.transitionTo(done)
+	s.logger.Info("ringpop self eviction done")
+	//TODO emit total timing
```
mennopruijssers: TODO?

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+		options: options,
+		logger:  node.logger,
+
+		hooks: make(map[string]SelfEvictHook),
+	}
+}
+
+// RegisterSelfEvictHook registers a named pre/post eviction hook pair.
+func (s *selfEvict) RegisterSelfEvictHook(hook SelfEvictHook) error {
+	name := hook.Name()
```
mennopruijssers: Should we also prevent registering hooks after self eviction is started?
Otherwise we might need an extra lock on the hooks to prevent a race.

swim/self_evict_test.go
```diff
@@ -0,0 +1,139 @@
...
+	err = s.node.RegisterSelfEvictHook(hookDuplicate)
+	s.Assert().Error(err, "expected an error when a duplicate hook gets attached to ringpop")
+}
+
+func (s *SelfEvictTestSuite) TestSelfEvict_SelfEvict() {
+	hooks := &MockSelfEvictHook{}
+	hooks.On("Name").Return("hooks")
+
+	hooks.On("PreEvict").Run(func(args mock.Arguments) {
+		phase := s.node.selfEvict.currentPhase()
```
mennopruijssers: Do we want to test a unexported function (`currentPhase`) or should it be exported?

util/util.go
```diff
@@ -244,12 +244,25 @@ func SelectDuration(opt, def time.Duration) time.Duration {
 	return opt
 }
 
-// Min returns min(a,b)
-func Min(a, b int) int {
-	if a < b {
-		return a
+// SelectBool takes an option and a default value and returns the default value if
+// the option is equal to zero, and the option otherwise.
```
mennopruijssers: s/zero/false?

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+		return ErrDuplicateHook
+	}
+
+	s.hooks[name] = hook
+
+	return nil
+}
+
+func (s *selfEvict) SelfEvict() error {
+	if s.currentPhase() != nil {
```
mennopruijssers: Should we add locking here?
When two go-routines call `SelfEvict` at exactly the same time, there's a race causing us to run the full sequence twice.

options.go
```diff
@@ -280,6 +283,24 @@ func TombstonePeriod(period time.Duration) Option {
...
+// other members when it evicts itself from the network
+func SelfEvictPingDisable(disable bool) Option {
+	return func(r *Ringpop) error {
+		r.config.SelfEvict.PingDisable = disable
+		return nil
+	}
+}
+
+// SelfEvictPingRatio configures the maximum percentage/ratio of the members to
+// actively ping while self evicting.
```
dansimau: I think we could improve the documentation here a bit by adding some more information about why you would want to tune this.

E.g.:

`// A higher number means xyz where as a lower number means xyz`

In other words, explain the use of this and the trade-off.

options.go
```diff
@@ -280,6 +283,24 @@ func TombstonePeriod(period time.Duration) Option {
 	}
 }
 
+// SelfEvictPingDisable configures if ringpop should disable actively pinging
+// other members when it evicts itself from the network
```
dansimau: Could we improve the documentation here a bit by explaining when you would use this?

options.go
```diff
@@ -280,6 +283,24 @@ func TombstonePeriod(period time.Duration) Option {
...
+		r.config.SelfEvict.PingDisable = disable
+		return nil
+	}
+}
+
+// SelfEvictPingRatio configures the maximum percentage/ratio of the members to
+// actively ping while self evicting.
+func SelfEvictPingRatio(ratio float64) Option {
+	return func(r *Ringpop) error {
+		r.config.SelfEvict.PingRatio = ratio
```
dansimau: Is it a problem that I could set it to values like `-0.5` or `500`?

ringpop.go
dansimau: Wolfenstein Enemy Territory?

ringpop.go
```diff
@@ -340,6 +346,28 @@ func (rp *Ringpop) setState(s state) {
 	}
 }
 
+// RegisterSelfEvictHook registers the eviction hooks that need to be executed
+// before and after self eviction from the membership. It will return an error
+// when ringpop has not been bootstrapped or when the hooks cannot be registered
+// with the membership protocol.
+func (rp *Ringpop) RegisterSelfEvictHook(hooks swim.SelfEvictHook) error {
```
dansimau: "when the hooks cannot be registered with the membership protocol" - when would this happen?

ringpop.go
```diff
@@ -340,6 +346,28 @@ func (rp *Ringpop) setState(s state) {
...
+	if !rp.Ready() {
+		return ErrNotBootstrapped
+	}
+	return rp.node.RegisterSelfEvictHook(hooks)
+}
+
+// SelfEvict can be called before shutting down the application. When calling
```
dansimau: Should we go so far as to say: "SelfEvict **should** be called before shutting down the application"?

swim/node.go
```diff
@@ -144,6 +153,11 @@ type NodeInterface interface {
 	ProtocolStats() ProtocolStats
 	Ready() bool
 	RegisterListener(l events.EventListener)
+
+	// extend the functionality of self eviction to a node
+	// Mockery didn't allow me to embed the interface so we make things WET here
```
dansimau: [Western European Time](http://acronymfinder.com/WET.html)?

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+)
+
+var (
+	// ErrDuplicateHook is returned when a hook with the given name has already
+	// been registered.
+	ErrDuplicateHook = errors.New("expected hook name to be unique")
+
+	// ErrEvictionInProgress is returned when ringpop is already in the process
+	// of evicting
+	ErrEvictionInProgress = errors.New("ringpop is already executing an eviction")
```
dansimau: nit: would it be more accurate to call this `ErrSelfEvictionInProgress` and say: "ringpop is already executing a self-eviction"?

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+// from the membership prior to shutting down
+type SelfEvict interface {
+	RegisterSelfEvictHook(hooks SelfEvictHook) error
+	SelfEvict() error
+}
+
+// SelfEvictHook is an interface describing a module that can be registered to
+// the self eviction hooks
+type SelfEvictHook interface {
+	// Name returns the name under which the eviction should be regitered
```
dansimau: s/regitered/registered

swim/self_evict.go
```diff
@@ -0,0 +1,244 @@
...
+	Name() string
```
dansimau: Why do hooks have a name?

swim/self_evict_test.go
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
dansimau: 2016? 😝

swim/self_evict_test.go
```diff
@@ -0,0 +1,139 @@
...
+// THE SOFTWARE.
+
+package swim
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/mock"
+
```
dansimau: nit: space

test/lib.sh
```diff
@@ -2,7 +2,7 @@
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="menno/self-eviction"
```
dansimau: does this need to be updated before landing?

ringpop.go
thanodnl: Haha, was a joke to DRY (Don't Repeat Yourself) when something is not dry it is wet, but I will remove the joke from the comments.

swim/self_evict.go
thanodnl: I was thinking the same, but for ease of implementation I wrote it in the exact same way as yours. I will consolidate this into 1 function since it is easier to read in golang than it would have been in nodejs.

swim/self_evict.go
thanodnl: I will prevent hook registration after the eviction process has been started.

swim/self_evict_test.go
thanodnl: I think we want coverage of the function anyway. I don't know if we need separate tests for it. I'm definitely in favor of _NOT_ exporting the function.

util/util.go
thanodnl: I think it is good to not call it false here, even though that is what the change is, but I think we should refer to it as the [zero value](https://golang.org/ref/spec#The_zero_value)

ringpop.go
thanodnl: Good question I will enumerate the cases in the documentation.

swim/node.go
thanodnl: Not DRY, joke will be removed.

swim/self_evict.go
thanodnl: @mennopruijssers originally implemented it with a name to prevent a system like `sevnup` or `serial` to register twice. By giving them a name we can prevent such cases. Besides that it can be used to add more specific information to log messages.

That said, I would also be happy to model it more after the `EventEmitter` where hooks are registered and deregistered by identity. This would also prevent the hooks from being registered twice, however we will lose the opportunity to add the name to the logs.

test/lib.sh
thanodnl: Yes, otherwise we will have the same problems as earlier where `dev` is not tested against `ringpop-common`'s `master` branch.

**Note to self:** DO NOT FORGET THIS TIME

options_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/uber/ringpop-go/swim"
+
```
thanodnl: white line

swim/self_evict.go
thanodnl: Fyi; names have been removed now to simplify the interface. We prevent double registration the same way we do for EventListeners (by equality).

ringpop.go
```diff
@@ -347,17 +348,20 @@ func (rp *Ringpop) setState(s state) {
...
 // RegisterSelfEvictHook registers the eviction hooks that need to be executed
-// before and after self eviction from the membership. It will return an error
-// when ringpop has not been bootstrapped or when the hooks cannot be registered
-// with the membership protocol.
+// before and after self eviction from the membership. An error will be returned
+// if ringpop was unable to register the hooks. This could happen in the following
+// cases:
+// - Ringpop has not been bootstrapped yet
+// - SelfEvict has already been called
+// - The hook was already registered
```
mennopruijssers: s/was/is.

swim/self_evict.go
```diff
@@ -108,63 +105,77 @@ type phase struct {
...
+		}
 	}
 
-	s.hooks[name] = hook
+	s.hooks = append(s.hooks, hook)
 
 	return nil
 }
 
 func (s *selfEvict) SelfEvict() error {
```
mennopruijssers: I think I prefer the previous commit:
```
s.preEvict()
s.evict()
s.postEvict()
s.done()

return nil
```

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: clean clean-mocks coveralls testpop lint mocks out setup test test-integration test-unit test-race
+.PHONY: clean clean-mocks coveralls testpop lint mocks out setup test test-integration test-unit test-race test-all
```
mennopruijssers: Why is this part of this PR?

swim/self_evict.go
```diff
@@ -0,0 +1,285 @@
...
+			"numberOfPings": phase.numberOfPings,
+			"targets":       targets,
+		}).Debug("starting proactive gossip on self evict")
+
+		var wg sync.WaitGroup
+		wg.Add(len(targets))
+		for _, target := range targets {
+			go func(target Member) {
+				defer wg.Done()
+				_, err := sendPing(s.node, target.address(), s.node.pingTimeout)
```
mennopruijssers: Note: this could result in sending a (other) change more than `maxPiggyBackCounter`. Don't think that it's a problem (it shouldn't) but wanted to point it out to double check.

swim/self_evict.go
```diff
@@ -0,0 +1,285 @@
...
+		maxNumberOfPings := int(math.Ceil(float64(numberOfPingableMembers) * s.options.PingRatio))
+
+		// final number of members to ping should not exceed any of:
+		numberOfPings := util.Min(
+			s.node.disseminator.maxP, // the piggyback counter
+			numberOfPingableMembers,  // the number of members we can ping
+			maxNumberOfPings,         // a configured percentage of members
+		)
+
+		phase.numberOfPings = numberOfPings
```
mennopruijssers: edge case: we might ping less if there's a membership change between setting `numberOfPingableMembers` and calling `RandomPingableMembers`.

test/run-integration-tests
```diff
@@ -70,6 +70,7 @@ run-test-for-cluster-size() {
         --enable-feature bidirectional-full-syncs \
         --enable-feature partition-healing \
         --enable-feature labels \
+        --enable-feature self-eviction \
```
mennopruijssers: 🎈 🎉 :dancing_men:👯  🎉 🎈

Makefile
thanodnl: I wanted to have an easy command to run all tests between changes to make sure I didn't have regressions (caught a couple of races before pushing) but there wasn't an easy command so I added it. Could be done in its own PR but would make it more complicated to land.

If you don't want it in here I can remove it but I thought it might be useful to have in the future.

swim/self_evict.go
thanodnl: Yes this is true, because we only decrement the piggyback count after a successful ping. This means that a gossip might be sent out slightly more. This is a non-intrusive operation for the network as nodes that already have this knowledge will do nothing with this information.

#WontChange

ringpop.go
```diff
@@ -352,6 +359,31 @@ func (rp *Ringpop) setState(s state) {
...
 
+// RegisterSelfEvictHook registers the eviction hooks that need to be executed
+// before and after self eviction from the membership. An error will be returned
+// if ringpop was unable to register the hooks. This could happen in the following
+// cases:
+// - Ringpop has not been bootstrapped yet
+// - SelfEvict has already been called
+// - The hook is already registered
+func (rp *Ringpop) RegisterSelfEvictHook(hooks swim.SelfEvictHook) error {
+	if !rp.Ready() {
```
dansimau: What is the reason that we prevent self-evict hooks from being registered before bootstrapping?

From an API perspective, I can imagine in an application lifecycle that I would want to have a setup/init where I register all my hooks and then set the world running by calling `Bootstrap()` (much the same way you'd register handlers before calling `Listen()`).

options.go
```diff
@@ -309,6 +312,39 @@ func LabelLimitValueSize(size int) Option {
 	}
 }
 
+// SelfEvictDisablePing configures if ringpop should disable actively pinging
+// other members when it evicts itself from the network. It is not advised to
+// disable the ping as it might cause the suspicion time to kick in as chances
+// are high that the node will be shutdown before its self-eviction is gossiped.
+// Only use this setting if you explicitly require no extra network traffic
+// during shutdown.
+func SelfEvictDisablePing(disabled bool) Option {
```
thanodnl: Not really except for exposing it to users in case there is a need to turn it off in the future. It is more of a failsafe to have it than a functional piece at the moment.

ringpop.go
thanodnl: The reason to 'prevent' registering hooks is technical. Self-Eviction is part of the SWIM Node which is only initialized after bootstrapping.

I agree that it would be better for usability if it can be registered before bootstrapping, I will see if it can be refactored to accommodate this.

Makefile
```diff
@@ -62,6 +62,10 @@ setup: dev_deps
 
 test:	test-unit test-integration
 
+# lint should happen after test-unit and test-examples as it relies on objects
+# being created during these phases
+test-all: test-unit test-race test-examples lint test-integration
```
thanodnl: Makes sense, I will move this to `test` as it makes sense to run all tests there.

options.go
dansimau: OK, it's just that I can't imagine a scenario where this would ever be a requirement: "Only use this setting if you explicitly require no extra network traffic during shutdown".

If there is no imaginable scenario where someone would use this option, then I'd remove it, as it means one less thing we have to support.

Makefile
dansimau: 👍

ringpop.go
dansimau: Ah, yeah, that sucks. I wondered if it was something like that.

I wouldn't spend too long on it; it's only an API nit. We could always improve on it in the future.

options.go
thanodnl: No that I think a bit more about it I need to agree with you.

The same functionality can be achieved by putting a negative value in the ping ratio (zero does not work as it will fall back on the default value). To be completely sure I will add a test for a negative ping ratio.

PR #180, opened 2016-08-31
use ringpop-common master again for integration tests.

thanodnl:
reset the ringpop-common branch for integration tests back to `master` after it has been wrongfully set to an other branch


dansimau:
LGTM

PR #181, opened 2016-09-06
Travis: consolidate matrix

thanodnl:
Travis jobs take longer and longer to run because of our matrix size. And secondly coveralls spams a lot not with a matrix of 15 runs (eg. 15 coveralls updates on the same code).

This fix consolidates the following runs into 1 and report coveralls after it:
- test-unit
- test-examples
- lint

Since the setup time of a test is in the order of 3 minutes it doesn't add much time and actually reduces over runtime of the tests by running these 4 things at once. They are also all quite stable so it is not expected to add extra retries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
 
 env:
   - RUN="make lint"
-  - RUN="make test-unit"
+  - RUN="make test-unit coveralls"
   - RUN="make test-integration"
   - RUN="make test-race"
   - RUN="make test-examples"
@@ -36,9 +36,6 @@ matrix:
 
 script:
   - test/travis
-
-after_success:
-  - test/update-coveralls
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -22,11 +22,9 @@ install:
   - make setup
 
 env:
-  - RUN="make lint"
-  - RUN="make test-unit"
+  - RUN="make test-unit test-examples lint coveralls"
```
dansimau: should we put lint first? will it fail fast then?

thanodnl: There was a small discussion about linting with @prashantv in https://github.com/uber/ringpop-go/issues/160 where he explained why the linter was not ran against the examples always. The TLDR is iirc that the linter runs against the object code not the source.

By running it at the end we are ensured that all relevant objects have been made, thats why I put it at the end.

dansimau: Gotcha

dansimau:
LGTM

thanodnl:
Merging this as it has been reviewed as well in #182

PR #178, opened 2016-08-26
Fix: Race conditions

thanodnl:
This PR is to remove the races from ringpop-go and let travis actually fail when there are race conditions in ringpop.

PR #182, opened 2016-09-07
[Races] Extract EventEmitter for reuse throughout ringpop

thanodnl:
In an effort to reduce/eliminate all race conditions found in ringpop and as a good practice this PR will extract all EventEmitter logic into its own package with complete test coverage and race free code.

It has been separated from #178 to ease the reviewing process and builds upon #181 to speedup runs on travis.


events/events.go
```diff
@@ -20,7 +20,11 @@
 
 package events
 
-import "time"
+import (
+	"sync"
+
```
thanodnl: remove whiteline.

events/events_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+type testEvent struct{}
+
+type EventsTestSuite struct {
+	suite.Suite
+
+	e emittingEventRegistrar
+}
+
+func (s *EventsTestSuite) SetupTest() {
```
thanodnl: Empty functions can be removed.

events/events_test.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (s *EventsTestSuite) TearDownTest() {
```
thanodnl: Empty functions can be removed.

.travis.yml
```diff
@@ -22,11 +22,9 @@ install:
   - make setup
 
 env:
-  - RUN="make lint"
-  - RUN="make test-unit"
+  - RUN="make test-unit test-examples lint coveralls"
```
CorgiMan: What's the reason for this?

events/events.go
```diff
@@ -31,9 +35,100 @@ type EventListener interface {
...
+	baseEventRegistrar
+}
+
+// EmitEvent will send the event to all registered listeners
+func (a *AsyncEventEmitter) EmitEvent(event Event) {
+	a.lock.RLock()
+	for _, listener := range a.listeners {
+		go listener.HandleEvent(event)
+	}
+	a.lock.RUnlock()
```
CorgiMan: This seems like a race. Aren't we freeing the lock immediately because we don't block on the handlers (but run them async)

.travis.yml
thanodnl: This is part of #181 - If you want changes on it please file them there. TL;DR it speeds up a complete travis matrix run a lot.

events/events.go
thanodnl: We can't omit the lock here because that would facilitate a race on `a.listeners` where listeners are added and removed while events are being emitted. [example](https://travis-ci.org/uber/ringpop-go/jobs/157849594#L530-L576)

.travis.yml
CorgiMan: GREAT

events/events.go
```diff
@@ -31,9 +34,100 @@ type EventListener interface {
 	HandleEvent(event Event)
 }
 
+// EventEmitter describes an interface that can be used to emit events
+type EventEmitter interface {
```
CorgiMan: Is it possible to combine EventEmitter and EventRegistrar? I was a bit confused why we have to; since having two seems to suggest that they should interact. I think having one EventEmitter for the three functions would be nice. What do you think?

events/events.go
CorgiMan: Alright seems good.

events/events.go
```diff
@@ -31,9 +34,100 @@ type EventListener interface {
...
+
+// SyncEventEmitter is an implementation of both an EventRegistar and EventEmitter
+// that emits events in the calling go routine.
+type SyncEventEmitter struct {
+	baseEventRegistrar
+}
+
+// EmitEvent will send the event to all registered listeners
+func (a *SyncEventEmitter) EmitEvent(event Event) {
+	a.lock.RLock()
```
CorgiMan: Could you add a comment here why you are looping like this pls?

events/events.go
```diff
@@ -31,9 +34,100 @@ type EventListener interface {
...
 // EventRegistrar is an object that you can register EventListeners on.
 type EventRegistrar interface {
-	RegisterListener(EventListener)
+	AddListener(EventListener)
+	RemoveListener(EventListener)
+}
+
+type baseEventRegistrar struct {
+	lock      sync.RWMutex
+	listeners []EventListener
```
CorgiMan: I think we have to add a comment saying how listeners is used when adding, removing and iterating. Also that it should never be mutated (only reassigned). I feels a bit fragile in the way it's done atm but I can't think of anything better. I think that's just the nature of concurrency..

events/events_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// expect listener to be invoked only once
+	wg1.Add(1)
+	l1 := &MockEventListener{}
+	l1.On("HandleEvent", testEvent{}).Run(func(args mock.Arguments) {
+		wg1.Done()
+	}).Return()
+	s.e.AddListener(l1)
+	// make sure listener does stick around for next tests
+	defer s.e.RemoveListener(l1)
```
CorgiMan: aren't these garbage collected?

events/events.go
```diff
@@ -31,9 +34,100 @@ type EventListener interface {
...
+		return
+	}
+
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	// Check if listener is already registered
+	for _, listener := range a.listeners {
+		if listener == l {
+			return
```
CorgiMan: can we return false here?

events/events.go
```diff
@@ -31,9 +34,100 @@ type EventListener interface {
 	HandleEvent(event Event)
```
CorgiMan: I'm always confused that this isn't called EventHandler. Is there a way we can change it?

events/events.go
```diff
@@ -31,9 +34,100 @@ type EventListener interface {
...
+	for i := range a.listeners {
+		if a.listeners[i] == l {
+			// create a new list excluding the listener that needs removal
+			listenersCopy := make([]EventListener, 0, len(a.listeners)-1)
+			listenersCopy = append(listenersCopy, a.listeners[:i]...)
+			listenersCopy = append(listenersCopy, a.listeners[i+1:]...)
+			a.listeners = listenersCopy
+			break
+		}
+	}
```
CorgiMan: return false if no listener is removed

CorgiMan:
I left a number of comments. I'm a big fan of what your doing here and how you are doing it!

CorgiMan:
I like it. Looks good.

events/events.go
```diff
@@ -31,9 +34,122 @@ type EventListener interface {
...
+	// listeners is a slice keeping all added EvenListener interfaces. The slice
+	// is only assinged, but never altered in place. When the slice is accessed
+	// an approriate lock needs to be aquired on listenersLock. After reading
+	// the slice it is safe to release the lock because the underlying array is
+	// never changed in place.
+	listeners     []EventListener
+	listenersLock sync.RWMutex
+}
+
+// AddListener adds a listener to the Event Registar. Events emitted on this
```
Motiejus: s/Registar/Registrar/

events/events_test.go
```diff
@@ -0,0 +1,170 @@
...
+		wg2.Done()
+	}).Return()
+	s.e.AddListener(l2)
+	// make sure listener does stick around for next tests
+	defer s.e.RemoveListener(l2)
+
+	// emit event by value
+	s.e.EmitEvent(testEvent{})
+
+	wg1.Wait()
```
Motiejus: If for some reason `wg1.Done()` is not executed due to a regression, this line will wait indefinitely.

Do you think it's an OK indication of the test failure if that happens? Or maybe there is another way to implement this check, but with a fixed timeout (e.g. 1s), and a clearer error message like `not enough events were emitted in a long time`?

events/events_test.go
```diff
@@ -0,0 +1,170 @@
...
+	removed := s.e.RemoveListener(nil)
+	s.Assert().False(removed, "expect a nil listener to not be removed since it has not been added")
+
+	s.e.AddListener(nil)
+
+	removed = s.e.RemoveListener(nil)
+	s.Assert().False(removed, "expect a nil listener to not be removed since it can't be added")
+}
+
+func (s *EventsTestSuite) TestDeregisterDuringHandleEvent() {
```
Motiejus: Rename to TestRemoveListenerDuringHandleEvent

events/events_test.go
```diff
@@ -0,0 +1,170 @@
...
+	l1.AssertNumberOfCalls(s.T(), "HandleEvent", 1)
+}
+
+func TestSyncEventEmitter(t *testing.T) {
+	suite.Run(t, &EventsTestSuite{
+		e: &SyncEventEmitter{},
+	})
+}
+
+func TestAsyncEventEmitter(t *testing.T) {
```
Motiejus: Can you also add the following tests:

* Emit an event to an empty group (no registered event listeners).
* Make `handleEvent` throw an exception, and verify the subsequent events call the handlers.

forward/forwarder.go
```diff
@@ -85,14 +85,14 @@ func (f *Forwarder) mergeDefaultOptions(opts *Options) *Options {
 
 // A Forwarder is used to forward requests to their destinations
 type Forwarder struct {
+	events.AsyncEventEmitter
```
Motiejus: Not sure I agree with the API here, since it will add `Forwarder.AddListener`, `Forwarder.RemoveListener` fields. It is not obvious they are about event emission.

ringpop.go
```diff
@@ -304,16 +314,10 @@ func (rp *Ringpop) Uptime() (time.Duration, error) {
...
 
-func (rp *Ringpop) emit(event interface{}) {
-	for _, listener := range rp.listeners {
-		go listener.HandleEvent(event)
-	}
-}
-
-// RegisterListener adds a listener to the ringpop. The listener's HandleEvent method
-// should be thread safe.
+// RegisterListener is DEPRECATED, use AddListener
```
Motiejus: Can you help me understand the call sites for which we have to call this function? Can ringpop applications subscribe to ring events?

events/events_test.go
thanodnl: In my opinion a never ending test suite signals to the user well enough that a regression has happened and when ran in verbose mode you will see in which test a deadlock happens.

The problem with such timeouts is that it adds to flappy tests during race detection (and also without the race detector but they are found more often with the race detector). I have removed all timeout's during our tests to make it pass the race detector test consistently so I'm not keen on adding one here again.

Motiejus: (y)

Motiejus:
LGTM

PR #186, opened 2016-11-04
Fix leaking timer goroutines

malandrew:
This fixes two leaking goroutines from the ringpop timers. There is still one more leaking goroutine that I'm aware of in rcrowley/go-metrics:

https://github.com/rcrowley/go-metrics/blob/master/meter.go#L218-L225

That's a ticker that would also require a stop channel.

### DIFF
--- a/ringpop.go
+++ b/ringpop.go
@@ -106,8 +106,9 @@ type Ringpop struct {
 
 	logger log.Logger
 
-	tickers   chan *clock.Ticker
-	startTime time.Time
+	tickers     chan *clock.Ticker
+	tickersStop chan bool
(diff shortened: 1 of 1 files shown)


ringpop.go
```diff
@@ -246,6 +258,8 @@ func (rp *Ringpop) stopTimers() {
 		for ticker := range rp.tickers {
 			ticker.Stop()
 		}
+		close(rp.tickersStop)
+		rp.tickersStop = nil
```
thanodnl: This line causes a data race with lines 229 and 247. Is it really needed to set it to `nil` when the timers are stopped?

PR #187, opened 2016-11-23
[Travis] Remove go1.5 support

thanodnl:
To improve speed on travis we will remove go1.5. This allows us to go back to the head of golint which dropped go1.5 support a while ago.

To speed up even further we consolidate the matrix and run test-race during our normal tests. It was kept separate in case the race tests were flappy after fixing the race condition, but has been proven stable. No need to run separately anymore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,8 @@ mocks:
 	test/gen-testfiles
 
 dev_deps:
-	go get github.com/uber/tchannel-go/thrift/thrift-gen
-	./scripts/go-get-version.sh github.com/golang/lint/golint/.../@c6242afa
+	go get -u github.com/uber/tchannel-go/thrift/thrift-gen
+	go get -u github.com/golang/lint/golint/...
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -47,8 +47,8 @@ mocks:
 	test/gen-testfiles
 
 dev_deps:
-	go get github.com/uber/tchannel-go/thrift/thrift-gen
-	./scripts/go-get-version.sh github.com/golang/lint/golint/.../@c6242afa
+	go get -u github.com/uber/tchannel-go/thrift/thrift-gen
+	go get -u github.com/golang/lint/golint/...
```
mennopruijssers: Does this require that everyone needs to run `make setup` again?

thanodnl: not really, older versions of `golint` will keep working.

thanodnl:
More and more tooling we rely on dropped go1.5 support, merging.

PR #188, opened 2016-11-28
[Feature] Identity carry over part 1 / RBTree abstraction

thanodnl:
This is the first part of identity carry over where we abstract the RBTree to not care about the datatypes stored as long as the key is sortable via a Compare function it will create a RBTree.

This will help in later stages where the `keytype` will be a struct with more data points instead of a string.

### DIFF
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -45,6 +45,12 @@ type Configuration struct {
 	ReplicaPoints int
 }
 
+type replicapoint int
+
+func (me replicapoint) Compare(other interface{}) int {
+	return int(me) - int(other.(replicapoint))
+}
+
 // HashRing stores strings on a consistent hash ring. HashRing internally uses
 // a Red-Black Tree to achieve O(log N) lookup and insertion time.
 type HashRing struct {
@@ -137,7 +143,7 @@ func (r *HashRing) addReplicasNoLock(server string) {
 	r.serverSet[server] = struct{}{}
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Insert(r.hashfunc(address), server)
+		r.tree.Insert(replicapoint(r.hashfunc(address)), server)
 	}
 }
 
@@ -171,7 +177,7 @@ func (r *HashRing) removeReplicasNoLock(server string) {
 	delete(r.serverSet, server)
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Delete(r.hashfunc(address))
+		r.tree.Delete(replicapoint(r.hashfunc(address)))
 	}
 }
 
@@ -276,9 +282,9 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
 	// collected all the servers we want, we have reached the
 	// end of the red-black tree and we need to loop around and inspect the
 	// tree starting at 0.
-	r.tree.LookupNUniqueAt(n, hash, unique)
+	r.tree.LookupNUniqueAt(n, replicapoint(hash), unique)
 	if len(unique) < n {
-		r.tree.LookupNUniqueAt(n, 0, unique)
+		r.tree.LookupNUniqueAt(n, replicapoint(0), unique)
 	}
 
 	var servers []string
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -276,7 +276,7 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
 	}
 
 	hash := r.hashfunc(key)
-	unique := make(map[string]struct{})
+	unique := make(map[valuetype]struct{})
 
 	// lookup N unique servers from the red-black tree. If we have not
 	// collected all the servers we want, we have reached the
@@ -289,7 +289,7 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
 
 	var servers []string
 	for server := range unique {
-		servers = append(servers, server)
+		servers = append(servers, server.(string))
 	}
 	return servers
 }
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -47,8 +47,8 @@ type Configuration struct {
 
 type replicapoint int
 
-func (me replicapoint) Compare(other interface{}) int {
-	return int(me) - int(other.(replicapoint))
+func (r replicapoint) Compare(other interface{}) int {
+	return int(r) - int(other.(replicapoint))
 }
 
 // HashRing stores strings on a consistent hash ring. HashRing internally uses
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -45,10 +45,10 @@ type Configuration struct {
 	ReplicaPoints int
 }
 
-type replicapoint int
+type replicaPoint int
 
-func (r replicapoint) Compare(other interface{}) int {
-	return int(r) - int(other.(replicapoint))
+func (r replicaPoint) Compare(other interface{}) int {
+	return int(r) - int(other.(replicaPoint))
 }
 
 // HashRing stores strings on a consistent hash ring. HashRing internally uses
@@ -143,7 +143,7 @@ func (r *HashRing) addReplicasNoLock(server string) {
 	r.serverSet[server] = struct{}{}
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Insert(replicapoint(r.hashfunc(address)), server)
+		r.tree.Insert(replicaPoint(r.hashfunc(address)), server)
 	}
 }
 
@@ -177,7 +177,7 @@ func (r *HashRing) removeReplicasNoLock(server string) {
 	delete(r.serverSet, server)
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Delete(replicapoint(r.hashfunc(address)))
+		r.tree.Delete(replicaPoint(r.hashfunc(address)))
 	}
 }
 
@@ -282,9 +282,9 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
 	// collected all the servers we want, we have reached the
 	// end of the red-black tree and we need to loop around and inspect the
 	// tree starting at 0.
-	r.tree.LookupNUniqueAt(n, replicapoint(hash), unique)
+	r.tree.LookupNUniqueAt(n, replicaPoint(hash), unique)
 	if len(unique) < n {
-		r.tree.LookupNUniqueAt(n, replicapoint(0), unique)
+		r.tree.LookupNUniqueAt(n, replicaPoint(0), unique)
 	}
 
 	var servers []string
--- a/hashring/rbtree.go
+++ b/hashring/rbtree.go
@@ -261,6 +261,40 @@ func (n *redBlackNode) search(key keytype) (valuetype, bool) {
 	return "", false
 }
 
+// walk walks the tree in sorted order calling the walking verion on everynode.
+// If the walker function should return wether or not the walk should continue,
+// eg. the first time walker returns false the walk function will exit and no
+// more calls to walker will be made. Since the walk is recursive the walk will
+// respond wether or not the walking has been exited prematurely, eg false is a
+// premature exit, true means natural exit
+func (n *redBlackNode) walk(walker func(*redBlackNode) bool) bool {
+	if n == nil {
+		// the end of the tree does not signal the end of walking, but we can't
+		// walk this node (nil) nor left or right anymore
+		return true
+	}
+
+	// walk left first
+	if !n.left.walk(walker) {
+		// stop if walker indicated to not continue
+		return false
+	}
+	// now visit this node
+	if !walker(n) {
+		// stop if walker indicated to not continue
+		return false
+	}
+	// lastly visit right
+	if !n.right.walk(walker) {
+		// stop if walker indicated to not continue
+		return false
+	}
+
+	// signal that we reached the end and walking should continue till we hit
+	// end of tree
+	return true
+}
+
 // Search searches for a value in the redBlackTree, returns the string and true
 // if found or the empty string and false if val is not in the tree.
 func (t *redBlackTree) Search(key keytype) (valuetype, bool) {
--- a/hashring/rbtree.go
+++ b/hashring/rbtree.go
@@ -30,10 +30,12 @@ type keytype interface {
 	Compare(other interface{}) int
 }
 
+type valuetype interface{}
+
 // redBlackNode is a node of the redBlackTree
 type redBlackNode struct {
 	key   keytype
-	str   string
+	value valuetype
 	left  *redBlackNode
 	right *redBlackNode
 	red   bool
@@ -84,9 +86,12 @@ func doubleRotate(root *redBlackNode, dir bool) *redBlackNode {
 
 // Insert inserts a value and string into the tree
 // Returns true on succesful insertion, false if duplicate exists
-func (t *redBlackTree) Insert(key keytype, str string) (ret bool) {
+func (t *redBlackTree) Insert(key keytype, value valuetype) (ret bool) {
 	if t.root == nil {
-		t.root = &redBlackNode{key: key, str: str}
+		t.root = &redBlackNode{
+			key:   key,
+			value: value,
+		}
 		ret = true
 	} else {
 		var head = &redBlackNode{}
@@ -104,7 +109,11 @@ func (t *redBlackTree) Insert(key keytype, str string) (ret bool) {
 		for {
 			if node == nil {
 				// insert new node at bottom
-				node = &redBlackNode{key: key, str: str, red: true}
+				node = &redBlackNode{
+					key:   key,
+					value: value,
+					red:   true,
+				}
 				parent.setChild(dir, node)
 				ret = true
 			} else if isRed(node.left) && isRed(node.right) {
@@ -225,7 +234,7 @@ func (t *redBlackTree) Delete(key keytype) bool {
 	// get rid of node if we've found one
 	if found != nil {
 		found.key = node.key
-		found.str = node.str
+		found.value = node.value
 		parent.setChild(parent.right == node, node.Child(node.left == nil))
 		t.size--
 	}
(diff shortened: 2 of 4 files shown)


hashring/rbtree.go
```diff
@@ -154,7 +167,7 @@ func (t *redBlackTree) Insert(val int, str string) (ret bool) {
 
 // Delete removes a value from the redBlackTree
 // Returns true on succesful deletion, false if val is not in tree
```
mennopruijssers: s/val/key/
s/tree/the tree/

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
+			return n.left.search(key)
 		}
 	} else if n.right != nil {
-		return n.right.search(val)
+		return n.right.search(key)
 	}
 	return "", false
 }
 
+// walk walks the tree in sorted order calling the walking verion on everynode.
```
mennopruijssers: s/version/function
s/everynode/every node

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
+// If the walker function should return wether or not the walk should continue,
```
mennopruijssers: I think this comment could use some love...
Maybe something like:
```
walk Walks the tree in sorted order invoking `walker` for each node. 
If the walker function returns `false`, walking is stopped and no more nodes will be visited.
Returns `true` if all nodes are visited; `false` if not.
```

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
+func (n *redBlackNode) walk(walker func(*redBlackNode) bool) bool {
+	if n == nil {
+		// the end of the tree does not signal the end of walking, but we can't
+		// walk this node (nil) nor left or right anymore
+		return true
+	}
+
+	// walk left first
+	if !n.left.walk(walker) {
+		// stop if walker indicated to not continue
```
mennopruijssers: s/not continue/break ?

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
+		return false
+	}
+
+	// signal that we reached the end and walking should continue till we hit
+	// end of tree
+	return true
+}
+
 // Search searches for a value in the redBlackTree, returns the string and true
```
mennopruijssers: s/a value/a key
s/the string/the value

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
 // if found or the empty string and false if val is not in the tree.
```
mennopruijssers: s/the empty string/an empty string
s/if val/if key

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
-func (t *redBlackTree) Search(val int) (string, bool) {
+func (t *redBlackTree) Search(key keytype) (valuetype, bool) {
 	if t.root == nil {
 		return "", false
 	}
-	return t.root.search(val)
+	return t.root.search(key)
 }
 
 // LookupNAt iterates through the tree from the node with value val, and
```
mennopruijssers: s/LookupNAt/LookupNUniqueAt
s/value val/key
s/unique strings/unique values
s/n strings/n values

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
 // returns the next n unique strings. This function is not guaranteed to
 // return n strings.
-func (t *redBlackTree) LookupNUniqueAt(n int, val int, result map[string]struct{}) {
-	findNUniqueAbove(t.root, n, val, result)
+func (t *redBlackTree) LookupNUniqueAt(n int, key keytype, result map[valuetype]struct{}) {
+	findNUniqueAbove(t.root, n, key, result)
 }
 
 // findNUniqueAbove is a recursive search that finds n unique strings
 // with a value bigger or equal than val
```
mennopruijssers: Basically same as above

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
 
 	// skip left branch when all its values are smaller than val
-	if node.val >= val {
-		findNUniqueAbove(node.left, n, val, result)
+	cmp := node.key.Compare(key)
+	if cmp >= 0 {
+		findNUniqueAbove(node.left, n, key, result)
 	}
 
 	// Make sure to stop when we have n unique strings
```
mennopruijssers: s/strings/values

hashring/rbtree.go
```diff
@@ -234,55 +247,91 @@ func (t *redBlackTree) Delete(val int) bool {
...
 
 // findNUniqueAbove is a recursive search that finds n unique strings
 // with a value bigger or equal than val
-func findNUniqueAbove(node *redBlackNode, n int, val int, result map[string]struct{}) {
+func findNUniqueAbove(node *redBlackNode, n int, key keytype, result map[valuetype]struct{}) {
 	if len(result) >= n || node == nil {
 		return
 	}
 
 	// skip left branch when all its values are smaller than val
```
mennopruijssers: s/values/keys
s/val/key

hashring/rbtree_test.go
```diff
@@ -30,6 +30,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type treeTestInt int
+
+func (me treeTestInt) Compare(other interface{}) int {
```
mennopruijssers: lint doesn't like `me`

hashring/rbtree_test.go
```diff
@@ -88,15 +94,15 @@ func validateRedBlackTree(n *redBlackNode) (int, error) {
 		return 0, err
 	}
 
-	if n.left != nil && n.left.val >= n.val ||
-		n.right != nil && n.right.val <= n.val {
-		err = errors.New(fmt.Sprint("binary tree violation at node val ", n.val))
+	if n.left != nil && n.left.key.Compare(n.key) >= 0 ||
+		n.right != nil && n.right.key.Compare(n.key) <= 0 {
+		err = errors.New(fmt.Sprint("binary tree violation at node val ", n.key))
```
mennopruijssers: s/val/key

hashring/rbtree_test.go
```diff
@@ -88,15 +94,15 @@ func validateRedBlackTree(n *redBlackNode) (int, error) {
...
 		return 0, err
 	}
 
 	if leftHeight != 0 && rightHeight != 0 {
 		if leftHeight != rightHeight {
-			err = errors.New(fmt.Sprint("black height violation at node val ", n.val))
+			err = errors.New(fmt.Sprint("black height violation at node val ", n.key))
```
mennopruijssers: s/val/key

hashring/rbtree_test.go
```diff
@@ -119,73 +125,73 @@ func TestInsert(t *testing.T) {
 
 	// 4, B
 	node := tree.root
-	assert.Equal(t, 4, node.val, "expected tree root val to be 4")
-	assert.Equal(t, "four", node.str, "expected tree root str to be 'four'")
+	assert.Equal(t, treeTestInt(4), node.key, "expected tree root val to be 4")
+	assert.Equal(t, "four", node.value, "expected tree root str to be 'four'")
```
mennopruijssers: s/str/value

hashring/rbtree_test.go
```diff
@@ -119,73 +125,73 @@ func TestInsert(t *testing.T) {
 
 	// 4, B
 	node := tree.root
-	assert.Equal(t, 4, node.val, "expected tree root val to be 4")
-	assert.Equal(t, "four", node.str, "expected tree root str to be 'four'")
+	assert.Equal(t, treeTestInt(4), node.key, "expected tree root val to be 4")
```
mennopruijssers: s/val/key

hashring/rbtree_test.go
```diff
@@ -119,73 +125,73 @@ func TestInsert(t *testing.T) {
...
+	assert.Equal(t, "four", node.value, "expected tree root str to be 'four'")
 	assert.Equal(t, false, node.red, "expected tree root to be black")
 	assert.NotNil(t, node.left, "expected tree root to have left child")
 	assert.NotNil(t, node.right, "expected tree root to have right child")
 
 	// 2,R
 	node = tree.root.left
-	assert.Equal(t, 2, node.val, "got unexpected node val")
-	assert.Equal(t, "two", node.str, "got unexpected node str")
+	assert.Equal(t, treeTestInt(2), node.key, "got unexpected node val")
```
mennopruijssers: s/val/key

hashring/rbtree_test.go
```diff
@@ -119,73 +125,73 @@ func TestInsert(t *testing.T) {
...
+	assert.Equal(t, "two", node.value, "got unexpected node str")
```
mennopruijssers: s/str/value

hashring/rbtree_test.go
```diff
@@ -119,73 +125,73 @@ func TestInsert(t *testing.T) {
...
 	assert.Equal(t, true, node.red, "got unexpected node colour")
 	assert.NotNil(t, node.left, "expected node to have left child")
 	assert.NotNil(t, node.right, "expected node to have right child")
 
 	// 1,B
 	node = tree.root.left.left
-	assert.Equal(t, 1, node.val, "got unexpected node val")
-	assert.Equal(t, "one", node.str, "got unexpected node str")
+	assert.Equal(t, treeTestInt(1), node.key, "got unexpected node val")
```
mennopruijssers: Ok, I stop pointing these out.

hashring/hashring.go
```diff
@@ -137,7 +143,7 @@ func (r *HashRing) addReplicasNoLock(server string) {
 	r.serverSet[server] = struct{}{}
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
```
thanodnl: I agree but due to changes in later PR's I'm not touching it here.

hashring/hashring.go
```diff
@@ -137,7 +143,7 @@ func (r *HashRing) addReplicasNoLock(server string) {
...
-		r.tree.Insert(r.hashfunc(address), server)
+		r.tree.Insert(replicapoint(r.hashfunc(address)), server)
```
thanodnl: This function is added in the hashring abstraction PR that is up already.

hashring/hashring.go
```diff
@@ -171,7 +177,7 @@ func (r *HashRing) removeReplicasNoLock(server string) {
 	delete(r.serverSet, server)
 	for i := 0; i < r.replicaPoints; i++ {
 		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Delete(r.hashfunc(address))
+		r.tree.Delete(replicapoint(r.hashfunc(address)))
```
thanodnl: That can be done after the hashring abstraction, it is out of scope for this PR.

hashring/rbtree_test.go
thanodnl: newer versions do :(, but I have changed it :)

hashring/hashring.go
mennopruijssers: 👍

hashring/hashring.go
mennopruijssers: 💯

hashring/hashring.go
mennopruijssers: 👍

hashring/rbtree.go
```diff
@@ -295,36 +292,36 @@ func (n *redBlackNode) walk(walker func(*redBlackNode) bool) bool {
...
-		return "", false
+		return nil, false
 	}
 	return t.root.search(key)
 }
 
-// LookupNAt iterates through the tree from the node with value val, and
-// returns the next n unique strings. This function is not guaranteed to
-// return n strings.
+// LookupNUniqueAt iterates through the tree from the node last node that is
```
mennopruijssers: s/the node last node/the last node ?

hashring/rbtree.go
```diff
@@ -295,36 +292,36 @@ func (n *redBlackNode) walk(walker func(*redBlackNode) bool) bool {
...
+// smaller than key, and returns the next n unique values. This function is not
+// guaranteed to return n values, less might be returned
 func (t *redBlackTree) LookupNUniqueAt(n int, key keytype, result map[valuetype]struct{}) {
 	findNUniqueAbove(t.root, n, key, result)
 }
 
-// findNUniqueAbove is a recursive search that finds n unique strings
-// with a value bigger or equal than val
+// findNUniqueAbove is a recursive search that finds n unique values
+// with a value bigger or equal than key
```
mennopruijssers: s/with a value/with a key

hashring/hashring.go
```diff
@@ -270,20 +276,20 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
...
 	hash := r.hashfunc(key)
-	unique := make(map[string]struct{})
+	unique := make(map[valuetype]struct{})
 
 	// lookup N unique servers from the red-black tree. If we have not
 	// collected all the servers we want, we have reached the
 	// end of the red-black tree and we need to loop around and inspect the
 	// tree starting at 0.
-	r.tree.LookupNUniqueAt(n, hash, unique)
+	r.tree.LookupNUniqueAt(n, replicapoint(hash), unique)
```
thanodnl: I am not sure if `hashPoint` will be a better name than `replicaPoint`, when talking about the hashring we always talk about replica points. In one of the later changes the `replicaPoint` gets extended with more data eg. identity and replica number of that identity.

This will look like:
```go
r.tree.LookupNUniqueAt(n, replicaPoint{point:hash}, unique)
```

which is more descriptive then the line above.

hashring/rbtree.go
```diff
@@ -234,55 +247,88 @@ func (t *redBlackTree) Delete(val int) bool {
...
+	} else if 0 < cmp {
 		if n.left != nil {
-			return n.left.search(val)
+			return n.left.search(key)
 		}
 	} else if n.right != nil {
-		return n.right.search(val)
+		return n.right.search(key)
+	}
+	return nil, false
```
mennopruijssers: Did you change the return here in purpose? (Used to be `return "", false`). I think `nil` makes more sense but wasn't sure it was intended or an unwanted side-effect

hashring/rbtree.go
```diff
@@ -234,55 +247,88 @@ func (t *redBlackTree) Delete(val int) bool {
...
+// bigger or equal than key
+func findNUniqueAbove(node *redBlackNode, n int, key keytype, result map[valuetype]struct{}) {
 	if len(result) >= n || node == nil {
 		return
 	}
 
-	// skip left branch when all its values are smaller than val
-	if node.val >= val {
-		findNUniqueAbove(node.left, n, val, result)
+	// skip left branch when all its keys are smaller than key
```
mennopruijssers: This comment belongs to the `if` statement. Lets move it one line down :)

hashring/rbtree.go
```diff
@@ -234,55 +247,88 @@ func (t *redBlackTree) Delete(val int) bool {
...
-		return n.right.search(val)
+		return n.right.search(key)
+	}
+	return nil, false
+}
+
+// walk Walks the tree in sorted order invoking `walker` for each node. If the
+// walker function returns `false`, walking is stopped and no more nodes will be
+// visited. Returns `true` if all nodes are visited; `false` if not.
+func (n *redBlackNode) walk(walker func(*redBlackNode) bool) bool {
```
CorgiMan: I think walk is not the best name for this function. "all" or "every" would be better imo.

CorgiMan: And walker would be "condition"

hashring/rbtree.go
```diff
@@ -234,55 +247,88 @@ func (t *redBlackTree) Delete(val int) bool {
...
+	if n == nil {
+		// the end of the tree does not signal the end of walking, but we can't
+		// walk this node (nil) nor left or right anymore
+		return true
 	}
-	return "", false
+
+	// walk left first
+	if !n.left.walk(walker) {
```
CorgiMan: All these conditions could be just one line, which I think reads perfectly fine too.
```
return n.left.walk(walker) && walker(n) && n.right.walk(walker)
```

mennopruijssers: True but I'm not sure if that improves readability/understandability.

hashring/rbtree.go
mennopruijssers: `All`, and `Every` doesn't cover the behaviour since it breaks out of the loop if `walker` return `false`.
`Traverse` or `TraverseUntil` with a `condition`-argument might make more sense...
@thanodnl thoughts?

CorgiMan: I think it does cover the behaviour. "returns true when the conditions is true for every node."

This is also why I think the one liner reads better

```
// returns true when the condition holds for all nodes on the left, this node, and all nodes on the right
return n.left.All(condition) && condition(n) && n.right.All(condition)
```

hashring/rbtree.go
CorgiMan: replacing 20 lines by two. I know which I would prefer. :)

hashring/rbtree.go
mennopruijssers: The Visitor patterns comes close but [Tree traversal](https://en.wikipedia.org/wiki/Tree_traversal) might cover it better. I'll rename it to `TraverseUntil` with a `condition`-argument.

hashring/rbtree.go
```diff
@@ -261,28 +261,28 @@ func (n *redBlackNode) search(key keytype) (valuetype, bool) {
 	return nil, false
 }
 
-// walk Walks the tree in sorted order invoking `walker` for each node. If the
-// walker function returns `false`, walking is stopped and no more nodes will be
+// traverseUntil traverses the nodes in the tree in-order invoking `traverse` for each node. If the
+// traverse function returns `false`, traversal is stopped and no more nodes will be
 // visited. Returns `true` if all nodes are visited; `false` if not.
-func (n *redBlackNode) walk(walker func(*redBlackNode) bool) bool {
+func (n *redBlackNode) traverseUntil(traverse func(*redBlackNode) bool) bool {
```
CorgiMan: I still think this function is overloaded and has too many use cases but I'm cool with leaving it as is. However, this should be really called traverseWhile. "Until" indicates that the traversal should stop when the condition becomes true. "Do this until I give a signal". That means you stop when "I give a signal" is true. 

"Do this *while* I give you the signal" indicates that you should stop when "I give you the signal" is false. Which is what this function does. Does that make sense?

I also think the function should be called condition to make that more clear.

mennopruijssers: I get your point; I'll change it to `traverseWhile` with a `condition`-arg.

PR #189, opened 2016-11-28
[Feature] Identity carry over part 2 / Membership package

thanodnl:
This PR is part 2 for identity carry over.

In this PR we introduce a stricter interface for what a `Membership` is in a `membership` package. The membership defines an 'observable membership' for the application. Only containing members that are an active part of the membership eg. No `faulty`, `leave` or `tombstone` members will be observable.

It introduces what a `Member` is via an interface and makes sure that the `swim.Member` implements it.

Lastly it adds an event that is triggered when changes happen to the membership. This event is different from the swim event that contains changes in that it contains the `Before` and `After` state of a `Member` that can be used by the application to respond to.


membership/events.go
```diff
@@ -0,0 +1,21 @@
...
+	// member is a new member the before state is nil
+	Before Member
+	// After is the state of the member after the change, if the
+	// member left the after state will be nil
+	After Member
+}
+
+// ChangeEvent indicates that the membership has changed. The event will contain
+// a list of changes that will show both the old and the new state of a member.
+// It is not guaranteed that any of the observable state of a member has in fact
```
mennopruijssers: Do we want to filter out un observable changes?

membership/interface.go
```diff
@@ -0,0 +1,14 @@
+package membership
+
+// Member defines a member of the membership that can be used by the rest of the
```
mennopruijssers: What does "used by the rest of the system" mean? Isn't is just a definition of a member in the membership?

membership/interface.go
```diff
@@ -0,0 +1,14 @@
...
+// system
+type Member interface {
+	// GetAddress returns the external address used by the rpc layer to
+	// communicate to the member. It is prefixed with Get for legacy reasons and
+	// can be removed after a refactor of the swim.Member
```
mennopruijssers: Can't we do that before we extract and interface? Otherwise we won't be able to get rid of it without (another) breaking change.

membership/interface.go
```diff
@@ -0,0 +1,14 @@
...
+	GetAddress() string
+
+	// Label reads the label for a givin key from the member. It also returns
```
mennopruijssers: s/givin/given

swim/memberlist.go
```diff
@@ -291,10 +294,12 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
 }
 
 func (m *memberlist) SetLocalStatus(status string) {
+	var before Member
 	m.members.Lock()
+	before = *m.local // make a copy of the before state
 	m.local.Status = status
 	m.members.Unlock()
-	m.postLocalUpdate()
+	m.postLocalUpdate(before)
```
mennopruijssers: As discussed yesterday, It might be easier to refactor the postLocalUpdate to allow a modifier `func` to be passed in. Move the locking there, and let postLocalUpdate keep track of the before and after state.

swim/memberlist.go
```diff
@@ -422,6 +437,27 @@ func (m *memberlist) postLocalUpdate() {
 
 	// kick in our updating mechanism
 	m.node.handleChanges(changes)
+
+	log.Println("calculating membership chage post local update")
```
mennopruijssers: Debugging artifact?

swim/memberlist.go
```diff
@@ -422,6 +437,27 @@ func (m *memberlist) postLocalUpdate() {
...
+	// prepare a membership change event for observable state changes
+	var memberChange membership.MemberChange
+	if before.isReachable() {
+		log.Println("before is reachable")
```
mennopruijssers: artifact?

swim/memberlist.go
```diff
@@ -422,6 +437,27 @@ func (m *memberlist) postLocalUpdate() {
...
+		memberChange.Before = before
+	}
+	if m.local.isReachable() {
+		log.Println("after is reachable")
```
mennopruijssers: artifact?

swim/memberlist.go
```diff
@@ -422,6 +437,27 @@ func (m *memberlist) postLocalUpdate() {
...
+		memberChange.After = *m.local
+	}
+
+	if memberChange.Before != nil || memberChange.After != nil {
+		log.Println("emitting change")
```
mennopruijssers: artifact?

swim/memberlist.go
```diff
@@ -517,13 +555,28 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
 			} else {
 				// otherwise it can be applied to the memberlist
 
+				// prepare the change and collect if there is an outside
```
mennopruijssers: I can't comment on line 554 but if we want to create memberChange-events for non-observable changes (see question above), we might want to append a `memberChange` about the local member on line 554 as well

swim/memberlist.go
```diff
@@ -517,13 +555,28 @@ func (m *memberlist) Update(changes []Change) (applied []Change) {
...
+				}
+
 				if !has {
 					// if the member was not already present in the list we will
 					// add it and assign it a random position in the list to ensure
 					// guarantees for pinging
 					m.members.byAddress[gossip.Address] = &gossip
 					i := m.getJoinPosition()
 					m.members.list = append(m.members.list[:i], append([]*Member{&gossip}, m.members.list[i:]...)...)
+
```
mennopruijssers: new line?

membership/interface.go
thanodnl: We should but in a separate PR. It is littered throughout the code:

```bash
$ git grep -E "\.Address" | wc -l
165
```

membership/events.go
mennopruijssers: Agreed, makes senses

swim/memberlist.go
mennopruijssers: 👍

swim/memberlist.go
```diff
@@ -294,12 +293,10 @@ func (m *memberlist) MakeFaulty(address string, incarnation int64) []Change {
 }
 
 func (m *memberlist) SetLocalStatus(status string) {
-	var before Member
-	m.members.Lock()
-	before = *m.local // make a copy of the before state
-	m.local.Status = status
-	m.members.Unlock()
-	m.postLocalUpdate(before)
+	m.updateLocalMember(func(member *Member) bool {
```
mennopruijssers: Sweet!

swim/memberlist.go
```diff
@@ -389,69 +382,75 @@ func (m *memberlist) SetLocalLabels(labels map[string]string) error {
...
+// made changes, only if changes are made the incarnation number will be bumped
+// and the new state will be gossiped to the peers
+func (m *memberlist) updateLocalMember(update func(*Member) bool) {
+	var changes []Change
+
 	m.members.Lock()
-	change := m.bumpIncarnation()
+
+	before := *m.local
+	didUpdate := update(m.local)
```
mennopruijssers: Is it easy to compare `before` with `m.local` instead of relying on the return value of `update`?

swim/memberlist.go
```diff
@@ -389,69 +382,75 @@ func (m *memberlist) SetLocalLabels(labels map[string]string) error {
...
+	if didUpdate {
```
mennopruijssers: Might be more readable to do 
```
if !didUpdate {
    m.members.Unlock();
    return
}
```
and remove the other `!didUpdate`?
The only downside is that there are two `unlock`'s...

CorgiMan: I agree with menno; moving the return up is more idiomatic. We could defer the unlock since it's safer in this case.

swim/memberlist.go
```diff
@@ -381,47 +382,81 @@ func (m *memberlist) SetLocalLabels(labels map[string]string) error {
...
-// make sure that its new state has a higher incarnation number and the change
-// will be recorded as a change to gossip around.
-func (m *memberlist) postLocalUpdate() {
-	// bump our incarnation for this change to be accepted by all peers
+// updateLocalMember takes an update function to upate the member passed in. The
+// update function can make mutations to the member and should indicate if it
+// made changes, only if changes are made the incarnation number will be bumped
+// and the new state will be gossiped to the peers
+func (m *memberlist) updateLocalMember(update func(*Member) bool) {
+	var changes []Change
```
CorgiMan: Why is this moved up? I don't find this more readable. Having it in this form is fine.

I thought we had the `changes := []Change{}` form because it is marshaled into "[]" by the json pkg. While `var changes []Cange` will be Marshaled as "null"

swim/memberlist.go
thanodnl: It would likely work to test the value of both structs which can be done with a single `==` statement. We need to test this though.

Two caveats are if the pointer of the `Labels` are replaced with a pointer to a map containing the same data. The other is when the map is changed in place (this should never be done as it will cause more issues). Both would result in the wrong detection of changes.

I am in favor of not returning the boolean from the function and test if the values have changed. The caveats described above should be considered bugs if they happen.

swim/memberlist.go
thanodnl: We can change it to the early return, might be easier to understand.

However we cannot use defer here!

There are calls executed below that acquire their own lock which would result in a deadlock.

swim/memberlist.go
thanodnl: That is a good catch, I will double check if it would return in a `null` in the protocol and if so add an explicit test with documentation to it.

membership/interface.go
```diff
@@ -0,0 +1,19 @@
+package membership
+
+// Member defines a member of the membership. It can be used by applications to
+// apply specific business logic on Members. Examples are:
+// - Get the address of a member for RPC calls, both forwarding of internal
+//   calls that should target a Member
+// - Decissions to include a Member in a query via predicates.
+type Member interface {
```
thanodnl: A separate `Membership` interface over `swim` its self gives benefits due to separation of concern. Historically the hashring has been very tightly coupled to the swim states eg. Alive, Suspect, Faulty ... and other assumptions like address and identity being the same.

This interface defines a separation of concern describing what a Membership looks like on a higher level. Only including members that are actively participating, no view over faulty, leave and tombstone nodes any more.

Due to the lack of the membership package and interface the first predicates have been implemented in the swim package. Making projects that want to use predicates relying on the swim package which have a lot of hard to understand subtleties in there for applications.

Lastly with this interface we can do more experimentation with different implementations of membership protocols. Currently we have only swim, but it might be beneficial to experiment with a zookeeper or etcd backed membership protocol for a more consistent view of the online members. This could be configured transparent to the service.

swim/memberlist.go
mennopruijssers: due to the `LabelMap` it's not possible to do a equality check. 😢

membership/interface.go
CorgiMan: Thanks for the explanation Nils!

PR #190, opened 2016-11-28
[Feature] Identity carry over part 3 / Hashring abstraction

thanodnl:
This PR is part 3 for identity carry over.

This PR hooks up the hashring to the newly introduced `membership` package. Instead of talking about servers as strings we respond to `memberhip.MemberChange` slices. This has the benefit that we have access to more fields exposed on a member than just its address.


hashring/hashring.go
```diff
@@ -129,88 +135,117 @@ func (r *HashRing) AddServer(address string) bool {
...
 		})
 	}
 	r.Unlock()
 	return ok
 }
 
 // This function isn't thread-safe, only call it when the HashRing is locked.
-func (r *HashRing) removeServerNoLock(address string) bool {
-	if _, ok := r.serverSet[address]; !ok {
+func (r *HashRing) removeServerNoLock(server membership.Member) bool {
```
mennopruijssers: Same here, I would expect `delete(r.serverSet, server.GetAddress())` here

hashring/hashring.go
```diff
@@ -129,88 +135,117 @@ func (r *HashRing) AddServer(address string) bool {
...
-		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Delete(replicaPoint(r.hashfunc(address)))
+		r.tree.Delete(
+			r.replicaPointForServer(server, i),
+		)
 	}
 }
 
+func (r *HashRing) ProcessMembershipChangesServers(changes []membership.MemberChange) {
+	r.Lock()
```
mennopruijssers: Don't think we need it but it might be a performance improvement to first loop through the changes to check for relevant changes. If there are, lock, process them, compute checksum and unlock

hashring/hashring.go
```diff
@@ -129,88 +135,117 @@ func (r *HashRing) AddServer(address string) bool {
...
 	for i := 0; i < r.replicaPoints; i++ {
-		address := fmt.Sprintf("%s%v", server, i)
-		r.tree.Delete(replicaPoint(r.hashfunc(address)))
+		r.tree.Delete(
+			r.replicaPointForServer(server, i),
+		)
 	}
 }
 
+func (r *HashRing) ProcessMembershipChangesServers(changes []membership.MemberChange) {
```
mennopruijssers: This function looks a lot like `addRemoveServersNoLock`. Is it possible to pre-process the changes (without locking) and use `addRemoveServersNoLock` for the actual heavy lifting?

hashring/hashring.go
```diff
@@ -234,7 +269,7 @@ func (r *HashRing) Servers() []string {
 
 // This function isn't thread-safe, only call it when the HashRing is locked.
 func (r *HashRing) copyServersNoLock() []string {
-	var servers []string
+	servers := make([]string, 0, len(r.serverSet))
```
mennopruijssers: 👍

hashring/hashring.go
thanodnl: I think the `addRemoveServersNoLock` should go away, basically this is a version of said function the also updates, but I found `addRemoveUpdateServersNoLock` not the best name. I will remove the first function.

ringpop_test.go
```diff
@@ -317,15 +318,15 @@ func (s *RingpopTestSuite) TestHandleEvents() {
 	s.Equal(int64(1), s.stats.read("ringpop.127_0_0_1_3001.refuted-update"), "missing refuted-update stat")
 
 	// double check the counts before the event
-	s.Equal(int64(11), s.stats.read("ringpop.127_0_0_1_3001.ring.server-added"), "incorrect count for ring.server-added before RingChangedEvent")
-	s.Equal(int64(2), s.stats.read("ringpop.127_0_0_1_3001.ring.server-removed"), "incorrect count for ring.server-removed before RingChangedEvent")
+	s.Equal(int64(9), s.stats.read("ringpop.127_0_0_1_3001.ring.server-added"), "incorrect count for ring.server-added before RingChangedEvent")
```
thanodnl: There was an actual bug that caused servers that were already in the hashring to show up in the event that gets statted here. If there was at least 1 change made to the ring it would stat that all changes were made. With the new code it will only stat the actual servers added and removed from the ring. This caused the count to drop by 1.

Secondly there was a tests where 3 changes were made in rapid succession for the same node which is an unrealistic scenario which cannot happen and cannot be expressed easily with the new event. Causing an other drop in the `ring.server-added` stat.

ringpop_test.go
```diff
@@ -317,15 +318,15 @@ func (s *RingpopTestSuite) TestHandleEvents() {
...
+	s.Equal(int64(1), s.stats.read("ringpop.127_0_0_1_3001.ring.server-removed"), "incorrect count for ring.server-removed before RingChangedEvent")
```
thanodnl: The same bug from above caused this stat to drop by 1.

hashring/hashring.go
```diff
@@ -129,88 +135,117 @@ func (r *HashRing) AddServer(address string) bool {
 }
 
 // This function isn't thread-safe, only call it when the HashRing is locked.
-func (r *HashRing) addServerNoLock(address string) bool {
-	if _, ok := r.serverSet[address]; ok {
+func (r *HashRing) addServerNoLock(server membership.Member) bool {
+	if _, ok := r.serverSet[server.GetAddress()]; ok {
```
thanodnl: I see what you are getting at. I think it is safe to inline the `addReplicasNoLock` in here which will make the whole function more coherent.

hashring/hashring_test.go
```diff
@@ -179,7 +187,7 @@ func TestLookup(t *testing.T) {
 
 func TestLookupDistribution(t *testing.T) {
 	ring := New(farm.Fingerprint32, 5)
-	addresses := genAddresses(1, 1, 1000)
+	addresses := genMembers(1, 1, 1000)
```
thanodnl: rename variable to `members`

hashring/hashring_test.go
```diff
@@ -258,14 +266,14 @@ func TestLookupNNoGaps(t *testing.T) {
 
 func TestLookupNOverflow(t *testing.T) {
 	ring := New(farm.Fingerprint32, 10)
-	addresses := genAddresses(1, 1, 10)
+	addresses := genMembers(1, 1, 10)
```
thanodnl: rename variable to `members`

hashring/hashring_test.go
```diff
@@ -258,14 +266,14 @@ func TestLookupNNoGaps(t *testing.T) {
...
 	ring.AddRemoveServers(addresses, nil)
 	assert.Len(t, ring.LookupN("a random key", 20), 10, "expected that LookupN caps results when n is larger than number of servers")
 }
 
 func TestLookupNLoopAround(t *testing.T) {
 	ring := New(farm.Fingerprint32, 1)
-	addresses := genAddresses(1, 1, 10)
+	addresses := genMembers(1, 1, 10)
```
thanodnl: rename variable to `members`

hashring/hashring_test.go
```diff
@@ -291,7 +299,7 @@ func TestLookupN(t *testing.T) {
 
 	unique := make(map[string]bool)
 
-	addresses := genAddresses(1, 1, 10)
+	addresses := genMembers(1, 1, 10)
```
thanodnl: rename variable to `members`

hashring/hashring_test.go
```diff
@@ -321,10 +329,12 @@ func TestLookupN(t *testing.T) {
 	assert.Len(t, unique, 9, "expected to get nine unique servers")
 }
 
-func genAddresses(host, fromPort, toPort int) []string {
-	var addresses []string
+func genMembers(host, fromPort, toPort int) []membership.Member {
+	var addresses []membership.Member
```
thanodnl: rename variable to `members`

ringpop.go
```diff
@@ -30,6 +30,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/uber/ringpop-go/membership"
+
```
thanodnl: rearrange imports

hashring/hashring.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/uber-common/bark"
 	"github.com/uber/ringpop-go/events"
 	"github.com/uber/ringpop-go/logging"
+	"github.com/uber/ringpop-go/membership"
```
thanodnl: rearrange imports

hashring/hashring_test.go
```diff
@@ -27,6 +27,8 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/uber/ringpop-go/membership"
```
thanodnl: rearrange imports

utils_test.go
```diff
@@ -25,11 +25,26 @@ import (
 	"sync"
 	"time"
 
-	"github.com/uber-common/bark"
+	"github.com/uber/ringpop-go/membership"
 	"github.com/uber/ringpop-go/swim"
+
```
thanodnl: rearrange imports

ringpop_test.go
mennopruijssers: 👍

PR #191, opened 2016-11-28
[Feature] Identity carry over part 4 / Identity carry over

thanodnl:
This PR is part 4 of identity carry over.

This PR actually adds the meat of the story with actual identity carry over. It extends the replica point with more information that will later be used by a new checksumming mechanism and a mechanism that will be backwards compatible with the current checksumming mechanism but added identity support.

Identities are stored in ringpop labels, this mechanism allows for reliable synchronization within the cluster. It uses a key in the private namespace of labels `__identity`. This prevent applications from altering it outside of the supported channels.

If the identity key is missing it will fall back to the address of the Member. This allows for backwards compatibility with other versions and an easy upgrade path when identity is not set on a member.


ringpop.go
```diff
@@ -783,6 +783,15 @@ func (rp *Ringpop) Labels() (*swim.NodeLabels, error) {
 	return rp.node.Labels(), nil
 }
 
+func (rp *Ringpop) SetIdentity(identity string) error {
```
mennopruijssers: Do we really want to support changing the identity after bootstrap? and why don't we support it before bootstrapping?

swim/member.go
```diff
@@ -77,6 +79,17 @@ func (m Member) Label(key string) (value string, has bool) {
 	return
 }
 
+func (m Member) Identity() string {
```
mennopruijssers: docs?

ringpop.go
thanodnl: Probably not, but this made it possible to build an application that could set its identity via RPC and validate the change got picked up by the rest of the network.

hashring/hashring.go
```diff
@@ -46,10 +46,14 @@ type Configuration struct {
 	ReplicaPoints int
 }
 
-type replicaPoint int
+type replicaPoint struct {
```
CorgiMan: +1

We have a lot of hashes. membershipHash, replicaPointHash, ringHash

hashring/hashring_test.go
```diff
@@ -302,44 +305,185 @@ func TestLookupN(t *testing.T) {
 	assert.Len(t, unique, 9, "expected to get nine unique servers")
 }
 
-func TestProcessMembershipChanges(t *testing.T) {
-	ring := New(farm.Fingerprint32, 10)
+type ProcessMembershipChangesSuite struct {
+	suite.Suite
+	ring    *HashRing
+	member1 fakeMember
```
thanodnl: It smells that we want an array (not a slice) here; `[4]fakeMember` but I am not sure since all tests seem to be a specific order of events on the members. Besides from initialization they don't seem to be operated on in batches.

What were your thought while writing the code and what made you choose for this approach?

mennopruijssers: Good point. `[4]fakeMember` should have worked as well.

PR #192, opened 2016-11-28
[Feature] Identity carry over part 5 / Multi ring checksum

thanodnl:
This PR is part 5 for identity carry over.

In this PR a new mechanism for calculating ring checksums is added. It aims to be forward compatible by naming checksums after the algorithm used so that new algorithms can be added and old algorithms can be removed when appropriate.

At the same time the old algorithm will start its EOL and all references to that are renamed to indicate legacy. The old algorithm does not capture all aspects of the hashring and will not warn about programmer error in the hashring.

The newly added `replica` checksummer will walk the ring from the beginning to the end and capture all significant datapoints of a `replicapoint` in its checksum that should prevent two rings that are out of sync to provide the same checksum.


hashring/checksum.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/dgryski/go-farm"
+)
+
+// Checksum computes a checksum for an instance of a HashRing. The
+// checksum can be used to compare two rings for equality.
+type Checksum interface {
+	// Compute calculates the checksum for the hashring that is passed in.
+	// Compute will be called while having atleast a readlock on the hashring so
+	// it is safe to read from the ring, but not safe to chagne the ring. There
```
mennopruijssers: s/chagne/change

hashring/checksum.go
```diff
@@ -0,0 +1,67 @@
...
+
+type replicaPointChecksum struct{}
+
+func (r *replicaPointChecksum) Compute(ring *HashRing) uint32 {
+	buffer := bytes.Buffer{}
+
+	ring.tree.root.walk(func(node *redBlackNode) bool {
+		buffer.WriteString(strconv.Itoa(node.key.(replicaPoint).point))
+		buffer.WriteString("-")
+		buffer.WriteString(node.value.(string))
```
mennopruijssers: Doesn't ik make more sense to use `node.key.identity` instead of `node.value` here?

hashring/hashring.go
```diff
@@ -89,32 +99,47 @@ func New(hashfunc func([]byte) uint32, replicaPoints int) *HashRing {
...
-	addresses := r.copyServersNoLock()
-	sort.Strings(addresses)
-	bytes := []byte(strings.Join(addresses, ";"))
-	old := r.checksum
-	r.checksum = farm.Fingerprint32(bytes)
-
-	if r.checksum != old {
+func (r *HashRing) Checksums() (checksums map[string]uint32) {
+	r.RLock()
+	// even though the map is immutable the pointer to it is not so it requires
```
mennopruijssers: Should we add a check if checksums are already calculated and compute them if not?

hashring/hashring.go
```diff
@@ -89,32 +99,47 @@ func New(hashfunc func([]byte) uint32, replicaPoints int) *HashRing {
...
+	// calculate the legacy identy only based checksum
+	legacy := identityChecksum{}
+	old := r.legacyChecksum
+	r.legacyChecksum = legacy.Compute(r)
+
+	if old != r.legacyChecksum {
 		r.logger.WithFields(bark.Fields{
-			"checksum":    r.checksum,
+			"checksum":    r.legacyChecksum,
 			"oldChecksum": old,
```
mennopruijssers: Do we want to log the other checksums?

hashring/hashring.go
```diff
@@ -89,32 +99,47 @@ func New(hashfunc func([]byte) uint32, replicaPoints int) *HashRing {
...
+// and updates the in memory map with a new map containing the new checksums.
+func (r *HashRing) computeChecksumsNoLock() {
+	r.checksums = make(map[string]uint32)
+
+	// calculate all configured checksums
+	for name, checksummer := range r.checksummers {
+		r.checksums[name] = checksummer.Compute(r)
+	}
+
+	// calculate the legacy identy only based checksum
```
mennopruijssers: s/identy/identity

hashring/checksum.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// Checksum computes a checksum for an instance of a HashRing. The
+// checksum can be used to compare two rings for equality.
+type Checksum interface {
+	// Compute calculates the checksum for the hashring that is passed in.
+	// Compute will be called while having atleast a readlock on the hashring so
+	// it is safe to read from the ring, but not safe to chagne the ring. There
+	// might be multiple Checksum Computes initiated at the same time, but every
+	// Checksum will only be called once per hashring at once
```
mennopruijssers: > every Checksum will only be called once per hashring at once
not sure if we should define this behaviour. Why restrict us when not needed?

hashring/checksum.go
```diff
@@ -0,0 +1,67 @@
...
+	Compute(ring *HashRing) (checksum uint32)
+}
+
+// addressChecksum calculates checksums for all addresses that are added to the
+// hashring.
+type addressChecksum struct{}
```
mennopruijssers: I don't think we use nor need the old style `addressChecksum`

hashring/checksum.go
```diff
@@ -0,0 +1,67 @@
...
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/dgryski/go-farm"
+)
+
+// Checksum computes a checksum for an instance of a HashRing. The
+// checksum can be used to compare two rings for equality.
+type Checksum interface {
```
CorgiMan: I don't think we should have an interface here. A simple `func Compute(ring *HashRing)` function would be sufficient. Any reason we want this interface.

Also the interface name Checksumer with a method would be more idiomatic I think.

thanodnl: I don't agree here. A function can also implement an interface when needed, but if you want to use a struct it is harder if you don't use an interface as input.

The naming you are right on, we should call the interface `Checksummer`.

PR #193, opened 2016-12-07
Correct ineffassign warnings

gliptak:
https://goreportcard.com/report/github.com/uber/ringpop-go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ addons:
     packages:
     - python-virtualenv
 
+go_import_path: github.com/uber/ringpop-go
+
 install:
   - go get -u github.com/Masterminds/glide
   - go get github.com/axw/gocov/gocov
--- a/ringpop_test.go
+++ b/ringpop_test.go
@@ -532,7 +532,7 @@ func (s *RingpopTestSuite) TestChecksum() {
 	s.Equal(ready, s.ringpop.state)
 	checksum, err = s.ringpop.Checksum()
 	s.NoError(err)
-	//s.NotZero(checksum)
+	s.NotZero(checksum)
 }
 
 // TestApp tests that App() returns the correct app name.
--- a/swim/ping_request_test.go
+++ b/swim/ping_request_test.go
@@ -172,7 +172,9 @@ func TestIndirectPing2(t *testing.T) {
 	cont := make(chan bool, 1)
 
 	tnodes := genChannelNodes(t, 4)
-	defer destroyNodes(tnodes...)
+	defer func() {
+		destroyNodes(tnodes...)
+	}()
 
 	// don't bootstrap the target
 	sender, helper1, helper2, target := tnodes[0], tnodes[1], tnodes[2], tnodes[3]
--- a/discovery/jsonfile/lib_test.go
+++ b/discovery/jsonfile/lib_test.go
@@ -57,5 +57,6 @@ func TestNiceJSONFile(t *testing.T) {
 	// Now actually test the DiscoverProvider:Hosts() will return the two
 	provider := New(tmpfile.Name())
 	res, err := provider.Hosts()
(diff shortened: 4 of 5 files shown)


gliptak:
https://travis-ci.org/uber/ringpop-go/jobs/181822850 Go 1.5 fails with:

```
# golang.org/x/tools/go/gcimporter15

../../../golang.org/x/tools/go/gcimporter15/bexport.go:557: undefined: constant.ToFloat

../../../golang.org/x/tools/go/gcimporter15/gcimporter.go:396: pkg.SetName undefined (type *types.Package has no field or method SetName)

make: *** [dev_deps] Error 2
```

mennopruijssers:
We removed go 1.5. Please rebase on the `dev` branch and edit your PR to change the base to `dev`.

mennopruijssers:
But thanks for looking into this! Would be awesome if you could sign or CLA: https://cla-assistant.io/uber/ringpop-go?pullRequest=193.

Thanks!

gliptak:
@mennopruijssers I signed the CLA and rebased PR

.travis.yml
```diff
@@ -8,6 +8,8 @@ addons:
     packages:
     - python-virtualenv
 
+go_import_path: github.com/uber/ringpop-go
```
mennopruijssers: Nice, this makes building when forking way easier!

ringpop_test.go
```diff
@@ -530,9 +530,8 @@ func (s *RingpopTestSuite) TestChecksum() {
 	s.Require().NoError(err, "unable to bootstrap single node cluster")
 
 	s.Equal(ready, s.ringpop.state)
-	checksum, err = s.ringpop.Checksum()
+	_, err = s.ringpop.Checksum()
```
mennopruijssers: I prefer uncommenting `s.NotZero(checksum)`. Just tested it locally and tests seems to pass.

swim/ping_request_test.go
```diff
@@ -179,7 +179,6 @@ func TestIndirectPing2(t *testing.T) {
 	bootstrapNodes(t, sender, helper1, helper2)
 	waitForConvergence(t, 100, sender, helper1, helper2)
 
-	tnodes = append(tnodes, newChannelNode(t))
```
mennopruijssers: Instead of removing this line, we should change the `defer destroyNodes(tnodes...)` to 
```
defer func() {
	destroyNodes(tnodes...)
}()
```

Otherwise the channelNode created on line 182 isn't destroyed correctly.

See this playground for a POC: https://play.golang.org/p/mE6J0Aaakg

.travis.yml
gliptak: Many might not build their forks with Travis ... But good to have regardless

ringpop_test.go
gliptak: done

swim/ping_request_test.go
gliptak: Thank you for the example. Updated

swim/ping_request_test.go
```diff
@@ -172,7 +172,9 @@ func TestIndirectPing2(t *testing.T) {
 	cont := make(chan bool, 1)
 
 	tnodes := genChannelNodes(t, 4)
-	defer destroyNodes(tnodes...)
+	defer func() {
```
CorgiMan: Why is this change necessary?

mennopruijssers: A couple of lines below, there's another node added to the `tnode` slice. That extra node won't be destroyed without wrapping `destroyNodes` in another func.

mennopruijssers: (We discovered this bug)

CorgiMan: Ah that makes sense! Thanks

PR #194, opened 2016-12-14
[Feature] Identity carry over part 6 / Configure `Identity` from constuctor

mennopruijssers:
Instead of specifying the identity after bootstrap, rename the existing `Identity` and `IdentityResolverFunc` options to `Address` and `AddressResolverFunc` and add an `Identity` option to specify the identity.

Note: this introduces a backwards incompatible change. `CHANGES.md` contains a `go fmt` command to resolve any issues.


CHANGES.md
```diff
@@ -1,6 +1,24 @@
 ringpop-go changes
 ==================
 
+NEXT
```
mennopruijssers: `NEXT` will be replaced with the version (probably v0.8.0) during the release.

options.go
```diff
@@ -173,37 +178,50 @@ func Statter(s log.StatsReporter) Option {
 	}
 }
 
-// Identity is used to specify a static hostport string as this Ringpop
-// instance's identity.
+// Identity is used to specify an identity as this Ringpop instance's identity.
```
dansimau: Help text could be better here.

E.g.:

```
// Identity can be used to specify a custom string as the unique identifier for
// this node. The identity should be unique amongst other Ringpop instances; it
// is used in the hashring.
//
// By default, the hostport/address of the node is used as the identity in the
// hashring. An error is thrown if a hostport is manually specified using this
// option, as this would lead to unexpected behaviour. If you want to override
// the node's listening address, use the `Address` option.
```

scripts/testpop/testpop.go
```diff
@@ -48,6 +48,8 @@ var (
 	faultyPeriod    = flag.Int("faulty-period", 24*60*60*1000, "The lifetime of a faulty member in ms. After that the member becomes a tombstone.")
 	tombstonePeriod = flag.Int("tombstone-period", 5000, "The lifetime of a tombstone member in ms. After that the member is removed from the membership.")
 
+	identity = flag.String("identity", "", "specify an identity")
```
dansimau: I think we should at least say: `specify an identity for this node in the hashring`. Otherwise no point in even having the help text. The fact that it's setting something called identity is already obvious from the name.

CHANGES.md
```diff
@@ -1,6 +1,24 @@
...
+------
+
+### Release notes
+
+#### Change to identity-option
```
dansimau: Nice

options.go
mennopruijssers: +1

CHANGES.md
```diff
@@ -1,6 +1,24 @@
...
+
+Prior to ringpop NEXT the address was used as the identity of a member. Starting with version NEXT, it's possible to
```
thanodnl: Don't forget to change `NEXT` here as well when releasing.

options.go
```diff
@@ -173,37 +173,37 @@ func Statter(s log.StatsReporter) Option {
...
-// IdentityResolver is a function that returns the listen interface/port
-// that Ringpop should identify as.
-type IdentityResolver func() (string, error)
+// AddressResolver is a function that returns the listen interface/port
+// that Ringpop should use as its address.
+type AddressResolver func() (string, error)
 
-// IdentityResolverFunc is used to specify a function that will called when
-// the Ringpop instance needs to resolve its identity (typically, on
+// AddressResolverFunc is used to specify a function that will called when
```
thanodnl: nit: s/will called/will be called/

test/lib.sh
```diff
@@ -2,7 +2,7 @@
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="menno/allow-multi-checksum"
+declare ringpop_common_branch="feature/identity"
```
thanodnl: Lets make sure to change this to `master` before landing.

options.go
```diff
@@ -173,6 +178,19 @@ func Statter(s log.StatsReporter) Option {
 	}
 }
 
+// Identity is used to specify an identity as this Ringpop instance's identity.
+// Since a ringpop instance is by default identified by its hostport, it's not allowed to manually specify a
+// hostport as its identity.
+func Identity(identity string) Option {
+	return func(r *Ringpop) error {
+		if util.HostportPattern.MatchString(identity) {
+			return errors.New("a hostport is not valid as an identity")
```
mennopruijssers: Was thinking the same but all other errors in this file are created inline. I chose to follow the code style (of this file).

test/lib.sh
mennopruijssers: done!

PR #195, opened 2016-12-16
Improve replica-point naming when identities are used

mennopruijssers:
Together with supporting identities we can fix a longstanding bugs where replica-points can collide when there are multiple instances running on the same machine.


test/lib.sh
```diff
@@ -2,7 +2,7 @@
 
 declare project_root="${0%/*}/.."
 declare ringpop_common_dir="${0%/*}/ringpop-common"
-declare ringpop_common_branch="master"
+declare ringpop_common_branch="better-replicapoint-naming"
```
CorgiMan: Don't forget to set this back to master when you land the diff

PR #197, opened 2016-12-21
[Hotfix] Fix forward header leaking

thanodnl:
We have found an issue when the `ctx` passed into a remote call is blindly passed in as a context to an outgoing call. Endpoints that are sharded via the ringpop adapter might accidentally leak the header indicating a forwarded message. When this happens subsequent requests to sharded endpoints might unintentionally not be sharded.

The fix proposed here will prevent leaking the forwarded header to the application by not setting the header if the call is made to a non remote endpoint, and secondly remove the forwarding header from the forwarded call before dispatching it to the implementation.

Still required:
- [x] Rename the header so services can protect them self without relying on other services to upgrade
- [x] Change the payload of the header to a list of keys being forwarded for to make it easier in the future to implement key-consistent-forwarding


forward/forwarder.go
```diff
@@ -186,6 +192,9 @@ func SetForwardedHeader(ctx thrift.Context) thrift.Context {
 // the message. When a message has already been forwarded by ringpop the
 // forwarding logic should not be used to prevent unbound forwarding.
 func HasForwardedHeader(ctx thrift.Context) bool {
-	_, ok := ctx.Headers()[forwardedHeaderName]
+	_, ok := ctx.Headers()[ForwardedHeaderName]
+	if ok {
+		delete(ctx.Headers(), ForwardedHeaderName)
```
mennopruijssers: I'm not sure if we want the delete as a side-effect of a call to `HasForwardedHeader`.
Maybe change the function name to `DeleteForwardedHeader` and return true if deletion was successful (=header was present)?

router/router.go
```diff
@@ -36,12 +36,15 @@ type router struct {
...
+	clientCache map[string]struct {
+		client   interface{}
+		isRemote bool
+	}
 }
 
 // A Router creates instances of TChannel Thrift Clients via the help of the ClientFactory
 type Router interface {
-	GetClient(key string) (interface{}, error)
+	GetClient(key string) (client interface{}, isRemote bool, err error)
```
mennopruijssers: Add docs

router/router.go
```diff
@@ -85,17 +91,19 @@ func (r *router) handleChange(change swim.Change) {
...
+func (r *router) GetClient(key string) (client interface{}, isRemote bool, err error) {
 	dest, err := r.ringpop.Lookup(key)
 	if err != nil {
-		return nil, err
+		return
 	}
 
 	r.rw.RLock()
-	client, ok := r.clientCache[dest]
+	cache, ok := r.clientCache[dest]
```
mennopruijssers: s/cache/cachedClient?

router/router.go
```diff
@@ -36,12 +36,15 @@ type router struct {
 	channel *tchannel.Channel
 
 	rw          sync.RWMutex
-	clientCache map[string]interface{}
+	clientCache map[string]struct {
```
mennopruijssers: Maybe define a struct for the cache entries?

forward/forwarder.go
```diff
@@ -164,24 +165,33 @@ func (f *Forwarder) ForwardRequest(request []byte, destination, service, endpoin
 var (
 	// ForwardedHeaderName is the name used by the ringpop adapter to indicate
```
mennopruijssers: I guess this doc needs to be updated now.

forward/forwarder_test.go
```diff
@@ -28,6 +28,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
mennopruijssers: out of order?

forward/forwarder_test.go
```diff
@@ -337,31 +339,26 @@ func TestForwarderTestSuite(t *testing.T) {
...
+	ctx = SetForwardedHeader(ctx, nil)
+
+	assert.Equal(t, "[]", ctx.Headers()[ForwardedHeaderName], "expected the forwarding header to be set and be an empty array instead of null for the nil pointer")
 
 	ctx, _ = thrift.NewContext(0 * time.Second)
 	ctx = thrift.WithHeaders(ctx, map[string]string{
 		"keep": "this key",
 	})
-	ctx = SetForwardedHeader(ctx)
+	ctx = SetForwardedHeader(ctx, []string{"foo"})
```
mennopruijssers: Maybe also add a test to prove multiple keys for completeness?

examples/keyvalue/README.md
```diff
@@ -0,0 +1,63 @@
+This application shows a complex example where a node makes requests to its self from a sharded endpoint.
```
mennopruijssers: s/its self/itself

examples/keyvalue/README.md
```diff
@@ -0,0 +1,63 @@
...
+    $ go build
+
+Start a custer of 5 nodes using [tick-cluster][1]:
+
+    $ tick-cluster.js --interface=127.0.0.1 -n 5 keyvalue &> tick-cluster.log &
+    $ sleep 5
+
+Set some reference keys that are sharded around the cluster using [tcurl][2]:
+
+    $ tcurl keyvalue -P hosts.json --thrift ./keyvalue.thrift KeyValueService::Set '{"key":"127.0.0.1:30010", "value": "foo"}'
```
mennopruijssers: Maybe add an explanation why the key is the node's "host:port + 0"?

examples/keyvalue/main.go
```diff
@@ -0,0 +1,149 @@
...
+		gen.KeyValueServiceConfiguration{
+			Get: &gen.KeyValueServiceGetConfiguration{
+				Key: func(ctx thrift.Context, key string) (shardKey string, err error) {
+					return key, nil
+				},
+			},
+
+			GetAll: &gen.KeyValueServiceGetAllConfiguration{
+				Key: func(ctx thrift.Context, keys []string) (shardKey string, err error) {
+					// use the node listening on 127.0.0.1:3004 as the node that should answer
```
mennopruijssers: why not just `master`? It doesn't really matter which one answers as long as all nodes in the cluster agree which one should answer right?

thanodnl: By knowing which node is answering we can craft both a call directly to the node and a call specifically NOT to the node. Otherwise we should call every node from the test to be sure we have all combinations covered.

mennopruijssers: makes sense; but it feels weird to hardcode a "dynamic" value (provided by command line argument) here.

test/remoteservice/remoteservice_test.go
```diff
@@ -109,6 +170,143 @@ func TestRingpopRemoteServiceAdapterCallRemote(t *testing.T) {
 	serviceImpl.AssertNotCalled(t, "RemoteCall", mock.Anything, "hello")
 }
 
+func TestRingpopRemoteServiceAdapterCallRemotePerservingHeaders(t *testing.T) {
```
CorgiMan: *Preserving

router/router.go
```diff
@@ -85,38 +95,44 @@ func (r *router) handleChange(change swim.Change) {
...
 	defer r.rw.Unlock()
 
 	// double check it is not created between read and complete lock
-	client, ok = r.clientCache[dest]
+	cachedEntry, ok = r.clientCache[dest]
 	if ok {
-		return client, nil
+		client = cachedEntry.client
+		isRemote = cachedEntry.isRemote
+		return
```
CorgiMan: Naked returns are discouraged for bigger functions. https://tour.golang.org/basics/7

forward/forwarder_test.go
thanodnl: test has been added a couple of lines below

test/remoteservice/remoteservice_test.go
```diff
@@ -198,7 +198,7 @@ func TestRingpopRemoteServiceAdapterCallRemotePerservingHeaders(t *testing.T) {
 			"hello": "world",
 			"foo":   "bar",
 			"baz":   "42",
-		}, headers, "exected existing headers to be perserved")
+		}, headers, "exected existing headers to be preserved")
```
CorgiMan: expect

PR #203, opened 2017-04-13
Add WithError() to NoLogger

malandrew:
The Bark logger interface now has a `WithError(err Error) bark.Logger` method. Just adding that to the NoLogger interface.

PR #204, opened 2017-04-28
glide up and fix noLogger

albertyfwu:
I ran into some issues while running glide up to update the noLogger to respect the latest version of bark.Logger. Since glide up didn't bump the version of bark, I manually edited glide.lock. Could somebody help advise if this is correct?


dansimau:
Is this a dupe of #203?

albertyfwu:
@dansimau Yes.

PR #205, opened 2017-06-15
update Sirupsen to the correct repo

datoug:


glide.lock
```diff
@@ -28,7 +28,7 @@ imports:
 - name: github.com/rcrowley/go-metrics
   version: dee209f2455f101a5e4e593dea94872d2c62d85d
 - name: github.com/Sirupsen/logrus
```
boy-uber: Do we need to change Sirupsen here too?

malandrew:
This is a dupe of #208

PR #206, opened 2017-06-20
Add option to enforce identical app name in pings

yulunli:
Always reject pings from wrong app name

### DIFF
--- a/swim/ping_handler.go
+++ b/swim/ping_handler.go
@@ -20,7 +20,12 @@
 
 package swim
 
-import "time"
+import (
+	"errors"
+	"time"
+
+	log "github.com/uber-common/bark"
+)
 
 func handlePing(node *Node, req *ping) (*ping, error) {
 	if !node.Ready() {
@@ -33,6 +38,23 @@ func handlePing(node *Node, req *ping) (*ping, error) {
 		Source:  req.Source,
 		Changes: req.Changes,
 	})
+
+	if req.App == "" {
+		if node.requiresAppInPing {
+			node.logger.WithFields(log.Fields{
+				"source": req.Source,
+			}).Warn("Rejected ping from unknown ringpop app")
+			return nil, errors.New("Pinged ringpop requires app name")
+		}
+	} else {
+		if req.App != node.app {
+			node.logger.WithFields(log.Fields{
+				"app":    req.App,
+				"source": req.Source,
+			}).Warn("Rejected ping from wrong ringpop app")
+			return nil, errors.New("Pinged ringpop has a different app name")
+		}
+	}
 
 	node.serverRate.Mark(1)
 	node.totalRate.Mark(1)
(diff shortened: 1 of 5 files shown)


swim/ping_handler.go
```diff
@@ -34,6 +37,18 @@ func handlePing(node *Node, req *ping) (*ping, error) {
 		Changes: req.Changes,
 	})
 
+	if req.App == "" {
+		if node.requiresAppInPing {
+			node.logger.Warnf("Rejected ping from unknown ringpop app %s", req.App)
```
mennopruijssers: `req.App` will always be empty here.

swim/ping_handler.go
```diff
@@ -34,6 +37,18 @@ func handlePing(node *Node, req *ping) (*ping, error) {
...
+			return nil, errors.New("Pinged ringpop requires app name")
+		}
+	} else {
+		if req.App != node.app {
+			node.logger.Warnf("Rejected ping from wrong ringpop app %s", req.App)
```
mennopruijssers: normally we would use something like `node.logger.WithField("app", req.App).Warn("Rejected ping from wrong ringpop app")`. That way searching in ES is easier because all the messages are the same.

swim/ping_handler.go
```diff
@@ -47,6 +62,7 @@ func handlePing(node *Node, req *ping) (*ping, error) {
 		Changes:           changes,
 		Source:            node.Address(),
 		SourceIncarnation: node.Incarnation(),
+		App:               node.app,
```
mennopruijssers: This is the reason why the integration tests are failing due to a json-schema validation error. You only added `app` to ping-request while also sending in the ping-response here.

swim/ping_handler.go
```diff
@@ -39,12 +39,12 @@ func handlePing(node *Node, req *ping) (*ping, error) {
 
 	if req.App == "" {
 		if node.requiresAppInPing {
-			node.logger.Warnf("Rejected ping from unknown ringpop app %s", req.App)
+			node.logger.WithFields("req", req).Warnf("Rejected ping from unknown ringpop app")
```
mennopruijssers: Don't add the full req. It could be huge (due to `Changes`). `Source` might be enough.

swim/ping_handler.go
```diff
@@ -39,12 +39,12 @@ func handlePing(node *Node, req *ping) (*ping, error) {
...
 			return nil, errors.New("Pinged ringpop requires app name")
 		}
 	} else {
 		if req.App != node.app {
-			node.logger.Warnf("Rejected ping from wrong ringpop app %s", req.App)
+			node.logger.WithFields("req", req).Warn("Rejected ping from wrong ringpop app")
```
mennopruijssers: Same but also consider adding `req.App`

PR #208, opened 2017-06-22
Use lower case logrus import

yulunli:


glide.yaml
```diff
@@ -1,6 +1,6 @@
 package: github.com/uber/ringpop-go
 import:
-- package: github.com/Sirupsen/logrus
+- package: github.com/sirupsen/logrus
```
mennopruijssers: I think you should also need to update `github.com/uber-common/bark` in glide.lock

malandrew: @yulunli github.com/uber-common/bark was fixed on June 16th by @akshayjshah so this should be okay now. I think you'll just need to update the diff to make sure the glide.lock file is still okay.

yulunli:
LGTM

PR #209, opened 2017-06-22
Allow RequiresAppInPing to be configured

yulunli:


options.go
```diff
@@ -59,6 +59,10 @@ type configuration struct {
 
 	// SelfEvict holds the settings with regards to self eviction
 	SelfEvict swim.SelfEvictOptions
+
+	// RequiresAppInPing configures if ringpop node should reject pings
+	// that don't contain app name
+	RequiresAppInPing bool
```
apersson: Please add an `ringpop.Option` to toggle this field.

PR #211, opened 2017-07-25
hashring.LookupN returns ordered results

malandrew:
The previous implementation of LookupN used a map internally to guarantee uniqueness, but when converted to a slice, ordering of servers in the hashring is lost. This diff updates the internal implementation to maintain a pointer to a slice that is appended during the depth-first search of the red-black tree. This ordered result is converted to a slice of strings to return to the caller.

cc/ @yulunli @akshayjshah

### DIFF
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -391,9 +391,9 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
 	// collected all the servers we want, we have reached the
 	// end of the red-black tree and we need to loop around and inspect the
 	// tree starting at 0.
-	r.tree.LookupNUniqueAt(n, replicaPoint{hash: hash}, unique, &orderedUnique)
+	r.tree.LookupOrderedNUniqueAt(n, replicaPoint{hash: hash}, unique, &orderedUnique)
 	if len(unique) < n {
-		r.tree.LookupNUniqueAt(n, replicaPoint{hash: 0}, unique, &orderedUnique)
+		r.tree.LookupOrderedNUniqueAt(n, replicaPoint{hash: 0}, unique, &orderedUnique)
 	}
 
 	var servers []string
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -383,10 +383,6 @@ func (r *HashRing) LookupN(key string, n int) []string {
 
 // This function isn't thread-safe, only call it when the HashRing is locked.
 func (r *HashRing) lookupNNoLock(key string, n int) []string {
-	if n >= len(r.serverSet) {
-		return r.copyServersNoLock()
-	}
-
 	hash := r.hashfunc(key)
 	unique := make(map[valuetype]struct{})
 	orderedUnique := make([]valuetype, 0, n)
--- a/hashring/hashring.go
+++ b/hashring/hashring.go
@@ -389,18 +389,19 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
 
 	hash := r.hashfunc(key)
 	unique := make(map[valuetype]struct{})
+	orderedUnique := make([]valuetype, 0, n)
 
 	// lookup N unique servers from the red-black tree. If we have not
 	// collected all the servers we want, we have reached the
 	// end of the red-black tree and we need to loop around and inspect the
 	// tree starting at 0.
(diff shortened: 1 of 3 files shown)


malandrew:
FWIW, I'm adding this so I can then add ordered LookupN to the router. My service is stateful and push based, so on graceful shutdown I have an interest in iterating through the N servers in the ring. If the forward fails to the first server, go to the next. If that fails, go to the next, etc.

hashring/hashring.go
```diff
@@ -389,18 +389,19 @@ func (r *HashRing) lookupNNoLock(key string, n int) []string {
...
 	hash := r.hashfunc(key)
 	unique := make(map[valuetype]struct{})
+	orderedUnique := make([]valuetype, 0, n)
 
 	// lookup N unique servers from the red-black tree. If we have not
 	// collected all the servers we want, we have reached the
 	// end of the red-black tree and we need to loop around and inspect the
 	// tree starting at 0.
-	r.tree.LookupNUniqueAt(n, replicaPoint{hash: hash}, unique)
+	r.tree.LookupNUniqueAt(n, replicaPoint{hash: hash}, unique, &orderedUnique)
```
akshayjshah: This is an unusual way to work with slices - it's typical to return the modified slice instead of passing pointers. Why pass a ref here?

hashring/hashring_test.go
```diff
@@ -310,7 +310,8 @@ func TestLookupNLoopAround(t *testing.T) {
 	ring.AddMembers(members...)
 
 	unique := make(map[valuetype]struct{})
-	ring.tree.LookupNUniqueAt(1, replicaPoint{hash: 0}, unique)
+	orderedUnique := make([]valuetype, 0, 1)
+	ring.tree.LookupNUniqueAt(1, replicaPoint{hash: 0}, unique, &orderedUnique)
```
akshayjshah: This doesn't test the correctness of the code you've added; please add unit tests to cover the new logic.

hashring/hashring_test.go
```diff
@@ -399,7 +400,7 @@ func (s *ProcessMembershipChangesSuite) TestAddMember0() {
 	s.ring.ProcessMembershipChanges([]membership.MemberChange{
 		{After: s.members[0]},
 	})
-	mock.AssertExpectationsForObjects(s.T(), s.l.Mock)
+	mock.AssertExpectationsForObjects(s.T(), &s.l.Mock)
```
thanodnl: This line seems to makes the tests fail with a panic.

hashring/hashring_test.go
thanodnl: I'd love a test for correctness here as well.

hashring/hashring.go
malandrew: The `append` in `findNUniqueAbove` replaces the previous slice with a new one at a different address. Is there a way I could accomplish this without using pointers to the slice and updating the pointer?

hashring/hashring_test.go
malandrew: Any suggestions on how to test this given the way the tests are written so far? Naively, I could just iterate a few times and confirm that the returned results for multiple interations of the same call are stable (i.e. same order), but that still wouldn't test that the order is what we expect.

hashring/hashring.go
akshayjshah: You could model this API after `append` and return the modified slice.

hashring/hashring_test.go
akshayjshah: If I understand this change correctly (debatable), you should be able to assert that (1) the returned slice has the correct length, (2) that the sort is stable, and (3) that the first item is the same as the returned value from `Lookup`.

@thanodl may have some ideas for more thorough testing; the only further tests I can imagine would require using the farmhash code directly.

malandrew: I just added tests. See what you all think. I did have to remove the optimization that copied the servers because it copied them out of order because it copies from a map. I also considered a straight traversal copy, but that would make the first server in the tree hot.

hashring/hashring.go
malandrew: Not sure I understand what you mean here. I somewhat suspect that what you're suggesting is what I ended up doing in my test code.

malandrew:
Unless there are no more objections, I'm going to land this in an hour. As far as I can tell all issues raised so far have been addressed.

PR #212, opened 2017-07-26
Implement router.GetNClients

malandrew:
This pull request implements GetNClients in the ringpop router. This is just the router equivalent of the hashring.LookupN function.


yulunli:
LGTM

router/router.go
```diff
@@ -101,6 +106,33 @@ func (r *router) GetClient(key string) (client interface{}, isRemote bool, err e
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, dest := range dests {
+		client, isRemote, err := r.getClientByHost(dest)
+		if err != nil {
+			return nil, err
+		}
+		clients = append(clients, ClientResult{client, isRemote})
```
thanodnl: nit; depending on how often this code is called you might want to consider allocating a slice with the capacity of `len(dests)` before the loop to prevent excessive allocations in the hot path.

router/router.go
```diff
@@ -101,6 +106,33 @@ func (r *router) GetClient(key string) (client interface{}, isRemote bool, err e
...
+	}
+	return clients, nil
+}
+
+func (r *router) getClientByHost(dest string) (client interface{}, isRemote bool, err error) {
```
thanodnl: Nice! 👍

PR #215, opened 2017-11-14
Update tchannel version and add a context option

liweiqing:
Update tchannel version and add a context option so clients using Ringpop.Forward have a way of passing context across ringpop for use cases such as distributed tracing.


yulunli:
LGTM

PR #217, opened 2018-06-11
Stop using tchannel.WriteArgs in requestSender.MakeCall

jcorbin:


jcorbin:
Oh I see... yeah what I'm doing here initially is quite flawed, will sort it out and update.

jcorbin:
Derp, I'd missed that `io.Copy` hides a read buffer allocation; updated to avoid even playing that game (I mean sure, we could've used `io.CopyBuffer`, but seemed better to just use `bytes.Buffer.ReadFrom` instead).

forward/request_sender.go
```diff
@@ -208,13 +236,82 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte, fwdError *err
...
+	}
+
+	if res == nil {
+		*res = []byte{}
+	}
+	*res = append(*res, buf.Bytes()...)
+	return nil
+}
+
+func writeArg(writable func() (tchannel.ArgWriter, error), bs []byte) error {
```
prashantv: you might want to look at [`tchannel.NewArgWriter`](https://godoc.org/github.com/uber/tchannel-go#NewArgWriter) which is basically this. suggest comparing perf to see if it hurts (it uses a closure)

forward/request_sender.go
```diff
@@ -156,6 +159,33 @@ func (s *requestSender) Send() (res []byte, err error) {
 	}
 }
 
+type _bufPool struct {
+	sync.Pool
+	maxCap int64
+}
+
+func (bp *_bufPool) Get() *bytes.Buffer {
+	const allocLimit = 100 * 1024 * 1024 // cap initial allocations at 100MiB
```
prashantv: this buffer pool is complicated by the limiting -- is that really needed? `sync.Pool` is frequently cleared (on every GC), so I'd drop the extra logic unless we have a good reason to need it.

forward/request_sender.go
```diff
@@ -208,13 +236,82 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte, fwdError *err
...
+	buf := bufPool.Get()
+	defer bufPool.Put(buf)
+	if err := bufReadArg(buf, resp.Arg3Reader); err != nil {
+		return err
+	}
+
+	if res == nil {
+		*res = []byte{}
+	}
+	*res = append(*res, buf.Bytes()...)
```
prashantv: we don't need the bytes if `res` is `nil`, can do:
```
if res != nil {
  *res = buf.Bytes()
}
return nil
```

forward/request_sender.go
```diff
@@ -208,13 +236,82 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte, fwdError *err
...
+	return err
+}
+
+func discardArg(readable func() (tchannel.ArgReader, error)) error {
+	var null [32 * 1024]byte
+	ar, err := readable()
+	if err != nil {
+		return err
+	}
+	for {
```
prashantv: can we use `ioutil.Discard.ReadFrom(..)` here?

forward/request_sender.go
jcorbin: Derp, I meant to make that a `*res != nil`... pointers to slices are a bit of a change up.

`res` itself will never be nil ( this isn't an exported interface, the only caller is here in the ringpop forwarding package, and it always passes a pointer to a nil slice.... which actually , append will always work... what even am I saying...

forward/request_sender.go
jcorbin: TIL; will do

forward/request_sender.go
jcorbin: well see... let's try the turn around:
- if we ever happen to forward a 500MiB message... do we always and forever want to be pre-allocating half-GiB buffers for every future message? I for one, say no ;-)
- in fact, I think 100MiB may be too generous in such scenario

prashantv: I would make the buffer pool very simple: just have a prefixed size (E.g., 1-4mb) and always return buffers of that size.

If we see performance issues with that, then we can make it smarter

forward/request_sender.go
```diff
@@ -208,13 +237,59 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte, fwdError *err
...
+		return err
+	}
+
+	buf := bufPool.Get()
+	defer bufPool.Put(buf)
+	if err := copyArg(buf, resp.Arg3Reader); err != nil {
+		return err
+	}
+
+	*res = append(*res, buf.Bytes()...)
```
prashantv: the append to `*res` is a little confusing, i think you want `*res = append([]byte(nil), buf.Bytes()...)`

i'd add a comment that you want a clone of `buf` (since it's pooled)

forward/request_sender.go
```diff
@@ -208,13 +237,59 @@ func (s *requestSender) MakeCall(ctx context.Context, res *[]byte, fwdError *err
...
+	}
+	return err
+}
+
+func copyArg(w io.Writer, readable func() (tchannel.ArgReader, error)) error {
+	ar, err := readable()
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(w, ar)
```
prashantv: this is going to allocate an internal buffer, is that ok?

jcorbin:
```
    If src implements the WriterTo interface, the copy is implemented by
    calling src.WriteTo(dst). Otherwise, if dst implements the ReaderFrom
    interface, the copy is implemented by calling dst.ReadFrom(src).
```

And both of our writers implement `io.ReaderFrom`, you taught me this when you asked for the `ioutil.Discard.ReadFrom` point ;-)

As a curiosity;
- I believe the order of preference expressed in `io.Copy` docs above...
- ...means that you should prefer to have a src that implements `io.WriterTo` when possible
- ...so I guess we could consider doing so in `tchannel-go.fragmentingReader` and/or the arg reader helper

This PR has just been a font of learning for me at least...

prashantv: Ah yes, since the call sites only pass in writers with `ReadFrom`, this will be efficient. To make sure things stay efficient, consider accepting an interface with `ReadFrom` (will need to cast `ioutil.Discard` which should succeed, can use a global to verify that on startup: `var _ io.ReaderFrom = ioutil.Discard.(io.ReaderFrom)`

PR #218, opened 2018-06-12
Re-use checksummer buffers

jcorbin:


hashring/checksummer.go
```diff
@@ -35,22 +36,30 @@ func (i *identityChecksummer) Checksum(ring *HashRing) uint32 {
...
+		if idi > 0 {
+			i.buf.WriteByte(';')
+		}
+		i.buf.WriteString(id)
+	}
+	return farm.Fingerprint32(i.buf.Bytes())
 }
 
-type replicaPointChecksummer struct{}
+type replicaPointChecksummer struct {
```
prashantv: this change the types unsafe when there's concurrency. are we sure they're not used from multiple goroutines? can we add a large note too

PR #233, opened 2022-07-20
Fixed Broken master, Updated tchannel-go to 1.32.0

gandhikrishna:
This PR aims to fix the broken master and upgrade `tchannel-go` to v1.32.0

**_(A). Broken Master_**

The current master of `ringpop-go` is broken, primarily with two issues:

`(1) Install Error: package io/fs: unrecognized import path "io/fs" (import path does not begin with hostname)`

This can be fixed once Go is at a minimum of 1.16. Currently, `ringpop-go` uses `Go 1.15 `see:https://github.com/uber/ringpop-go/blob/master/.github/workflows/test.yaml#L21

`(2) package github.com/vektra/mockery: cannot find package "github.com/hashicorp/hcl/hcl/printer" in any of:
	/opt/hostedtoolcache/go/1.15.15/x64/src/github.com/hashicorp/hcl/hcl/printer (from $GOROOT)
	/home/runner/work/ringpop-go/ringpop-go/src/github.com/hashicorp/hcl/hcl/printer (from $GOPATH)`

This is happening because we are using Go 1.15 in the legacy GOPATH mode rather than in the Go Modules mode. Because GOPATH mode doesn't understand versioning, it's unable to distinguish between HCL version 1 and HCL version 2, and so it's failing to access what looks to be an HCL version 1 dependency. It seems like a common issue with legacy GOPATH mode see:https://github.com/spf13/viper/issues/1076

With this PR, we fixed the above two issues by upgrading Go to 1.16.6 for CI jobs(GitHub workflows here) and removing `mockery` from GOPATH, which caused the issue (2). `mockery` removal required mock check-in for `remoteservicetest`. Added documentation with steps to regenerate mocks.

**_(B). Tchannel-go upgrade to v1.32.0._**
Upgrade tchannel-go, regenerated thrift-gen code.

_Can this bring backward incompatible change to APIs?_
No

_Can this bring wire in-compatibility with apache/thrift?_
No, the vendor version of apache thrift is the same as `GitHub/apache/thrift` currently consumed in `ringpop-go`.

The CI job responsible for uploading coverage reports is failing see:https://github.com/uber/ringpop-go/runs/7464193918?check_suite_focus=true#step:14:17
This is `shogo82148/actions-goveralls` (dependency we use for uploading converge report) not able to find `main.js` file in the path. This is happening again due to legacy GOPATH mode in ringpop-go.


test/remoteservice/README.md
```diff
@@ -0,0 +1,16 @@
+Steps for code re-generation:
```
AllenLuUber, line 1: thanks for adding this!

AllenLuUber (approved):

